PR #2, opened 2016-05-19
timezones: Always use the timezone of blog setting

kirrg001:
refs TryGhost/Ghost#6406
- adding timeZone Service to get the offset (=timezone reg. moment-timezone) overall available
- new publishedAtOffset date as CP using timeZone service and moment-timezone to calculate offset incl. DST
- removing timezone-obj transform as it became obsolete with moment-timezone
- reading timezones from configuration/timezones api endpoint
- adding a moment-utc transform to only work with utc times in backend
- when switching the timezone in the select box, the user will be shown the local time of the selected timezone
- added clock service to show actual time ticking below select box
- default timezone is '(GMT) Greenwich Mean Time : Dublin, Edinburgh, London'
- if no timezone is saved in the settings yet, the default value will be used
- showing local time in 'Publish Date' when it's a draft and no actual publishedAt value exists
- Removed the format 'DD MMM YY @ HH:mm (UTC Z)' which resolves to '01 Jan 16 @ 14:00 (UTC +02:00)'
- Changing the date.js helper in core/server for moment-timezone
- Fix timezone select: updates `selectedTimezone` to return the matching object from `availableTimezones`
- Including timezones in test for date-helper
- update to moment-timezone 0.5.1
- moving form-group of 'selectTimezone' further up so
- Tests:
	- Set except for clock service in test env
	- adding fixtures to mirage
	- adding 'service.ajax' to navigation-test.js
	- adding 'service:ghostPaths' to navigation-test.js
- Code improvements
- Changing clockservice to ES6

### DIFF
--- a/app/components/gh-datetime-input.js
+++ b/app/components/gh-datetime-input.js
@@ -4,7 +4,10 @@ import boundOneWay from 'ghost/utils/bound-one-way';
 import {formatDate} from 'ghost/utils/date-formatting';
 import {invokeAction} from 'ember-invoke-action';
 
-const {Component} = Ember;
+const {
+    Component,
+    RSVP
+} = Ember;
 
 export default Component.extend(TextInputMixin, {
     tagName: 'span',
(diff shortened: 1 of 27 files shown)


ErisDS:
This PR needs a rebase - because it's based off of a branch on the origin, it's going to be easiest to make a new PR.

I've also found a few minor issues with how dates are being displayed in the admin panel. First, I noticed that it no longer says "ago" after dates in the content list:



These should read "Published a month ago".

Then I noticed that the date output in the datetime attribute in the HTML is also not correct:



Hopefully clearing these minor issues up isn't too tricky! It also made me wonder if [publishedAtOffset](https://github.com/TryGhost/Ghost-Admin/pull/2/files#diff-a2b42bc6fa8e2eb95734ef6703bbdf31R71) is the best name for that variable? I'm not quite sure what it represents?!

AileenCGN:
closed in favour of #25 

The new PR includes the timezone changes as well as the UI for scheduling. The issue above is fixed in this PR.

AileenCGN:
This PR can be closed in favour of #25 

The new PR includes the timezone changes as well as the UI for scheduling. The issue above is fixed in this PR.

AileenCGN:
> It also made me wonder if publishedAtOffset is the best name for that variable? I'm not quite sure what it represents?!

publishedAtOffset is the publishedAt variable that contains the timezone offset of the blog. Happy, if you have a better naming solution for this variable...

PR #5, opened 2016-05-19
hide markdown preview #6825

sakulstra:
This pr allowed hiding the markdown preview window.
Hopefully correctly rebased pr from #6825

### DIFF
--- a/app/components/gh-editor.js
+++ b/app/components/gh-editor.js
@@ -14,6 +14,7 @@ export default Component.extend(ShortcutsMixin, {
     activeTab: 'markdown',
     editor: null,
     editorDisabled: undefined,
+    previewIsHidden: false,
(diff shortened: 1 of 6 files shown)


sakulstra:
Sry I somehow committed some old stuff first.
I think I got everything fixed now. Thank you all for your help and patience ;)

kevinansfield:
Hi @sakulstra, thanks for taking the time out to spike on the preview hide/show toggle. After some design discussions we've decided that if the show/hide toggle were to be implemented it would need to be a bit more involved, ie. increasing the font-size, adjusting padding, improving the way that image insertion is done, and so on.

We're also currently reviewing the editor in general where the desired outcome would obviate the need for such a toggle so I'll be closing this PR for now. Thanks again for the initial spike and sorry for taking such a long time to respond.

PR #6, opened 2016-05-19
[SPIKE] New mobiledoc based editor

kevinansfield:
requires TryGhost/Ghost#6255

Spike/prototype to test out new editor functionality based around https://github.com/bustlelabs/mobiledoc-kit

TODO:

*NB. This has expanded into a dumping ground for ideas and things that will need looking at moving forward, I expect a number of these will be pulled out into separate issues.*

- [x] Isolate existing editor
- [x] Add "New Editor" labs flag
- [x] Swap new editor in/out based on labs flag
- [x] Integrate mobiledoc editor component
  - [x] Initial `ember-mobiledoc-editor` integration
  - [x] Basic toolbar styles
  - [ ] Basic editor content area styles
- [x] Add `mobiledoc` field to posts table
- [x] Generate HTML from mobiledoc when post is saved on the server
- [x] Display appropriate editor and notification if opening a markdown/mobiledoc post that doesn't correspond with the current labs flag setting
- [x] [SimpleMDE](https://github.com/NextStepWebs/simplemde-markdown-editor) markdown card **(in progress)**
  - [x] Display simplemde interface when card is in edit mode
  - [x] Display rendered html when card is in display mode
  - [x] Lazy-load card assets
  - [x] Render simplemde markdown card during server rendering
- [ ] Improved UI for inserting cards
- [ ] Look into possibility of having discrete rich-text content blocks (at least from a UI perspective) instead of the whole document area acting as a rich-text editor
  - [ ] Remove always-present rich-text toolbar and only show it when editing a rich-text area
- [ ] CodeMirror card
- [ ] Ghost markdown card (adapt existing editor to work as a mobiledoc card)
- [ ] Method for cards to register themselves
  - [ ] Enumerate and instantiate ember component cards on client
  - [ ] Enumerate and instantiate server-side cards for rendering
  - [ ] Display dynamically loaded cards in the "insert card" UI
  - [ ] Investigate options for packaging cards as discrete units

ISSUES:
- [x] Mobiledoc's [Position](https://github.com/bustlelabs/mobiledoc-kit/blob/master/src/js/utils/cursor/position.js) class seems to have trouble when the content area is scrollable
- [ ] Cards have a space above & below that appear to be editable text but it's not possible to type unless pressing 'enter' to create a new paragraph
- [ ] Pressing bold/italic toolbar buttons does not immediately change mode - it's necessary to select some text and then press the button.
- [ ] When the cursor is at the end of a bold/italic section the corresponding toolbar button is not highlighted even though typing will continue to be bold/italic

Investigate:
- [ ] Options for converting existing markdown/html to mobiledoc
- [ ] Options for image cards


kevinansfield:
Closing as this has served it's purpose and we're now working on a new mobiledoc implementation.

PR #19, opened 2016-05-22
[WIP] Promise Buttons (with state!)

acburdine:
TODO:

- [ ] make buttons actually work
- [ ] tests
- [ ] evaluate which buttons need to be switched

### DIFF
--- a/app/components/gh-promise-button.js
+++ b/app/components/gh-promise-button.js
@@ -1,0 +1,157 @@
+import Ember from 'ember';
+import {strictInvokeAction, invokeAction} from 'ember-invoke-action';
+
+const {Component, computed, run} = Ember;
+
+/**
+ * Promise-enabled buttons. PromiseButton is a component that will take a promise,
+ * render as a button, and when clicked, will trigger the promise and then show its state
+ * once the promise has either resolved or rejected.
+ *
+ * Usage:
+ * ```handlebars
+ * {{#gh-promise-button class="btn btn-sm" promise=(action 'callServer')}}
+ *     Click Me!
+ * {{/gh-promise-button}}
+ * ```
+ *
+ * The `promise` variable is a closure action that must return a promise. If not,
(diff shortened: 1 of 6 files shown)


acburdine:
@kevinansfield re: your comment on the initial PR, I think I have a solution.

Would it be too much of a pain to simply set the currently working promise as an attribute on the same model/controller/what-have-you that the action being clicked by the button exists. Then the button could just observe it, and do something like this:

```js
{
    controllerPromise: computed.alias('promise'),
    triggerPromise: observer('controllerPromise', function () {
        this.get('controllerPromise').then(() => {
            return resolve(...args);
        }).catch(() => {
            return reject(...args);
        });
    })
}
```

The only issue I guess is if the promise is triggered by something other than the button, there's the chance that an error wouldn't be picked up by the button and instead be picked up by the direct promise handler. There could be a bunch of timing issues there.... 😞

kevinansfield:
I think we probably need to drop the idea of promise handling inside the button and instead have a system where state is passed in. It makes everything explicit and avoids weird behaviour when the upstream promise handling doesn't behave in a way that the button expects.

If every action that has a spin-button linked to it updates an associated `actionState` property to either `null`, `pending`, `fulfilled`, or `rejected` then the spin button can react accordingly. If it receives the `fulfilled` or `rejected` state then it can kick off it's own internal timer to return to it's default state once the result has been visible for long-enough.

We could have two components `spinButton` and `promiseButton` - `promiseButton` could be used where an action cannot be triggered externally, it would either wrap or extend `spinButton` and handle the state updates. However, I'm not sure if that would cause confusion over which to use where or if the consistency of using `spinButton` and `actionState` everywhere is preferable - I'm leaning towards the latter because a consistent state property would be useful outside of spin buttons.

My preference is definitely towards setting a simple state property rather than a promise because strings or other basic types are much easier to use in CPs than promises.

acburdine:
Closing to take a different approach

PR #25, opened 2016-05-25
Scheduler UI

AileenCGN:
refs TryGhost/Ghost#6413 and TryGhost/Ghost#6870

needs TryGhost/Ghost#6861

- **Post Settings Menu (PSM)**:'Publish Date' input accepts a date from now, min. 2 minutes to allow scheduler processing on the server. Also, there will always be some delay between typing the date and clicking on the 'Schedule Post' button. If the user types a future date for an already published post, the date will be reseted and he sees the message, that the post needs to be unpublished first. Once, the date is accepted, the label will change to 'Scheduled Date'.

- adds a CP 'timeScheduled' to post model, which will return `true` if the publish time is currently in the future.

- **Changes to the button flow in editor**:
- if the the CP `timeScheduled` returns true, a different drop-down-menu will be shown: 'Schedule Post' replaces 'Publish Now' and 'Unschedule' replaces 'Unpublish'.

- Covering the _edge cases_, especially when a scheduled post is about to be published, while the user is in the editor.
	- First, a new CP `scheduleCountdown` will return the remaining time, when the estimated publish time is 15 minutes from now. A notification with this live-ticker is shown next to the save button. Once, we reach a 2 minutes limit, another CP `statusFreeze` will return true and causes the save button to only show `Unschedule` in a red state, until we reach the publish time
	- Once the publish time is reached, a CP `scheduledWillPublish` causes the buttons and the existing code to pretend we're already dealing with a publish post. At the moment, there's no way to make a background-fetch of the now serverside-scheduled post model from the server, so Ember doesn't know about the changed state at that time.
	- Changes in the editor, which are done during this 'status freeze'-process will be saved back correctly, once the user hits 'Update Post' after the buttons changed back. A click on 'Unpublish' will change the status back to a draft.

- adds CP `isScheduled` for scheduled posts
- adds CP `offset` to component `gh-posts-list-item` and helper `gh-format-time-scheduled` to show schedule date in content overview.

TODOs:
- [x] new sort order for posts (1. scheduled, 2. draft, 3. published) (refs TryGhost/Ghost#6932)
- [ ] Write Issue to move posts sorting from posts controller to model and refactor to use `Ember.comparable` mixin and create own component for save button
- [x] Flows for draft -> scheduled -> published like described in TryGhost/Ghost#6870 incl. edge cases and button behaviour
- [x] Tests
- [x] new PSM behaviour for time/date in future
- [x] display publishedAt date with timezone offset on posts overview

### DIFF
--- a/app/controllers/posts.js
+++ b/app/controllers/posts.js
@@ -9,7 +9,7 @@ const {
 const {equal} = computed;
 
 // a custom sort function is needed in order to sort the posts list the same way the server would:
-//     status: ASC
+//     status: scheduled, draft, published
 //     publishedAt: DESC
 //     updatedAt: DESC
 //     id: DESC
@@ -32,7 +32,7 @@ function comparator(item1, item2) {
     }
 
     idResult = compare(parseInt(item1.get('id')), parseInt(item2.get('id')));
-    statusResult = compare(item1.get('status'), item2.get('status'));
+    statusResult = statusCompare(item1, item2);
     updatedAtResult = compare(updated1.valueOf(), updated2.valueOf());
     publishedAtResult = publishedAtCompare(item1, item2);
 
@@ -50,6 +50,38 @@ function comparator(item1, item2) {
     }
 
     return statusResult;
+}
+
+function statusCompare(item1, item2) {
+    let status1 = item1.get('status');
+    let status2 = item2.get('status');
+
+    // if any of those is empty
+    if (!status1 && !status2) {
+        return 0;
+    }
+
+    if (!status1 && status2) {
+        return -1;
+    }
+
+    if (!status2 && status1) {
+        return 1;
+    }
+
+    // We have to make sure, that scheduled posts will be listed first
+    // after that, draft and published will be sorted alphabetically and don't need
+    // any manual comparison.
+
+    if (status1 === 'scheduled' && (status2 === 'draft' || status2 === 'published')) {
+        return -1;
+    }
+
+    if (status2 === 'scheduled' && (status1 === 'draft' || status1 === 'published')) {
+        return 1;
+    }
+
+    return compare(status1.valueOf(), status2.valueOf());
 }
 
 function publishedAtCompare(item1, item2) {
--- a/app/models/post.js
+++ b/app/models/post.js
@@ -43,6 +43,7 @@ export default Model.extend(ValidationEngine, {
     config: service(),
     ghostPaths: service(),
     timeZone: service(),
+    clock: service(),
 
     absoluteUrl: computed('url', 'ghostPaths.url', 'config.blogUrl', function () {
         let blogUrl = this.get('config.blogUrl');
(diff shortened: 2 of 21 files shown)


app/controllers/posts.js
```diff
@@ -74,10 +74,10 @@ export default Controller.extend({
     postListFocused: equal('keyboardFocus', 'postList'),
     postContentFocused: equal('keyboardFocus', 'postContent'),
 
-    sortedPosts: computed('model.@each.status', 'model.@each.publishedAt', 'model.@each.isNew', 'model.@each.updatedAt', function () {
```
kevinansfield: The `model.` bit is still needed here, if you've seen the straight `@each` example somewhere it's likely from when controllers still proxied to the `model` property but Ember hasn't done that for a long time. However, you can use the brace expansion syntax to make it more concise:

```javascript
sortedPosts: computed('model.@each.{isNew,publishedAt,status,updatedAt}', function () {
```

AileenCGN: Ahhh... sorry... Didn't but it back.. I did some changes but reverted them before pushing. Those, I forgot...

app/models/post.js
```diff
@@ -68,6 +69,15 @@ export default Model.extend(ValidationEngine, {
 
     isPublished: equal('status', 'published'),
     isDraft: equal('status', 'draft'),
+    isScheduled: equal('status', 'scheduled'),
+
+    // Checks every second, if we reached the scheduled date
+    timeScheduled: computed('publishedAt', 'clock.second', function () {
```
kevinansfield: I'm not sure this is a good place to handle this - it will mean we have a task running every second for every blog post that is in the store.

I think perhaps a posts list component would be a better place for it - that way it can track only the posts that are rendered on the screen and should be able to set up tracking only for those posts which are scheduled.

Leave it as-is for now, this is more of a note for an area to re-visit later.

ErisDS:
Not sure what the various cases at work here are, but I'm seeing this:



I'm expecting that the message should be updating at the same time as the button, to say "Unschedule to make changes now" after the 2mins message?

Otherwise, I think this is working pretty darned nicely for a v1. I have 1 suggestion that, whatever it is that changes the button & message once the post is actually published (just a timer?) should perhaps either  update the message to say "The post has been published" or manually show a notification (the one that says "view post") if that is at all possible? 

If these things are hard, we can merge without and try to add them later.

AileenCGN:
Yes! Sorry, didn't update the spec now. @JohnONolan and I changed it, so it will display this:

image

between 15 and 2 minutes of publishing, and this:

image

two minutes before. Will update the spec now!

AileenCGN:
Regarding this:

>Otherwise, I think this is working pretty darned nicely for a v1. I have 1 suggestion that, whatever it is that changes the button & message once the post is actually published (just a timer?) should perhaps either update the message to say "The post has been published" or manually show a notification (the one that says "view post") if that is at all possible?

What about, changing the notification, we have already? Because toaster notification don't work at the moment. Everything is a big workaround-hack. But yeah, it's basically a timer, which signals, that the post got published.

AileenCGN:
Ok, I made a notification like this:

image

This stays, until the user either refreshes the page or clicks on the save button.

ErisDS:
Cool - might be worth confirming the wording of that with @JohnONolan if you haven't already?

AileenCGN:
I had, and I will change it to "Post has been published."

AileenCGN:
I changed it again. The handling of this notification wasn't really nice. Now, the user will get a regular 'toaster'-notification. Is that ok for you? @ErisDS, @JohnONolan 

image

I think it's also more consistent?

PR #29, opened 2016-05-30
Timezones finish up

AileenCGN:
closes TryGhost/Ghost#6406

follow-up PR of #2

- adds a `timeZone` Service to provide the offset (=timezone reg. moment-timezone) of the users blog settings
- `gh-datetime-input` will read the offset of the timezone now and adjust the `publishedAt` date with it. This is the date which will be shown in the PSM 'Publish Date' field. When the user writes a new date/time, the offset is considered and will be deducted again before saving it to the model. This way, we always work with a UTC publish date except for this input field.
- gets `availableTimezones` from `configuration/timezones` API endpoint
- adds a `moment-utc` transform on all date attr (`createdAt`, `updatedAt`, `publishedAt`, `unsubscribedAt` and `lastLogin`) to only work with UTC times on serverside
- when switching the timezone in the select box, the user will be shown the local time of the selected timezone
- `createdAt`-property in `gh-user-invited` returns now `moment(createdAt).fromNow()` as `createdAt` is a moment date already
- added clock service to show actual time ticking below select box
- default timezone is '(GMT) Greenwich Mean Time : Dublin, Edinburgh, London'
- if no timezone is saved in the settings yet, the default value will be used
- shows the local time in 'Publish Date'  in PSM by default, until user overwrites it
- adds dependency `moment-timezone 0.5.4` to `bower.json`

---------

**Tests:**

- sets except for clock service in test env
- adds fixtures to mirage
- adds `service.ajax` and `service:ghostPaths` to navigation-test.js
- adds unit test for `gh-format-timeago` helper
- updates acceptance test `general-setting`
- adds acceptance test for `editor`
- adds integration tests for `services/config` and `services/time-zone`

---------

**Todos:**

- [ ] Integration tests: ~~`services/config`~~, ~~`services/time-zone`~~, `components/gh-datetime-input`
- [x] Acceptance test: `editor`
- [ ] Unit tests: `utils/date-formatting`
- [ ] write issue for renaming date properties (e. g. `createdAt` to `createdAtUTC`) and translate those for server side with serializers

### DIFF
--- a/app/components/gh-datetime-input.js
+++ b/app/components/gh-datetime-input.js
@@ -4,7 +4,11 @@ import boundOneWay from 'ghost/utils/bound-one-way';
 import {formatDate} from 'ghost/utils/date-formatting';
 import {invokeAction} from 'ember-invoke-action';
 
-const {Component} = Ember;
+const {
+    Component,
+    RSVP,
+    inject: {service}
+} = Ember;
 
 export default Component.extend(TextInputMixin, {
     tagName: 'span',
@@ -14,15 +18,21 @@ export default Component.extend(TextInputMixin, {
     inputClass: null,
     inputId: null,
     inputName: null,
+    timeZone: service(),
 
     didReceiveAttrs() {
-        let datetime = this.get('datetime') || moment();
+        let promises = {
+            datetime: RSVP.resolve(this.get('datetime') || moment.utc()),
+            offset: RSVP.resolve(this.get('timeZone.offset'))
+        };
 
         if (!this.get('update')) {
             throw new Error(`You must provide an \`update\` action to \`{{${this.templateName}}}\`.`);
         }
 
-        this.set('datetime', formatDate(datetime));
+        RSVP.hash(promises).then((hash) => {
+            this.set('datetime', formatDate(hash.datetime || moment.utc(), hash.offset));
+        });
     },
 
     focusOut() {
--- a/app/controllers/post-settings-menu.js
+++ b/app/controllers/post-settings-menu.js
@@ -31,6 +31,7 @@ export default Controller.extend(SettingsMenuMixin, {
     notifications: service(),
     session: service(),
     slugGenerator: service(),
+    timeZone: service(),
 
     initializeSelectedAuthor: observer('model', function () {
         return this.get('model.author').then((author) => {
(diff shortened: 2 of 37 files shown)


app/templates/posts.hbs
```diff
@@ -24,8 +24,8 @@
                                         {{#if post.page}}
                                             <span class="page">Page</span>
                                         {{else}}
-                                            <time datetime="{{post.publishedAtOffset}}" class="date published">
-                                                Published {{gh-format-timeago post.publishedAt}}
+                                            <time datetime="{{post.publishedAt}}" class="date published">
+                                                Published {{gh-format-timeago post.publishedAt}} ago
```
ErisDS: I don't think adding "ago" here is the correct fix?
The word wasn't there before, and according to the [moment docs](http://momentjs.com/docs/#/displaying/fromnow/), it should be being output by `fromNow()`. This bug is super weird, it looks like there is something going on which may be affecting moment's locale strings?

AileenCGN: Found the bug: I used `fromNow(moment.utc())` although `fromNow()` only excepts a boolean value. Changed it to `from(moment.utc())` as described [here](http://momentjs.com/docs/#/displaying/from/) in the moment docs and it works fine now.

app/controllers/post-settings-menu.js
```diff
@@ -292,50 +293,57 @@ export default Controller.extend(SettingsMenuMixin, {
             if (!userInput) {
                 // Clear out the publishedAt field for a draft
                 if (this.get('model.isDraft')) {
-                    this.set('model.publishedAt', null);
+                    this.set('model.publishedAtOffset', null);
```
kevinansfield: `publishedAtOffset` is a computed property so you don't want to set it directly (it will overwrite the CP).

If you want `publishedAtOffset` to return `null` I would suggest ensuring it returns null and doesn't error when `publishedAt` is `null`, then reverting this change so that we continue clearing the `publishedAt` value.

app/services/time-zone.js
```diff
@@ -0,0 +1,30 @@
...
+    _parseTimezones(settings) {
+        let activeTimezone = settings.get('activeTimezone');
+        let offset = activeTimezone;
+
+        return offset;
+    },
+
+    _settings: computed(function () {
+        let store = this.get('store');
+        return store.queryRecord('setting', {type: 'blog'});
```
kevinansfield: This will need to be updated to query `{type: 'blog,theme,private'}` so that we don't run the risk of other areas of the admin saving incomplete settings data.

I'd like to make this problem go away by creating a `settings` service which handles loading/refreshing/saving the blog settings but that's something for a different PR 😄

app/utils/date-formatting.js
```diff
@@ -29,7 +29,10 @@ function parseDateString(value, offset) {
     // in UTC timezone
     moment.tz.setDefault(offset);
 
-    return value ? moment(verifyTimeStamp(value), parseDateFormats, true) : undefined;
```
kevinansfield: Would you be better off here making sure that `post.publishedAtOffset` returns `null` when `publishedAt` is null? 

It seems to me the difference between `publishedAt` and `publishedAtOffset` behaviour is now leaking to external components such as this date formatting util which has now also lost the ability to display nothing for empty values.

app/utils/date-formatting.js
```diff
@@ -24,8 +24,15 @@ function verifyTimeStamp(dateString) {
...
+function parseDateString(value, offset) {
+    // We need the offset here, otherwise the date will be parsed
+    // in UTC timezone
+    moment.tz.setDefault(offset);
+
+    // If the user clears out the date input field, the value will be a moment by default. In that case,
+    // verifyTimeStamp() runs into an error. Therefore, we will just return the value in the parseDateFormats
+    // format.
+    return moment.isMoment(value) ? moment(value, parseDateFormats, true) : moment(verifyTimeStamp(value), parseDateFormats, true);
```
AileenCGN: > Would you be better off here making sure that post.publishedAtOffset returns null when publishedAt is null?

@kevinansfield did you see the change to that? So you reckon, I should return null instead of the modified (without `verifyTimeStamp()` return value?

kevinansfield: I think in the post model the `publishedAtOffset` CP should return `null` when it's dependent property `publishedAt` is `null`. Currently it always returns a moment value no matter what it's dependent property value is.

If that change is made then the behaviour of `parseDateString` can be put back to how it was before:
```
return value ? moment(verifyTimeStamp(value), parseDateFormats, true) : undefined;
```

app/models/post.js
```diff
@@ -63,6 +64,19 @@ export default Model.extend(ValidationEngine, {
     scratch: null,
     titleScratch: null,
 
+    // Computed Date properties
+    // timeZone.offset service returns the moment-timezone from the
+    // activeTimezone, set in the blogs settings
+
+    publishedAtOffset: computed('timeZone.offset', 'publishedAt', function() {
```
kevinansfield: I'm wondering if `publishedAtOffset` is actually needed at the model layer - right now it's only being used to pass its value into `gh-datetime-input` component which then adjusts it to match the blog timezone anyway so we could simply pass the UTC `publishedAt` value to the component instead.

I think removing it from the model layer will help to keep any offset adjustments outside of the main code and in the display layer. The `*At`->`*AtUTC` rename would also become cleaner as it will only be the core model attributes changing name to make it clear that we always deal with UTC internally.

app/components/gh-datetime-input.js
```diff
@@ -4,7 +4,12 @@ import boundOneWay from 'ghost/utils/bound-one-way';
 import {formatDate} from 'ghost/utils/date-formatting';
 import {invokeAction} from 'ember-invoke-action';
 
-const {Component} = Ember;
+const {
+    Component,
+    RSVP,
+    inject: {service},
+    computed
```
kevinansfield: `computed` is defined but never used

tests/integration/services/time-zone-test.js
```diff
@@ -0,0 +1,52 @@
...
+            type: 'blog',
+            key: 'activeTimezone',
+            value: 'Africa/Cairo'
+        }
+    ];
+
+    server.get('/ghost/api/v0.1/settings/', function () {
+        return [200, {'Content-Type': 'application/json'}, JSON.stringify({settings})];
+    });
+};
```
kevinansfield: This semi-colon is unnecessary

app/mirage/config.js
```diff
@@ -207,6 +207,31 @@ export function testConfig() {
...
+    this.put('/posts/:id', function (db, request) {
+        let [attrs] = JSON.parse(request.requestBody).posts;
+        let post;
+
+        if (isBlank(attrs.slug) && !isBlank(attrs.title)) {
+            attrs.slug = attrs.title.dasherize();
+        }
+
+        // NOTE: this does not use the post factory to fill in blank fields
+        post = db.posts.insert(attrs);
```
kevinansfield: As this is a `put` request we want to be updating the post rather than inserting a new one. Example following the `this.put('/tags/:id', ...` route further down:

```
        let {id} = request.params;
        let [attrs] = JSON.parse(request.requestBody).tags;
        let post = db.posts.update(id, attrs);

        return {
            posts: [post]
        };
```

AileenCGN: But the validation here, shall stay?
```        if (isBlank(attrs.slug) && !isBlank(attrs.title)) {
            attrs.slug = attrs.title.dasherize();
        }```

kevinansfield: No, that's just replicates something the server does when a new post is created, I think we can leave it out for updates.

PR #34, opened 2016-06-01
deps: validator-js@5.3.0

kevinansfield:
no issue
- brings client-side validator.js version up-to-date with server-side version
- move `validator-js` from a bower dependency to an npm dependency
- remove `validator` global from `.jshintrc`
- update all uses of `validator.*` to use ES6 modules (should cut down on the amount of code imported)
- remove custom validator extensions
  - uses of `validator.empty` have been updated to use Ember's `isBlank` that is imported as if it was an ES6 module via `ember-cli-shims` (we should probably start using `ember-cli-shims` everywhere as it paves the way for migrating to true ES6 modules once Ember has been officially split up)

This is currently failing but I haven't had a chance to properly debug yet. I believe it may be something related to https://github.com/chriso/validator.js/pull/496#discussion-diff-52817638


acburdine:
@kevinansfield The validations refactor _should_ (I hope) remove the need for the validator.js library.

ErisDS:
If validator.js is removed, how will the client side do things like determining if a URL is valid?

acburdine:
@ErisDS there's validation ulitities built into ember-changeset-validations :smile:

ErisDS:
The benefit to using validator.js on both the client and server is that you can be sure they'll both permit/reject the same things. So for complex things like email addresses and urls, there is no mis-match between what is accepted.

I'm just flagging this up as something to consider. If you run into issues with edge cases, it may be better to have custom validators wrap validator.js?

acburdine:
👍  yeah true, will definitely re-evaluate as the refactor goes farther along.

acburdine:
going to close this as it's being handled in the validation refactor

PR #37, opened 2016-06-03
Update ember-cli-mirage to 0.2.4

acburdine:
supersedes #32 

Updates ember-cli-mirage to version 0.2.0
- [x] ensure lodash is updated (remove from bower?)
- [x] update api
- [x] make tests pass


acburdine:
@kevinansfield With Mirage 0.2 comes a lot of improvements on model relationships, especially where sideloading/embedding is concerned. Should this also target that area of the mirage config and mimic the actual API behavior?

I'm thinking primarily of doing things like taking `GET /posts/?include=tags,author` and making that actually affect the mirage behavior.

kevinansfield:
Yeah, my plan was to completely switch over to the new mirage ORM and mimic as much of the API behaviour as possible. I tried it on a side project and had issues with many to many relationships but I don't think we'll have the same issues in ghost as our setup is a bit different.

acburdine:
@kevinansfield Still running into some various minor acceptance test issues, but could still use a review at some point (no rush though 😉)

acburdine:
@kevinansfield it's green now 🎉

mirage/utils.js
```diff
@@ -0,0 +1,76 @@
...
+export function versionMismatchResponse() {
+    return new Response(400, {}, {
+        errors: [{
+            errorType: 'VersionMismatchError'
+        }]
+    });
+}
+
+// minimalist deparam implementation
+export function deparam(string) {
```
acburdine: @kevinansfield I added this and removed the `jquery-deparam` dependency because Mirage didn't have access to it for some reason (may have been because I was using it in dev and not test mode), but regardless this was enough for our tests, so I think it's probably good to stay.

kevinansfield: Yeah, `jquery-deparam` is only pulled in for testing at the moment. I think it may be better to include it on all builds as it's a pretty small dependency and there are a few places we do some not so robust checks for url params. Hopefully Ember's router will expose it's param handling at some point.

I think the main issue with the minimalist implementation here is that it doesn't cover the `?foo[]=bar` type syntax - I'm not sure we're using that anywhere at the moment but it's quite likely to crop up with post filtering.

acburdine: hmm...ok. Do we want to add it back in if we need it in the future then, or should I just add it back in for all environments now?

kevinansfield: I'd say add it back in for all environments, it would be good to standardise on one library.

I'm aware that we have a hacky regex to parse a query param in [at least one place](https://github.com/TryGhost/Ghost-Admin/blob/lts/app/routes/setup.js#L31), it's probably worth an audit to make sure we're always using `deparam` when manually grabbing query parameters.

PR #38, opened 2016-06-03
Add version header to API requests

kevinansfield:
no issue
- modifies the version info included in `env.APP.version` to only include the `major.minor` version numbers
- update base adapter to include `X-Ghost-Version` header
- update `ajax` service to include `X-Ghost-Version` header

### DIFF
--- a/app/adapters/base.js
+++ b/app/adapters/base.js
@@ -2,6 +2,7 @@ import Ember from 'ember';
 import RESTAdapter from 'ember-data/adapters/rest';
 import ghostPaths from 'ghost/utils/ghost-paths';
 import DataAdapterMixin from 'ember-simple-auth/mixins/data-adapter-mixin';
+import config from 'ghost/config/environment';
 
 const {
     inject: {service}
@@ -14,6 +15,10 @@ export default RESTAdapter.extend(DataAdapterMixin, {
     namespace: ghostPaths().apiRoot.slice(1),
 
     session: service(),
+
+    headers: {
+        'X-Ghost-Version': config.APP.version
+    },
 
     shouldBackgroundReloadRecord() {
         return false;
--- a/app/services/ajax.js
+++ b/app/services/ajax.js
@@ -1,6 +1,7 @@
 import Ember from 'ember';
 import AjaxService from 'ember-ajax/services/ajax';
 import {AjaxError} from 'ember-ajax/errors';
+import config from 'ghost/config/environment';
 
 const {inject, computed} = Ember;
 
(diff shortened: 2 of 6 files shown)


ErisDS:
Sorry, I didn't get here quick enough!

I think we pass ember the safe version, I think it's even called `safeVersion` - not sure if that makes sense to use, but if it saves doing the processing everywhere, it's probably worthwhile?

Also, I wanted to discuss whether we should be using `X-Ghost-Version`, which was my original suggestion, or, having read [this](https://www.troyhunt.com/your-api-versioning-is-wrong-which-is/) yesterday, whether we should actually use an Accepts header?

kevinansfield:
> I think we pass ember the safe version, I think it's even called safeVersion - not sure if that makes sense to use, but if it saves doing the processing everywhere, it's probably worthwhile?

The issue I saw is that ember-cli by default will read the `package.json` file and include the full version info into the built files, this overrides that behaviour so that we never leak the patch version. This approach also breaks a tie between the server-generated index.html and the client which I think fits better with the client/server split?

> Also, I wanted to discuss whether we should be using X-Ghost-Version, which was my original suggestion, or, having read this yesterday, whether we should actually use an Accepts header?

To me the `Accepts` header is more about mime-types and what formats the client will accept back from the server. Maybe it makes sense if you are changing from a SOAP protocol to a JSON protocol but I an explicit version header better conveys our use case for version mismatches?

kevinansfield:
Alternatively, we already have a request parameter in the form of `/api/v0.1/*` - should we be utilising that? If the version number there is increased to `v0.2` then the API could return a version mismatch error for any request it receives on older version numbers and would also cover non Ghost-Admin API users.

ErisDS:
@kevinansfield all good points.

RE: the safe version, I thought you might be able to take it from the server, rather than have to do the calculation yourself. 

I think I understand what you're saying about keeping the two more separate and that makes sense, however in that case, I believe the logic on the client needs to be the same as on the server:

The regex in:

`version.match(/^(\d+\.)?(\d+)/)[0]`

explicitly grabs only the first two numbers separated by a dot.

Whereas:

`version.replace(/\.\d+$/, '')` assumes the last pattern in the version string `.number` when in fact it might be `-rc1` for example? Perhaps the ember CLI version doesn't include that part already, but it if  it does, we need to account for it.

---

RE: headers & versioning

In a sense, the api version in the URL is BS. We make changes all the time and we don't change the version because if we did, all requests would fail because it changes the location of the resource... which as discussed in the article I linked to.. seems really wrong in hindsight.

Ideally, long term I'd like to use accepts headers to control what the API actually returns, so we might have this header to get our format:

`Accept: application/vnd.ghost+json`

And this one, to get a proper JSONAPI format: 
`Accept: application/vnd.api+json`

This just changes the format though - not the version of the API.

At the moment the version of the API is intrinsically linked to the version of Ghost in ways we don't truly have much control over because it's all baked deep into the model layer. You can almost guarantee subtle differences in the response between each version of Ghost, and that's part of the reason why we still have the API In beta.

To take the API out of beta, we'll be removing the version string from the API - you may still send that format of URL to the API, but it will process it out.

Instead, we'll need a different versioning strategy for the API and I don't know what that's going to look like yet. 

It's particularly tricky because the endpoints and data available can be different dependent on each blog's configuration, so it's not like you can build one simple client based on an API version, you actually need to inspect information about the server that goes beyond a simple version - like if subscribers is enabled.

For now what we need is a mismatch error sent if a request specifies a header and the version in the header is different to that on the server. Using it is optional, it won't impact what version of data you get back and it doesn't interfere with long term plans for an API version which would be a different thing.

TLDR; it's totally fine as it is

PR #44, opened 2016-06-04
Refactor signin/signup/setup/reset routes to use correct route mixin

acburdine:
no issue
- aligns those routes more with the ember-simple-auth method of doing things

### DIFF
--- a/app/routes/reset.js
+++ b/app/routes/reset.js
@@ -1,20 +1,20 @@
 import Route from 'ember-route';
 import injectService from 'ember-service/inject';
-import Configuration from 'ember-simple-auth/configuration';
+import UnauthenticatedRouteMixin from 'ember-simple-auth/mixins/unauthenticated-route-mixin';
 import styleBody from 'ghost-admin/mixins/style-body';
 
-export default Route.extend(styleBody, {
+export default Route.extend(styleBody, UnauthenticatedRouteMixin, {
     classNames: ['ghost-reset'],
 
     notifications: injectService(),
     session: injectService(),
 
     beforeModel() {
-        this._super(...arguments);
         if (this.get('session.isAuthenticated')) {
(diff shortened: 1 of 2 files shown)


tests/acceptance/setup-test.js
```diff
@@ -29,17 +31,7 @@ describe('Acceptance: Setup', function () {
...
-            expect(currentURL()).to.equal('/');
-        });
-
-        visit('/setup/two');
-        andThen(() => {
-            expect(currentURL()).to.equal('/');
-        });
-
-        visit('/setup/three');
+        visit('/setup');
```
acburdine: So there's a bug in the Ember andThen helper apparently... I kept running into a failure here where successive `visit` calls in the same `it` call seem to not follow async behavior correctly.

I tried turning on route transition logging, and Ember was attempting to transition to `/` correctly, but the `andThen` was resolving before that transition could be completed. This only happened after the first `visit`/`andThen` combination.

@kevinansfield I know it's bad practice to change this type of assertion in testing, but wasn't sure how else to fix this. I'm not 100% sure why the route change makes it not work 😞

acburdine: After more investigation it appears that because ember-simple-auth calls `transition.abort` in the beforeModel before calling `transitionTo`, it causes the glitch in the testing. Going to open an issue in the relevant repositories.

acburdine:
currently blocked on simplabs/ember-simple-auth#992

acburdine:
Note, the simple-auth PR was merged, and I've updated this to depend on the master branch of ember-simple-auth just for testing purposes. Will change once a new version of esa is released.

kevinansfield:
@acburdine the ESA 1.2.0 upgrade has been merged but I think this PR needs to be re-worked a bit before it can be merged. For example, the `UnauthenticatedRouteMixin` is no longer directly suitable for the setup routes because we have different behaviour depending on the authentication type that is being used.

PR #45, opened 2016-06-05
[WIP] Stateful spin buttons (take 2)

acburdine:
supersedes #19 

This is an initial experimentation into adding resolved/rejected state to the spin-button component.


acburdine:
@kevinansfield This currently is implemented in the editor, and works well. This seems to be a bit of a simpler approach 😉

acburdine:
Closing this as there is a better approach to be made with this using ember-concurrency.

PR #51, opened 2016-06-08
Check for Ghost Desktop manual update

felixrieseberg:
If the user is running a version of Ghost Desktop that requires a manual update, we display a little warning message referring to a blog post on dev.ghost.org. If the `userAgent` does not include `ghost-desktop`, we obviously do nothing.

cc @JohnONolan @acburdine @kevinansfield

### DIFF
--- a/app/routes/application.js
+++ b/app/routes/application.js
@@ -96,6 +96,7 @@ export default Route.extend(ApplicationRouteMixin, ShortcutsRoute, {
         signedIn() {
             this.get('notifications').clearAll();
             this.send('loadServerNotifications', true);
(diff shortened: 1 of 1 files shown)


ErisDS:
I'm just reading the history of this!

ErisDS:
@felixrieseberg just want to check... is this what you expect?



https://regex101.com/r/uN5tQ2/2

Seems like it might be missing some versions to me

felixrieseberg:
As crazy as it sounds, those are the correct versions. All other ones either update themselves or received other notifications.

PR #57, opened 2016-06-10
Add UTC timezone as default in settings model

AileenCGN:
no issue

- Changes the default timezone from 'Europe/Dublin' to 'Etc/UTC' in setting model
- Updates acceptance test for settings general
- adds mirage fixtures values


acburdine:
@AileenCGN you need to modify `app/mirage/fixtures/timezones.js` as well 😉

AileenCGN:
Thanks, @acburdine! It's good now :wink:

PR #58, opened 2016-06-10
internal tags feature

acburdine:
refs TryGhost/Ghost#6165
- change behavior to use 'visibility' property
- add tests


kevinansfield:
@acburdine this needs a rebase but otherwise 👍

acburdine:
@kevinansfield done :)

PR #61, opened 2016-06-11
Update grunt-jscs and ember-suave to 3.0.0

acburdine:
replaces #41, #60
- update ember-suave and grunt-jscs to 3.0
- standardize Ember global de-structuring rules across app & tests


package.json
```diff
@@ -26,6 +26,7 @@
   "devDependencies": {
     "broccoli-asset-rev": "2.4.2",
     "csscomb": "3.1.8",
+    "cst": "0.3.0",
```
acburdine: @kevinansfield I didn't appear to need this locally, but without this addition it kept failing on Travis.

PR #65, opened 2016-06-14
Error handling refactor

kevinansfield:
closes https://github.com/TryGhost/Ghost/issues/6974
- update "change password" fields/process to use inline validations
- remove `notifications.showErrors` and update all uses of it to `showAPIError`
- display multiple API errors as alerts rather than toaster notifications
- refactor `notifications.showAPIError`
  - remove `notifications.showErrors`, use a loop in `showAPIError` instead
  - properly determine the message from `AjaxError` or `AdapterError` objects
  - determine a unique key if possible so that we don't lose multiple different alerts
- add `ServerUnreachable` error for when we get a status code of 0 (eg, when the ghost service has been shut down)
- simplify error messages for our custom ajax errors

---
TODO:
- [x] update password change validation to use inline validation
  - [x] update tests
- [x] remove all calls to `notifications.showErrors` (rename to `_showErrors` to signify it being an internal method)
  - [x] update `._showErrors` to use alerts instead of toaster notifications
- [x] handle standard ember-ajax errors in `notifications.showAPIError`
- [x] handle server-returned error messages in `notifications.showAPIError`
- [x] update error display tests
- [x] remove/improve use of `getRequestErrorMessage`
- [x] no error message when attempting to edit a post when server is offline


app/services/notifications.js
```diff
@@ -1,14 +1,11 @@
...
-    computed: {filter},
-    A: emberA,
-    get,
-    set,
-    isArray
-} = Ember;
+import {isBlank} from 'ember-utils';
+import {A as emberA, isEmberArray} from 'ember-array/utils';
+import {filter} from 'ember-computed';
+import {dasherize} from 'ember-string';
```
kevinansfield: I started testing the move to modularisation through [ember-cli-shims](https://github.com/ember-cli/ember-cli-shims), working well so far, if it pans out it would be good to get a single PR put together that updates all of the imports.

acburdine: I can take care of that at some point soon.

kevinansfield:
Ready for review 😃

PR #72, opened 2016-06-16
Renaming date properties to contain `UTC`

AileenCGN:
closes TryGhost/Ghost#6985

- renames all date properties to end with `UTC`:
	- `publishedAt` in `post` model
	- `createdAt` in `post`, `role`, `subscriber`, `tag` and `user` model
	- `updatedAt` in `post`, `role`, `subscriber`, `tag` and `user` model
	- `unsubscribedAt` in `subscriber` model
	- `lastLogin` in `user` model
- adds `attrs` transforms in matching serializers `post`, `tag` and `user`
- new serializers files for `subscribers` and `role` to add `attr` transforms
- adds unit tests for all serializers


app/controllers/subscribers.js
```diff
@@ -108,10 +108,13 @@ export default Controller.extend(PaginationMixin, {
 
         sortByColumn(column) {
             let table = this.get('table');
+            let valuePath = column.get('valuePath');
+
+            valuePath = column.get('valuePath') === 'createdAtUTC' ? 'created_at' : column.get('valuePath').trim().underscore();
```
AileenCGN: Not sure, how we could fix this... `valuePath` in `columns` is `createdAtUTC` now. That means, everytime `this.get('valuePath').trim().underscore()` was done, the property name changed to `created_at_utc` which, of course, is not correct for the server request.

I tried to work with an `alias`, but then the `columns` build would error... I couldn't figure out, if it's because of `ember-light-table` or if there's a better way to fix that... 

Happy for some input/ideas here! 😃

app/controllers/subscribers.js
```diff
@@ -108,10 +108,13 @@ export default Controller.extend(PaginationMixin, {
 
         sortByColumn(column) {
             let table = this.get('table');
+            let valuePath = column.get('valuePath');
+
+            valuePath = valuePath === 'createdAtUTC' ? 'created_at' : valuePath.trim().underscore();
```
AileenCGN: Not sure, how we could fix this... `valuePath` in `columns` is `createdAtUTC` now. That means, everytime `this.get('valuePath').trim().underscore()` was done, the property name changed to `created_at_utc` which, of course, is not correct for the server request.

I tried to work with an `alias`, but then the `columns` build would error... I couldn't figure out, if it's because of `ember-light-table` or if there's a better way to fix that... 

Happy for some input/ideas here! 😃

kevinansfield: Would `order: column.get('valuePath').trim().replace(/UTC$/, '').underscore()` work?

AileenCGN: It does work! 👍

kevinansfield: Cool, all this bit of code is doing is setting the query params (`.../subscribers/?order=created_at&direction=asc`) - right now those query params are passed directly to the server so need to match the server-side column names.

In this case a little bit of string manipulation to remove the `UTC` part inline is fine, if the logic get more complex or needs to be repeated in many areas then we can look at modifying the adapter so our URLs can use the client's column names but the adapter converts them to server column names when making requests to the API.

app/controllers/post-settings-menu.js
```diff
@@ -290,39 +290,39 @@ export default Controller.extend(SettingsMenuMixin, {
...
                 }
                 return;
             }
 
             // The user inputs a date which he expects to be in his timezone. Therefore, from now on
             // we have to work with the timezone offset which we get from the timeZone Service.
             this.get('timeZone.blogTimezone').then((blogTimezone) => {
-                let newPublishedAt = parseDateString(userInput, blogTimezone);
-                let publishedAt = moment.utc(this.get('model.publishedAt'));
+                let newPublishedAtUTC = parseDateString(userInput, blogTimezone);
```
kevinansfield: This may be a little confusing as we get date from `parseDateString` that is actually in the blog's timezone, not in UTC. We actually convert it into UTC further down on line 319.

I would suggest two variables here, one that holds the blog-timezone date and one that holds the UTC value.

AileenCGN: Makes sense! 👍

PR #89, opened 2016-06-22
Use node 0.12 for lint testing (temporary)

acburdine:
To be used until https://github.com/jscs-dev/node-jscs/issues/2279 is resolved


acburdine:
It's fixed now, closing.

PR #90, opened 2016-06-24
Grunt buildAboutPage tooling

acburdine:
refs TryGhost/Ghost#6977

- move buildAboutPage task to client gruntfile

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,11 +5,15 @@ language: node_js
 node_js:
   - "0.10"
 
+env:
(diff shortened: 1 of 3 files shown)


ErisDS:
This is updating the about task as well as moving it, correct?

acburdine:
@ErisDS yeah that specific about task handles input from both Ghost and Ghost Admin

acburdine:
@kevinansfield This is good for review

PR #91, opened 2016-06-25
fix meta-data problem, which will cause SEO of post fail.

starcwl:
fix meta-data input field action calls

no issue
- fixed focus-out action calls in post-settings-menu meta fields which miss arguments metaDescriptionScratch and metaTitleScratch.

### DIFF
--- a/app/templates/post-settings-menu.hbs
+++ b/app/templates/post-settings-menu.hbs
@@ -121,14 +121,14 @@
             <form {{action "discardEnter" on="submit"}}>
             {{#gh-form-group errors=model.errors property="metaTitle"}}
                 <label for="meta-title">Meta Title</label>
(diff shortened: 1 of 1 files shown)


acburdine:
hey @starcwl, thanks for catching this!

It looks good to merge for me, but would you be willing to update the commit message a bit?

Something like:

```
fix meta-data input field action calls

no issue
- fixed focus-out action calls in post-settings-menu meta fields
```

(you don't have to use that exact wording, but it would be nice to have something more like [this](https://github.com/TryGhost/Ghost/wiki/Git-workflow#commit-messages).)

Thanks! 😀

starcwl:
@acburdine thanks your advice. 😄

acburdine:
@starcwl Can you change _this_ commit to use the standard commit format? Sorry If that wasn't clear 😄

starcwl:
@acburdine  I have update the commit.  Does it ok now?

acburdine:
:+1:

PR #94, opened 2016-06-29
update bower deps and start switch from bower to npm

kevinansfield:
no issue
- `ember-cli-node-assets@0.1.3`
  - allows npm module assets to be loaded in a similar way to bower assets
  - switching to npm deps means we get automated greenkeeper updates
- `codemirror@5.16.0` && bower->npm
- `ember@2.6.1`
- `jquery-deparam@0.5.2` && bower->npm
- `moment` bower->npm
- `moment-timezone` bower->npm
- `password-generator` bower->npm
- `blueimp-md5` bower->npm

This still leaves some bower dependencies, reasons for this include:
- the dependency only provides a bower component
- the dependency is pulled in through bower via another dependency (notably the `ember*`, `pretender`, `lodash`, and `Faker` deps)
- the dependency will be removed soon (`jquery-ui`, `jquery-file-upload`)
- the switch didn't work
  - `validator.js` see #34
  - `device.js` uses incompatible module code in it's npm build
  - `keymster.js` fails due to missing `.toUpperCase()`

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -10,7 +10,6 @@
     "Faker": "3.1.0",
     "fastclick": "1.0.6",
     "google-caja": "6005.0.0",
-    "jquery-deparam": "0.5.1",
     "jquery-file-upload": "9.12.3",
     "jquery-ui": "1.11.4",
     "jqueryui-touch-punch": "furf/jquery-ui-touch-punch#4bc009145202d9c7483ba85f3a236a8f3470354d",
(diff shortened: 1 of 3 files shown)


kevinansfield:
⚡️ready for review

Interesting possibility with `ember-cli-node-assets` is that it can place files in the public dir, this could be useful with Codemirror to enable lazy-loading and decrease the core JS file size.

acburdine:
LGTM 👍

PR #99, opened 2016-06-30
lazy-load codemirror on code injection screen

acburdine:
refs TryGhost/Ghost#6149
- concats codemirror.js on build, lazy-loads using getScript on settings/code-injection page

TODO:
- [x] figure out how much the filesize diff is
- [x] get production outputting uglified js correctly
- [x] disable separate builds during tests?
- [x] 💃


acburdine:
@kevinansfield I went ahead and left the styles in the CSS output, because:

- I wasn't sure how to reliably lazy-load CSS
- they're small so it shouldn't make much of a difference

kevinansfield:
> I wasn't sure how to reliably lazy-load CSS
It's usually as simple as inserting a style tag. If you check the mobiledoc editor spike the codemirror card has an example - if I remember correctly you can give the style tag an id so you can check that to ensure you don't double-load. However as you say it's small so not much point worrying about it.

One thing that is useful is using `ember-ajax` rather than straight jquery so that the loading is run loop aware and you don't need to wrap in another promise. Looks like a good start though! Interested to see hoe much of a filesize difference it makes 😄

acburdine:
@kevinansfield is it possible to load a script with ember-ajax? I didn't see anything in the docs that made me think you could, and I feel like wrapping it in a promise manually is easier that re-implementing all of the custom logic in jQuery's getScript method.

kevinansfield:
@acburdine https://github.com/TryGhost/Ghost-Admin/pull/6/files#diff-8576c4b426760b71e0952abe573196edR49 - `ember-ajax` is a wrapper around `jQuery.ajax` so all the same methods/processing options are still there

acburdine:
Oh cool, thanks! 👍

kevinansfield:
You can also [check for the global](https://github.com/TryGhost/Ghost-Admin/pull/6/files#diff-51dfe979bace14376f21078d70a355fdR34) (in this case `CodeMirror`) to avoid having to keep track of when scripts have been loaded which is useful if the logic ever gets moved into a component.

acburdine:
@kevinansfield got it working and works well. I'm thinking of writing a short script that abstracts out some of the broccoli tree manipulation for the ember-cli-build file, in the case that we want to lazy-load more things in the future.

Also, I can make jquery-ui work with this as well if we want, just because I feel like in the short-term it would reduce filesize until I can get ember-sortable to work.

acburdine:
Filesize diff (for reference)

| version | ghost.js | ghost.min.js | vendor.js | vendor.min.js |
|---:|---:|---:|---:|---:|
| master | 1,651,512 (1.6 M) | 855,384 (835 K) | 7,634,320 (7.2 M) | 1,950,683 (1.9 M) |
| with this PR | 1,653,550 (1.6 M) | 856,443 (836 K) | 7,195,148 (6.9 M) | 1,748,703 (1.7 M) |

acburdine:
@kevinansfield ready for review 👍

PR #104, opened 2016-07-01
400 "Version Mismatch" error handling

kevinansfield:
refs https://github.com/TryGhost/Ghost/issues/6949

Handle version mismatch errors by:
- displaying an alert asking the user to copy any data and refresh
- disabling navigation so that unsaved data is not accidentally lost

Detailed changes:
- add `error` action to application route for global route-based error handling
- remove 404-handler mixin, move logic into app route error handler
- update `.catch` in validation-engine so that promises are rejected with the
  original error objects
- add `VersionMismatchError` and `isVersionMismatchError` to ajax service
- add `upgrade-status` service
  - has a method to trigger the alert and toggle the "upgrade required" mode
  - is injected into all routes by default so that it can be checked before
    transitioning
- add `Route` override
  - updates the `willTransition` hook to check the `upgrade-status` service
    and abort the transition if we're in "upgrade required" mode
- update notifications `showAPIError` method to handle version mismatch errors
- update any areas where we were catching ajax errors manually so that the
  version mismatch error handling is obeyed
- fix redirect tests in editor acceptance test
- fix mirage's handling of 404s for unknown posts in get post requests
- adjust alert z-index to to appear above modal backgrounds

TODO:
- [x] 400 version mismatch handling
  - [x] display an alert when saving through Ember Data
  - [x] disable navigation once we've received a "400 Version Mismatch" error
  - [x] update handling for ember-ajax errors
  - [x] update `notifications.showAPIError` to handle version mismatch errors
  - [x] ensure that everywhere we catch an ajax error we are either throwing or calling `showAPIError`
- [x] fix any broken tests
- [x] update alert styling so that they are visible above modal backgrounds


kevinansfield:
⚡ ready for review

PR #105, opened 2016-07-02
Fix image component displaying alt text in uploader

acburdine:
closes TryGhost/Ghost#7064
- pull alt text from parsed markdown
- move text to image uploader component proper


kevinansfield:
@acburdine would you mind taking a look at the failing test when you get a moment? Seems like it should be a quick fix and would be good to get this in 0.9. Thanks 😄

acburdine:
@kevinansfield should be good to go now 😄

PR #108, opened 2016-07-05
Revert to ember-wormhole@0.3.6

kevinansfield:
closes https://github.com/TryGhost/Ghost/issues/7073
- reverts `ember-wormhole` to 0.3.6
- adds working test that was broken under `ember-wormhole` 0.4.0

### DIFF
--- a/package.json
+++ b/package.json
@@ -66,7 +66,7 @@
     "ember-sortable": "1.8.1",
     "ember-suave": "3.0.1",
     "ember-watson": "0.8.2",
-    "ember-wormhole": "0.4.0",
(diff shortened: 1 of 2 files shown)


acburdine:
I'm assuming this has something to do with the fastboot compatibility merged in before 0.4 (that's the only major thing I saw), so do we want to open an issue with ember-wormhole as well?

kevinansfield:
Yeah, would be good to raise an issue however I'm not sure if it's `ember-wormhole` or `ember-power-select` where the problem lies. Saying that it was `ember-wormhole` that threw the error so maybe best to start there.

PR #17, opened 2016-05-22
[WIP | SPIKE] Inline Validations Refactor

acburdine:
closes TryGhost/Ghost#6451
- transition from validationengine and validator.js related dependencies to ember-cp-validations
- stop using DS.Errors

TODO:
- [x] initial experimentation
- [x] build out new components
- [ ] convert old validations => new ones
- [ ] tests
- [ ] remove unneeded dependencies
- [ ] :dancer:

PR #109, opened 2016-07-05
[WIP] DS.Errors is bad for client validations (in other words, the Validations Refactor)

acburdine:
replaces #17, closes TryGhost/Ghost#6451

This PR refactors all of the uses of DS.Errors throughout the application and converts them to ember-changeset & ember-changeset-validations

TODO:
- [x] Build out new input components that use changesets
- [x] switch out all validators to new syntax
- [x] rebase once #110 is merged
- [x] fix issues with ember-changeset philosophy & my own idiocy
- [x] discuss changeset overrides
- [x] remove old code
- [ ] review changeset usage across the application
- [ ] tests
- [ ] 💵  profit


acburdine:
@kevinansfield I've gotten the setup/signup/signin stuff working correctly, would be good to get a look-see on it at some point just to make sure it still works as intended 😄 

As for the rest of the app....well things are a bit broken so those specific flows mentioned are probably the only things that work at the moment 😉

poteto:
@acburdine I'm not familiar with Ghost's codebase, but let me know if I can help at all! I noticed you had to modify ember-changeset a little, I'd be happy to chat sometime to see if we can make any of those changes upstream :)

acburdine:
@poteto thanks! I think I've got all of the infrastructure bits worked out for this refactor.

Re: upstream changes though, was planning on opening an issue detailing some of the modifications I made and why. Most had to do with implementing some custom stuff for recording when a record has been validated, but there was another bit about what actually gets saved to the changeset.

Will have an issue open soon :smile:

app/styles/layouts/auth.css
```diff
@@ -17,6 +17,10 @@
     margin-bottom: 1.5rem;
 }
 
+.gh-signin .form-group.error {
```
acburdine: @JohnONolan / @kevinansfield So on the signin form there's currently an issue where any error messages that get displayed get kind of squished in between the forms. This is due to the selection above that declares `.gh-signin .form-group` to have less of a margin-bottom than all other form-groups.

I played with two approaches to fixing the squished error message text. One was to remove the shorter margin-bottom so it would use the default larger one. The other was this change, which only expands when there is an error class applied to the form-group. This change could in theory break when an `.error` class is applied to the form-group, but there's no error message present; however, given the structure of the validation form group components this won't happen. Although it does look a little odd to have it expand and collapse when the error message disappears and reappears.

My question is what the preferable behavior is? Here are some screenshots of the different results.

Currently:
screen shot 2016-07-20 at 10 56 20 am

With the first option (making the margin-bottom larger for error and non-error states):
http://recordit.co/TCv9P5qjak

With the second option (making the margin-bottom larger when the form-group has an error state):
http://recordit.co/81AyZ7ImVd

Personally, I don't have an opinion either way, so any advice is appreciated 😄

kevinansfield: Has the behaviour here changed? Last time I looked the field should be highlighted red with the error message displayed underneath the form (same place server-related errors are displayed). There was some discussion about that somewhere but it was a long time ago, I can see if I can find it if needed.

acburdine: @kevinansfield No, the behavior hasn't changed at all, but I noticed that any error message text between the fields of the signin form is squished between them. I'm fairly certain this exists on master too, but will double check.

acburdine: OH wait nevermind I see what you mean 😄  However, on master there are no specific errors displayed, just "Please fill out the form to sign in." Any specific field validations don't seem to be displayed.

kevinansfield: Yep, that was the desired behaviour last time we looked at the sign-in validation.

Is that behaviour easy to replicate with the new validation? It's probably more desirable to keep it as-is for now and look at changing the behaviour as part bigger design review.

acburdine: It shouldn't be too difficult to replicate. I'll look over the signin again once i've refactored the rest of the code 😄

acburdine:
Going to tentatively close this as well in favor of a smaller, less "blow up all the things at once" approach to re-writing validations.

PR #95, opened 2016-06-29
Shim-ify the ember client

acburdine:
no issue

Converts ember global imports to ember-cli-shim imports where possible

PR #110, opened 2016-07-06
More Shimming of Controllers

acburdine:
refs #101, #95
- shim controllers that were missed initially


acburdine:
@kevinansfield ready for review 👍

PR #112, opened 2016-07-08
"503 Maintenance" error handling 

kevinansfield:
refs TryGhost/Ghost#6976
- adds custom `MaintenanceError` and associated error checking functions
- updates app route and notifications service to handle `503` errors via the `upgrade-status` service


acburdine:
@kevinansfield do you want all of these to get merged at the same time?

kevinansfield:
If possible as they do all build on each other. I would normally squash them down into 1 PR but I think in this case having the separated commits with their already large commit messages will be better?

acburdine:
ok, I will get them looked at today/this weekend (still stuck deep in validations refactoring 😄)

acburdine:
LGTM!

PR #126, opened 2016-07-15
Recommend Node v4 & use as default

ErisDS:
Now that we're going to recommend v4, this means making it the default test environment.

Whilst I was looking around, I also updated the package.json file so the `engines` match Ghost's exactly. Not sure if there was a reason why they were different?

refs TryGhost/Ghost#7098

- Use Node v4 as our default build environment
- Change engines in package.json to match Ghost core

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ dist: trusty
 
 language: node_js
 node_js:
-  - "0.10"
+  - "4"
 
 env:
   global:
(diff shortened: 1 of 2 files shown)


kevinansfield:
> Not sure if there was a reason why they were different?

No reason that I'm aware of, I think it may have been there simply because the client's package.json was created a long time ago and wasn't kept in sync previously?

PR #128, opened 2016-07-15
Fix editor error handling

kevinansfield:
closes https://github.com/TryGhost/Ghost/issues/7099
- update editor base controller's `showErrorAlert` method to handle `AdapterError` or `AjaxError` objects
- update PSM's error handling so that we don't pass an error object wrapped in an array to `showAPIError` (this was causing the default error message to always be displayed and the global version mismatch/maintenance mode handling to be skipped if an error was encountered when changing a value in the PSM triggered a save)
- add missing title attribute to Mirage's Post factory
- remove some leftover `console.log`s

TODO:
- [x] add test for editor API error handling


kevinansfield:
⚡️ready for review

acburdine:
LGTM 👍

PR #135, opened 2016-07-21
Fix blank signup screen

kevinansfield:
closes https://github.com/TryGhost/Ghost/issues/7117
- adds guard to `sanitizeInput` method of `gh-trim-focus-input` for null/undefined values
- adds acceptance test for successful signup screen flow
- removes unneeded validation/update handling for a non-editable email field
- adds "At least 8 characters" placeholder to password field
- fixes enter key not submitting the form when name or password field has focus

### DIFF
--- a/app/components/gh-trim-focus-input.js
+++ b/app/components/gh-trim-focus-input.js
@@ -32,7 +32,11 @@ const TrimFocusInputComponent = GhostInput.extend({
     },
 
     sanitizeInput(input) {
-        return input.trim();
+        if (input && typeof input.trim === 'function') {
+            return input.trim();
+        } else {
+            return input;
+        }
     },
 
     _focus() {
--- a/app/mirage/config.js
+++ b/app/mirage/config.js
@@ -155,6 +155,7 @@ export function testConfig() {
     // this.urlPrefix = '';    // make this `http://localhost:8080`, for example, if your API is on a different server
     this.namespace = 'ghost/api/v0.1';    // make this `api`, for example, if your API is namespaced
     // this.timing = 400;      // delay for each request, automatically set to 0 during testing
+    // this.logging = true;
 
     /* Authentication ------------------------------------------------------- */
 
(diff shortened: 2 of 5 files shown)


acburdine:
@kevinansfield there seems to be a legitimate test bug here. Chrome and Firefox are disconnecting it appears before all the tests are run 😞

kevinansfield:
Very strange, seems like there is a test interdependency somewhere between one of the Settings acceptance tests and the Signup acceptance test. Hopefully it won't take too long to fully track down.

kevinansfield:
I was missing a mocked request in Mirage, no idea why the infinite loop was only occurring when the Navigation, Slack, and Tags tests were running first, removing any one of those resulted in the test running fine 😕.

Error handling in the Signup route when the attempt to check the validity of an invitation fails probably wants looking at but I think it's OK for now.

acburdine:
Ok :smile: a lot of that behavior has been re-worked as part of the validations refactor, so hopefully that should clean a lot of those issues up.

acburdine:
Hey so I just reviewed this and looks good 👍 However, there are some slight UI bugs that exist on the signup page that might be good to get in this PR. Bugs that I found:

- No placeholder in password field (looks different than the other two fields)
- Pressing "enter" in the password field doesn't complete the form if the field is valid (haven't checked the other two fields but might be the same)

If we want to just merge this one I can create another PR in a bit with fixes for the actual page

kevinansfield:
> No placeholder in password field (looks different than the other two fields)

I'm not sure a placeholder is needed here? The form has labels above the inputs and we don't want to suggest a password as people will undoubtedly just copy it. What text were you thinking of using for the placeholder?

acburdine:
I was just thinking "Password". My initial though though was that because the other fields have placeholders it just looked out-of-place for the password field to not have one.

kevinansfield:
Maybe follow the example in Setup and use "At least 8 characters". I'll add that and an enter key fix to this PR, I already have the files open 😄

kevinansfield:
⚡ ready for review again 😄

acburdine:
LGTM 👍

PR #137, opened 2016-07-22
Convert the Post-Settings-Menu into a Component

acburdine:
no issue
- converts the Post Settings Menu into a component

This is something that's been on my mind for a while, and working on some of the validations refactor I realized it would be much easier to do particular things if the PSM were a component rather than a controller. Plus, it aligns things more with the ember way.
### TODO
- [x] tests


acburdine:
had to fix a lint issue 😅 but should be good to go now

PR #139, opened 2016-07-23
Add promise and run timer scheduling service

acburdine:
no issue
- cleans up various instances of promise and run timer state being passed around

### TODO
- [x] add tests

### DIFF
--- a/app/mixins/editor-base-route.js
+++ b/app/mixins/editor-base-route.js
@@ -1,6 +1,5 @@
 import $ from 'jquery';
 import Mixin from 'ember-metal/mixin';
-import RSVP from 'rsvp';
 import run from 'ember-runloop';
 
 import ShortcutsRoute from 'ghost-admin/mixins/shortcuts-route';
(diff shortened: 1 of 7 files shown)


acburdine:
If anyone has a better name for the service then `scheduler` I'm willing to change it....it just seemed like the easiest name to call it.

app/mixins/editor-base-route.js
```diff
@@ -57,11 +56,10 @@ export default Mixin.create(styleBody, ShortcutsRoute, {
             // so we abort the transition and retry after the save has completed.
             if (state.isSaving) {
                 transition.abort();
-                return run.later(this, function () {
-                    RSVP.resolve(controller.get('lastPromise')).then(() => {
-                        transition.retry();
-                    });
-                }, 100);
+
+                this.get('promise').schedule('post', () => {
```
kevinansfield: Are we ok not having the `return` here?

kevinansfield:
I can see the use case and what led you down this path but I think the naming in the current implementation makes it a bit unintuitive. `scheduler.promise` isn't scheduling anything but rather building a promise queue and it hides the explicitness that we want a promise to happen after something else.

I'm also not 100% on moving the known/explicit `run` usage into a separate utility. Some of the things that this part of the refactor is targeting are symptoms of deeper issues, such as the title observer logic which should really be handled via an action I think?

For the queuing of promises maybe it would make more sense to use `ember-concurrency`'s `enqueue` method? That would also give us some better primitives to use in other parts of the application where we work with asynchronicity.

acburdine:
That makes sense. This was the product of me running into a lot of state being passed between the editor and the PSM, which in order to make the PSM a component I needed to fix.

acburdine:
Closing to re-look at this using `ember-concurrency` instead

PR #149, opened 2016-07-26
Handle manual timezone override

kevinansfield:
refs https://github.com/TryGhost/Ghost/pull/7142, https://github.com/TryGhost/Ghost/pull/7143
- moves timezone selection template and logic into a component
- detect if the current `activeTimezone` is not in our pre-defined list of timezones, if it isn't:
  - add a line indicating that there has been a manual override with the current `activeTimezone` value
  - add a blank option to the timezone select list

timezone-override

### DIFF
--- a/app/components/gh-timezone-select.js
+++ b/app/components/gh-timezone-select.js
@@ -1,0 +1,61 @@
+import Component from 'ember-component';
+import computed, {mapBy} from 'ember-computed';
+import injectService from 'ember-service/inject';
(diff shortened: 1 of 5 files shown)


kevinansfield:
Now wondering if `automatically` would be a better word to use seeing as we're only expecting the override to happen automatically during upgrade and later during setup.

kirrg001:
`automatically` 👍

kevinansfield:
Updated to use `Your timezone has been automatically set to x`

PR #164, opened 2016-07-30
Improve cross-browser support

acburdine:
This PR improves test success in all browsers, and fixes various issues with IE and Edge browsers.

Specific issues fixed:
- `<template>` element does not have a `content` property - closes TryGhost/Ghost#7149
- IE `flex` containers with a min-width property don't correctly vertically-align child elements - [bug](https://connect.microsoft.com/IE/feedback/details/816293/ie11-flexbox-with-min-height-not-vertically-aligning-with-align-items-center)
- Left side of split button in IE shows different height then the right side

### DIFF
--- a/app/components/gh-ed-preview.js
+++ b/app/components/gh-ed-preview.js
@@ -55,6 +55,15 @@ export default Component.extend({
         let template = document.createElement('template');
         template.innerHTML = html;
         let fragment = template.content;
+
+        if (!fragment) {
+            fragment = document.createDocumentFragment();
+
+            while (template.childNodes[0]) {
+                fragment.appendChild(template.childNodes[0]);
+            }
+        }
+
         let dropzones = fragment.querySelectorAll('.js-drop-zone');
         let components = this.get('imageUploadComponents');
 
--- a/tests/acceptance/ghost-desktop-test.js
+++ b/tests/acceptance/ghost-desktop-test.js
@@ -10,23 +10,27 @@ import startApp from '../helpers/start-app';
 import destroyApp from '../helpers/destroy-app';
 import { authenticateSession } from 'ghost-admin/tests/helpers/ember-simple-auth';
 
-const originalUserAgent = window.navigator.userAgent;
+const originalAgent = window.navigator.userAgent;
 
-const _setUserAgent = function (userAgent) {
-    let userAgentProp = {get() { return userAgent; }, configurable: true};
-    Object.defineProperty(window.navigator, 'userAgent', userAgentProp);
-};
+const setUserAgent = function (userAgent) {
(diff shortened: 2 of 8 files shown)


acburdine:
Note: tests are still failing pretty hard in Edge and IE; however, manual testing verifies that this fixes the editor issues there.

tests/acceptance/ghost-desktop-test.js
```diff
@@ -10,23 +10,27 @@ import startApp from '../helpers/start-app';
...
+
+const setUserAgent = function (userAgent) {
+    let userAgentProp = {
+        get() {
+            return userAgent;
+        },
+        configurable: true
+    };
+
+    try {
```
acburdine: Firefox and Chrome handle navigation overrides differently than Safari. Safari doesn't let you change the `userAgent` property directly, so the `catch` block handles the exception from the first method and then creates a dummy `window.navigator` object that works in Safari.

acburdine: Note: tested this locally in Safari (with `ember test`) and it passed.

tests/unit/components/gh-selectize-test.js
```diff
@@ -31,8 +31,8 @@ describeComponent(
                 component._selectize.setValue(['item 3', 'item 2']);
             });
 
-            expect(component.get('value'), 'component value').to.deep.equal(['item 3', 'item 2']);
-            expect(component.get('selection'), 'component selection').to.deep.equal(['item 3', 'item 2']);
+            expect(component.get('value').toArray(), 'component value').to.deep.equal(['item 3', 'item 2']);
```
acburdine: For some reason, Safari handles ember arrays differently than Chrome/Firefox. To get around that, we cast the ember array to a regular array to verify the deep equal. (Note: I tried wrapping the expected value in an emberA - which would look like `expect(someEmberArray).to.deep.equal(emberA(['a', 'b']));` but that didn't work either)

app/components/gh-ed-preview.js
```diff
@@ -55,6 +55,15 @@ export default Component.extend({
         let template = document.createElement('template');
         template.innerHTML = html;
         let fragment = template.content;
+
+        if (!fragment) {
```
acburdine: So my original attempt at this used a polyfill, but because we manually create the `template` element using `document.createElement`, the only polyfill that would have worked is to override document.createElement entirely. This was a bit tricky, especially given that we set the `innerHTML` of the template before we actually get the content itself. Attempts at using the ES5 `get` method were unsuccessful. Also, determining where to load the polyfill in the stack of JS files was problematic. Ultimately this was easier, and avoided overriding such a critical method.

PR #168, opened 2016-08-01
Fix subscriber height being too tall for viewport

dbalders:
Closes TryGhost/Ghost#7163

* With flex-column, setting height to 100% will make the child the same
as the full height of the flex container, not fill up the remaining
space, so it was bleeding down below the viewport


acburdine:
Thanks @dbalders!

PR #169, opened 2016-08-02
Run browser tests in parallel

acburdine:


acburdine:
Ok, as it turns out this really doesn't save any time and exacerbates some of the timing issues. Nevermind 😄

PR #170, opened 2016-08-02
Invite user modal fixes

dbalders:
Closes TryGhost/Ghost#7132
- Adds tabindex -1 to the select span to allow the tab to go straight
  from input to select
- Adds a focus to the new user input on open
  - Needs the run later otherwise it is triggering too fast and not
    getting the input

### DIFF
--- a/app/components/modals/invite-new-user.js
+++ b/app/components/modals/invite-new-user.js
@@ -46,6 +46,13 @@ export default ModalComponent.extend(ValidationEngine, {
         this.set('hasValidated', emberA());
     },
 
+    didInsertElement() {
+        let emailInput = this;
+        run.later(() => {
+            emailInput.$('#new-user-email').focus();
+        }, 500);
+    },
+
     validate() {
         let email = this.get('email');
 
--- a/app/templates/components/modals/invite-new-user.hbs
+++ b/app/templates/components/modals/invite-new-user.hbs
@@ -26,7 +26,7 @@
 
         <div class="form-group for-select">
             <label for="new-user-role">Role</label>
-            <span class="gh-select" tabindex="0">
+            <span class="gh-select" tabindex="-1">
                 {{gh-select-native id="new-user-role"
                     content=roles
                     optionValuePath="id"
--- a/tests/integration/components/modals/invite-new-user-test.js
+++ b/tests/integration/components/modals/invite-new-user-test.js
@@ -1,0 +1,37 @@
+/* jshint expr:true */
+import { expect } from 'chai';
+import {
+    describeComponent,
+    it
+} from 'ember-mocha';
+import hbs from 'htmlbars-inline-precompile';
+import startApp from 'ghost-admin/tests/helpers/start-app';
+import run from 'ember-runloop';
+import $ from 'jquery';
+
+describeComponent(
+    'modals/invite-new-user',
+    'Integration: Component: modals/invite-new-user',
+    {
+        integration: true
+    },
+    function() {
+        let application;
+
+        beforeEach(function () {
+            application = startApp();
+        });
+
+        it('renders', function() {
+            this.render(hbs`{{modals/invite-new-user}}`);
+            expect(this.$()).to.have.length(1);
+        });
+
+        it('input has focus', function() {
+            this.render(hbs`{{modals/invite-new-user}}`);
+            run.later(this, function () {
+                expect($(this).is(':focus')).to.be.true;
+            }, 500);
+        });
+    }
+);


app/components/modals/invite-new-user.js
```diff
@@ -46,6 +46,12 @@ export default ModalComponent.extend(ValidationEngine, {
         this.set('hasValidated', emberA());
     },
 
+    didRender() {
```
dbalders: If there is an easier way (or less hardcoded) to do this than the way I am, let me know. Guessing there is, but this is currently functioning correctly

acburdine:
@dbalders another way you could go about this is to convert the `gh-input` into a `gh-trim-focus-input`. Those already have the focus functionality built in. See the signin form template for an example 😄

dbalders:
@acburdine kk, thanks, ill look into that

dbalders:
@acburdine the trim isn't liking the `email` at the beginning. Getting `Cannot read property 'trim' of undefined` in the `sanitizeInput` function. Any idea what to change that to?

dbalders:
so, i fix the problem from the last comment, but the field is still not taking focus without some sort of delay built in. The code for that works now, but obviously shouldn't stay the way it is. Any thoughts on this?

acburdine:
@dbalders ah, I see :)

So the trim-focus-input is doing what it's designed to do I believe, the issue is that it's trying to be focused before the component is fully rendered (I think).

Regardless, switching back to the regular input's probably good.

app/components/modals/invite-new-user.js
```diff
@@ -46,6 +47,12 @@ export default ModalComponent.extend(ValidationEngine, {
         this.set('hasValidated', emberA());
     },
 
+    // didRender() {
```
acburdine: So instead of doing a `run.later`, try this:

```js
import run, {scheduleOnce} from 'ember-runloop'; // at top of file

// in place of didRender
didInsertElement() {
    scheduleOnce('afterRender', () => {
        this.$('#new-user-email').focus();
    });
}
```

app/components/modals/invite-new-user.js
acburdine: (copied from earlier line note)

So instead of doing a `run.later`, try this:

```js
import run, {scheduleOnce} from 'ember-runloop'; // at top of file

// in place of didRender
didInsertElement() {
    scheduleOnce('afterRender', () => {
        this.$('#new-user-email').focus();
    });
}
```

app/components/modals/invite-new-user.js
```diff
@@ -4,6 +4,7 @@ import {A as emberA} from 'ember-array/utils';
 import run from 'ember-runloop';
 import ModalComponent from 'ghost-admin/components/modals/base';
 import ValidationEngine from 'ghost-admin/mixins/validation-engine';
+import run, {scheduleOnce} from 'ember-runloop';
```
acburdine: oops, should have mentioned it in the other note...this line is partially a duplicate of line 4 😄

acburdine:
So I played around with this and did some googling, and it seems to me that because of the transitions that occur as part of the `liquid-tether` modal setup, setting the focus of the input using `scheduleOnce('afterRender')` just isn't possible (the input is _technically_ in the DOM but not visible, so focusing fails). So - it appears using `run.later` is the better way to go about this 😄 /cc @kevinansfield 

There is one issue with focusing the input on render (regardless of the way it's done) - when the user clicks on the 'X' close button - there is a brief flash of validation error on the field before the modal closes. It shouldn't be hard to stop that (see the `closeModal` method on the base modal and the current `willDestroyElement` of the invite user modal).

Also - would you be willing to write an acceptance/integration test for this?

dbalders:
@acburdine yep! all sounds good :). I will ping when its done

dbalders:
Oh @acburdine, regarding the validation error showing up, that is because of the onfocusout being run. Talked about that on my last comment on TryGhost/Ghost#7131. Waiting to see what to do for that. So, if its cool, may just do separate PRs. One for this, then one to fix whatever is decided in the other one? Does that sound like a good idea? Or, we can just solve the issue here and I can do both.

dbalders:
@acburdine how does this look?

app/components/modals/invite-new-user.js
```diff
@@ -2,6 +2,7 @@ import RSVP from 'rsvp';
 import injectService from 'ember-service/inject';
 import {A as emberA} from 'ember-array/utils';
 import run from 'ember-runloop';
+import $ from 'jquery';
```
acburdine: you can remove this line and use `this.$('#new-user-email').focus()` below instead

dbalders: That actually gives a `Cannot read property '$' of undefined` because `this` is undefined (b/c it is run later im guessing).

acburdine: Or even better than that:

```js
run.later(() => {
    this.$('#new-user-email').focus();
}, 500);
```

Using es2016 arrow functions :wink:

dbalders: updated

kevinansfield:
Original issue resolved via https://github.com/TryGhost/Ghost-Admin/commit/4bf815dba4f8c803f4ffadb97c7d15a1ee87dc1d and https://github.com/TryGhost/Ghost-Admin/pull/1020

PR #172, opened 2016-08-02
Fix focusing URL field of the "new nav item" row causes validation er…

shawnheide:
- [x] Commits are squashed 
- [x] The build will pass

Closes #7140

-change navigation-item model to check if url is identical to '/' instead of blank since blank values are overwritten by '/' before checks

### DIFF
--- a/app/models/navigation-item.js
+++ b/app/models/navigation-item.js
@@ -20,6 +20,6 @@ export default EmberObject.extend(ValidationEngine, {
     isBlank: computed('label', 'url', function () {
         let {label, url} = this.getProperties('label', 'url');
 
-        return isBlank(label) && isBlank(url);
+        return isBlank(label) && (isBlank(url) || url === '/');
     })
 });


acburdine:
@shawnheide can you please update the commit message to target Ghost itself?

e.g. `closes TryGhost/Ghost#7140` instead of `closes #7140`

Thanks! 😄

shawnheide:
Yep, no problem. Thanks for letting me know.

shawnheide:
Looks like I accidentally ran the tests under my Ghost fork instead of Ghost-Admin. The Travis tests failed because the url is hardcoded as blank in some of the tests, even though it seems that this can't happen in practice. One option is to just add the 'isBlank(url)' back to the 'isBlank' validation:
``` javascript
return isBlank(label) && (isBlank(url) || url === '/)';
```
I ran the tests, properly this time, and everything passed with this change. 

The other option would be to rewrite the tests, replacing the expected blanks with '/'. Any thoughts on the best way to move forward with this?

shawnheide:
@acburdine Any suggestions on how to proceed?

acburdine:
@shawnheide Sorry for not getting to this sooner. Unfortunately @dbalders stole your ⚡ a little bit on the issue, however his fix introduced another bug, which if you're willing it would be great to fix!

If you're interested, see [this comment](https://github.com/TryGhost/Ghost/issues/7140#issuecomment-238024478) for details on the new bug.

Let me know if you have any more questions 😄

dbalders:
Sorry about that @shawnheide. didn't see yours when I pr'd.

shawnheide:
No worries. It was interesting to see that we had separate ways of fixing it. Cheers.

PR #175, opened 2016-08-03
Full name field validation style fixes

dbalders:
Closes TryGhost/Ghost#7134
- Added a lightred variable for any errors that need to be a little big
  lighter
- Added the light error class to the input and error
- Fixed input not being highlighted on error

screen shot 2016-08-02 at 7 49 20 pm

### DIFF
--- a/app/styles/layouts/user.css
+++ b/app/styles/layouts/user.css
@@ -117,7 +117,7 @@
     .user-details-top label[for="user-name"] {
         color: transparent;
     }
-    .user-details-top .user-name {
+    .user-name {
         border-color: #fff;
     }
 }
--- a/app/styles/patterns/forms.css
+++ b/app/styles/patterns/forms.css
@@ -40,6 +40,9 @@ input {
     color: var(--red);
 }
 
+.error .response.light-error {
+    color: var(--lightred);
+}
 
 /* Form Groups
 /* ---------------------------------------------------------- */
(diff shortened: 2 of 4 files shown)


kevinansfield:
Thanks for taking a look at this @dbalders but I think this may need a heavier design change - the image above is still very difficult to read on my screen and changing the colour will only work when the colour of the cover image (which we don't control) is something the provides suitable contrast.

dbalders:
@kevinansfield agreed. I played around with adding a white background to the error since that way we would always know it had good contrast, but thought I would start here and see what you thought. Should I try that or did you have something else in mind?

dbalders:
If we did go with a white background, just adding the background and leaving it full width would look like this: 

screen shot 2016-08-03 at 9 56 31 am

Or, by adding a parent container, can just add the background to the text (also added a red border to it to show that option too):

screen shot 2016-08-03 at 9 59 56 am

kevinansfield:
@dbalders thanks for taking another stab at this and sorry for so long with no response!

@JohnONolan any thoughts on this issue? David's latest changes solve the readability problem but I don't feel the design matches anything from the existing styles, any ideas for a quick-win here?

kevinansfield:
Note this has been resolved in master via https://github.com/TryGhost/Ghost-Admin/pull/695, I'm not sure we'll be revisiting the design in LTS.

kevinansfield:
Closing this for now to clean up the PR list. We can re-visit later if it's determined the fix really needs to go into LTS.

PR #176, opened 2016-08-03
Move ember build tasks over from main repository

acburdine:
refs TryGhost/Ghost#6977
- uses npm script's custom options to streamline ember commands

**NOTE** this will not work with npm versions < 2.0.0, due to the use of custom arguments to npm run scripts. However, as the latest maintenance releases of Node 0.10 ship with npm 2 as default, and node 0.10 support will be dropped in October, I think it's ok, especially as it only affects development.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -8,10 +8,7 @@ var _              = require('lodash'),
     getTopContribs = require('top-gh-contribs'),
     moment         = require('moment'),
     chalk          = require('chalk'),
-    Promise        = require('bluebird'),
-
-    escapeChar = process.platform.match(/^win/) ? '^' : '\\',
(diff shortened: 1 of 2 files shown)


Gruntfile.js
```diff
@@ -85,21 +82,61 @@ module.exports = function(grunt) {
...
+        },
+
         shell: {
             'npm-install': {
                 command: 'npm install'
             },
 
             'bower-install': {
-                command: path.resolve(cwd + '/node_modules/.bin/bower install')
+                command: 'bower install'
```
kevinansfield: Does this change require `bower` to be installed globally? The previous method had been used in order to prevent that requirement

kevinansfield: Ignore me, I'd missed the `preferLocal` option below 😁

PR #177, opened 2016-08-03
Move csscomb watch task from server repo

acburdine:
refs TryGhost/Ghost#6977

(forgot to include this with the other PR 😁 )

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -87,13 +87,14 @@ module.exports = function(grunt) {
         bgShell: {
             ember: {
                 cmd: 'npm run build -- --watch',
-                bg: true,
-                stdout: function (out) {
-                    grunt.log.writeln(chalk.cyan('Ember-cli::') + out);
-                },
-                stderror: function (error) {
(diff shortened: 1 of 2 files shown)


Gruntfile.js
```diff
@@ -87,12 +87,16 @@ module.exports = function(grunt) {
...
+                bg: true
+            }
+        },
+
+        watch: {
+            csscomb: {
+                files: ['app/styles/**/*.css'],
+                tasks: ['shell:csscombfix'],
+                options: {
+                    livereload: 35730
```
acburdine: The reason why this port is specified is because the default port would conflict with the watch task of the server repo when running `grunt dev`

kevinansfield: Will the livereload do anything in this case? We'd need to inject the live reload script into the admin HTML and set it up to listen to the correct port, maybe setting `livereload: false` will be better until we can utilise it?

acburdine: oops yeah, that makes sense. I'll just remove that section entirely :)

PR #194, opened 2016-08-09
Fix new url field focus bug

dbalders:
Closes TryGhost/Ghost#7140

* Now checks to see if it is new or not before editing the url

### DIFF
--- a/app/components/gh-navitem-url-input.js
+++ b/app/components/gh-navitem-url-input.js
@@ -131,7 +131,7 @@ export default TextField.extend(InvokeActionMixin, {
                 url = url.replace(baseUrlParts.pathname.slice(0, -1), '');
             }
 
-            if (url !== '') {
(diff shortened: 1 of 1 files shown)


app/components/gh-navitem-url-input.js
```diff
@@ -131,7 +131,7 @@ export default TextField.extend(InvokeActionMixin, {
                 url = url.replace(baseUrlParts.pathname.slice(0, -1), '');
             }
 
-            if (url !== '') {
+            if ((url !== '') | (!this.get('isNew'))) {
```
acburdine: Shouldn't this be more like: `if (url !== '' || !this.get('isNew'))`?

dbalders: thanks

acburdine:
thanks @dbalders!

PR #203, opened 2016-08-11
Convert tagNames to use computed.mapBy

acburdine:
- fixes travis test bug

This seems to happen fairly often in Travis tests, and causes them to need to be restarted incredibly often. I'm not sure if this points to a deeper issue, but as I cannot reproduce the issue locally my only thought was to figure out where the issue was coming from and ~~fix~~ work around it.

/cc @kevinansfield


acburdine:
Turns out that didn't actually fix it, investigating further.

acburdine:
Tests aren't quite done yet, but all the chrome ones have passed :)

✨  Ready for review

PR #206, opened 2016-08-12
Use new Mocha testing API

Turbo87:
This PR is related to https://github.com/switchfly/ember-mocha/pull/84. It serves as a real-life example and testbed for the proposed new API. Feel free to give feedback on either one of the PRs.

### DIFF
--- a/tests/integration/components/gh-image-uploader-with-preview-test.js
+++ b/tests/integration/components/gh-image-uploader-with-preview-test.js
@@ -1,46 +1,41 @@
 /* jshint expr:true */
 import {expect} from 'chai';
-import {
-    describeComponent,
-    it
-} from 'ember-mocha';
+import {describe, it} from 'mocha';
+import {setupComponentTest} from 'ember-mocha';
 import hbs from 'htmlbars-inline-precompile';
 import run from 'ember-runloop';
 import sinon from 'sinon';
 
-describeComponent(
-    'gh-image-uploader-with-preview',
-    'Integration: Component: gh-image-uploader-with-preview',
-    {
+describe('Integration: Component: gh-image-uploader-with-preview', function() {
+    setupComponentTest('gh-image-uploader-with-preview', {
         integration: true
-    },
-    function() {
-        it('renders image if provided', function() {
-            this.set('image', 'http://example.com/test.png');
+    });
 
-            this.render(hbs`{{gh-image-uploader-with-preview image=image}}`);
+    it('renders image if provided', function() {
+        this.set('image', 'http://example.com/test.png');
 
-            expect(this.$('.gh-image-uploader.-with-image').length).to.equal(1);
-            expect(this.$('img').attr('src')).to.equal('http://example.com/test.png');
+        this.render(hbs`{{gh-image-uploader-with-preview image=image}}`);
+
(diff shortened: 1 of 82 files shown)


acburdine:
@Turbo87 would you be willing to rebase this PR on top of master and create two commits for this (one for the bower dependency update of ember-mocha, the other for all of these changes)? Thanks!

Turbo87:
@acburdine yeah, I'm waiting for the next release of ember-cli-mocha to include these changes and then I'll rebase to use the official release.

Turbo87:
@acburdine I've just updated this to use `ember-mocha@0.12.0`, but it seems that the `Integration: Component: gh-image-uploader-with-preview renders upload form when no image provided` test is now failing. unfortunately I haven't yet been able to figure out why...

acburdine:
@Turbo87 I'll look into this today and see if I can't figure out what the issue is.

tests/integration/components/gh-image-uploader-with-preview-test.js
```diff
@@ -1,46 +1,41 @@
...
-            expect(this.$('input[type="file"]').length).to.equal(1);
-        });
+    it('renders upload form when no image provided', function () {
+        this.render(hbs`{{gh-image-uploader-with-preview image=image}}`);
 
-        it('triggers remove action when delete icon is clicked', function () {
-            let remove = sinon.spy();
-            this.set('remove', remove);
-            this.set('image', 'http://example.com/test.png');
+        expect(this.$('input[type="file"]').length).to.equal(1);
```
acburdine: @Turbo87 This is the failing test right here, and after inserting a console.log statement I found the error.

It appears the component is retaining the value from this line: `this.set('image', 'http://example.com/test.png');` in the `it` above.

Is it an incorrect understanding that each `it` is supposed to be run with a clean slate? i.e. `this.get('image')` should return undefined for the second it because it hasn't been explicitly set in the current test.

If I'm incorrect in that understanding, adding a `this.set('image', null);` above this line should fix the failing test.

Turbo87: 🤔 

that indeed sounds like a bug. it should test with a clean slate. I'll investigate. Thanks for figuring this out!

Turbo87: I've opened https://github.com/emberjs/ember-mocha/issues/110 to track this, and I have two failing test cases here that reproduce the issue

Turbo87: @acburdine since @rwjblue doesn't agree with me in https://github.com/emberjs/ember-mocha/issues/110 I've changed the test code here to explicitly reset the `image` property before running the test. This should get you going while we figure out what the intended behavior is supposed to be.

Turbo87: released the fix as v0.9.4 and removed the patch commit here again

Turbo87: updated to ember-cli-mocha v0.12.1 to bust the cache on TravisCI. Seems to work as expected now.

acburdine:
Thanks @Turbo87 for all the hard work!

PR #214, opened 2016-08-18
✨ validate file uploads against "accept" mime-type before uploading

kevinansfield:
closes https://github.com/TryGhost/Ghost/issues/7144
- allow `gh-file-uploader`'s `accept` attr to be specified
- allows a `validate` action to be passed into `gh-image-uploader` and `gh-file-uploader` components that runs after a file is selected and before the upload starts
- adds a default `validate` action to `gh-image-uploader` and `gh-file-uploader` that triggers the normal `UnsupportedFileType` error when the selected file's mime-type does not match the `accept` attribute

### DIFF
--- a/app/components/gh-file-uploader.js
+++ b/app/components/gh-file-uploader.js
@@ -9,7 +9,8 @@ import { invoke, invokeAction } from 'ember-invoke-action';
 import {
     isVersionMismatchError,
     isRequestEntityTooLargeError,
-    isUnsupportedMediaTypeError
+    isUnsupportedMediaTypeError,
+    UnsupportedMediaTypeError
 } from 'ghost-admin/services/ajax';
 
 export default Component.extend({
@@ -20,6 +21,8 @@ export default Component.extend({
     labelText: 'Select or drag-and-drop a file',
     url: null,
     paramName: 'file',
+    accept: 'text/csv',
+    validate: null,
 
     file: null,
     response: null,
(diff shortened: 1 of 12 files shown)


app/components/gh-file-uploader.js
```diff
@@ -150,10 +170,22 @@ export default Component.extend({
...
-            this.set('file', fileList[0]);
-            run.schedule('actions', this, function () {
-                this.generateRequest();
-            });
+            // can't use array destructuring here as FileList is not a strict
+            // array and fails in Safari
+            // jscs:disable requireArrayDestructuring
+            let file = fileList[0];
+            // jscs:enable requireArrayDestructuring
+            let validationResult = invokeAction(this, 'validate', file);
```
acburdine: Instead of the hack above, couldn't you do something like:

```js
let validationResult = invokeAction(this, 'validate', file) || this._defaultValidate(file);
```

invokeAction will return undefined if no action exists for that name.

acburdine:
It looks like there's a couple of timing issues perhaps that this introduced?

app/components/gh-file-uploader.js
kevinansfield: That would fail if the we have a validate action and it returns `false` for something which isn't what the default validation checks, it may be possible if we use a long-form conditional and explicitly check for `undefined` but then we could probably just check for the existence of `this.get('validate')`  🤔 - I'll try that route

acburdine:
This might be a tad overkill - but since the actions/default validation/etc. stuff is exactly the same in both uploaders - would it be possible to extract it into a mixin (in the interests of DRY-ness)?

kevinansfield:
Cleaned up the validation and added fixes for the tests - validation was failing rather than timing errors although I didn't see the original failed tests. Should be good for review/merge now 😎

kevinansfield:
> This might be a tad overkill - but since the actions/default validation/etc. stuff is exactly the same in both uploaders - would it be possible to extract it into a mixin (in the interests of DRY-ness)?

Yep, I did mention in another uploader issue (https://github.com/TryGhost/Ghost/issues/6661) about addressing all the duplication between uploaders as there's a lot more than just the validation but I wanted to see the `gh-profile-image` implementation at the same time so that we abstract the right things.

acburdine:
So this is looking good, except there's a bit of a visual bug in all the upload elements I've looked at

This is one (from the PSM - probably happens in the tag settings menu as well)
screen shot 2016-08-20 at 8 00 26 am

And all of the other ones that I've seen look like this:
screen shot 2016-08-20 at 8 01 51 am

I'm thinking the height might need to be changed?

acburdine:
Also - do we want to do the same for the labs import uploader? currently the server will return an unsupported media type error but since we're validating it client here it might not be a bad idea.

kevinansfield:
> I'm thinking the height might need to be changed?

I've already changed this in the [theme uploads PR](https://github.com/TryGhost/Ghost-Admin/pull/210/files#diff-b2e9dc4e88517fd5bcaa218ac5e1bd66L13) but I need to do some testing to ensure it doesn't have any side-effects in other areas.

> Also - do we want to do the same for the labs import uploader?

I've added that although the labs importer is in need of refactoring. Once the uploader components have been refactored it would be good to revisit the labs import to have a proper progress bar and better error styling.

acburdine:
LGTM, thanks 👍

PR #222, opened 2016-08-24
🐛 prevent task-buttons dropping clicks after focus-out events

kevinansfield:
refs https://github.com/TryGhost/Ghost/issues/7255
- adds a `.appear-disabled` CSS class that doesn't prevent pointer events like `.disabled`
- updates `gh-task-button`:
  - use `.appear-disabled` class instead of actually disabling button
  - add check to guard against the button's assigned task being run multiple times whilst the spinner is running

This resolves the [user profile slug issue](https://github.com/TryGhost/Ghost/issues/7255) where clicking the Save button whilst the slug input has focus would only trigger the input's focus-out event due to it immediately disabling the button.

There is a simpler alternative where we use the button's `click` handler to disable the button then re-enable once the task finishes but this would mean that an external save event (eg. a keyboard shortcut) would make the button spin and look disabled but a click would still run the task twice - the proposed solution in this PR will always prevent the button's task from running simultaneously.

### DIFF
--- a/app/components/gh-task-button.js
+++ b/app/components/gh-task-button.js
@@ -18,13 +18,27 @@ import layout from 'ghost-admin/templates/components/gh-spin-button';
 export default SpinButton.extend({
     layout, // This is used to we don't have to re-implement the template
 
+    classNameBindings: ['showSpinner:appear-disabled'],
+
     task: null,
 
     submitting: reads('task.last.isRunning'),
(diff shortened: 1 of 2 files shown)


acburdine:
LGTM, thanks! I'll look at adding tests for this button soon, I just ran into ember concurrency issues the first time around 😕

PR #223, opened 2016-08-24
🐛 fix post-settings-menu slug input

kevinansfield:
refs https://github.com/TryGhost/Ghost/issues/7255
- adds missing `slugGenerator` service injection


acburdine:
Crap, didn't realize I'd missed that injection 😞

PR #226, opened 2016-08-24
✨ detailed theme validation errors

kevinansfield:
no issue

Display the detailed validation errors that we get back from gscan so that users know how to fix their theme

- add `ThemeValidationError` to custom ajax error types
- intercept failed theme uploads from the uploader component and switch to an "invalid theme" state if we have a theme validation error, otherwise display an error in the upload component as usual

TODO:
- [x] styling
- [x] tests

screen shot 2016-08-24 at 18 59 18


kevinansfield:
@JohnONolan: I've thrown in some very basic styling for now. Any thoughts on what's there (screenshot included above) or is it good enough for now?

JohnONolan:
👍 🚢

PR #236, opened 2016-08-26
🐛  Allow .zip file upload for file import

AileenCGN:
closes TryGhost/Ghost#7277

Adds `application/zip` to supported import file types.

### DIFF
--- a/app/controllers/settings/labs.js
+++ b/app/controllers/settings/labs.js
@@ -11,7 +11,7 @@ export default Controller.extend({
     submitting: false,
     showDeleteAllModal: false,
 
-    importMimeType: 'application/json',
+    importMimeType: ['application/json', 'application/zip'],
 
     ghostPaths: injectService(),
     notifications: injectService(),


app/controllers/settings/labs.js
```diff
@@ -11,7 +11,7 @@ export default Controller.extend({
     submitting: false,
     showDeleteAllModal: false,
 
-    importMimeType: 'application/json',
+    importMimeType: ['application/json', 'application/zip'],
```
kirrg001: i think it makes sense to start adding two tests  👍 one for json, one for zip

AileenCGN: It does, agreed 100%. Nothing there yet to test the existing file upload yet. Might take a bit longer, as I don't think it's possible in an acceptance test... 😔 
Maybe @acburdine has an idea for that?

kirrg001: same for here: tests as separate PR. i would like to merge this PR 👍

PR #241, opened 2016-09-01
Fix json db uploads and add checking for subscriber uploads

acburdine:
closes TryGhost/Ghost#7295
- check json content on db import if type is not specified
- ensure file type/extension is being checked in subscriber upload

### DIFF
--- a/app/components/gh-file-uploader.js
+++ b/app/components/gh-file-uploader.js
@@ -4,6 +4,7 @@ import injectService from 'ember-service/inject';
 import computed from 'ember-computed';
 import {isBlank} from 'ember-utils';
 import run from 'ember-runloop';
+import {isEmberArray} from 'ember-array/utils';
 
 import { invoke, invokeAction } from 'ember-invoke-action';
 import {
@@ -21,7 +22,8 @@ export default Component.extend({
     labelText: 'Select or drag-and-drop a file',
     url: null,
     paramName: 'file',
-    accept: 'text/csv',
+    accept: ['text/csv'],
+    extensions: ['csv'],
     validate: null,
 
     file: null,
@@ -72,6 +74,15 @@ export default Component.extend({
                 this.send('upload');
             });
         }
+    },
+
+    didReceiveAttrs() {
+        this._super(...arguments);
+        let accept = this.get('accept');
+        let extensions = this.get('extensions');
+
+        this._accept = (!isBlank(accept) && !isEmberArray(accept)) ? accept.split(',') : accept;
+        this._extensions = (!isBlank(extensions) && !isEmberArray(extensions)) ? extensions.split(',') : extensions;
     },
 
     willDestroyElement() {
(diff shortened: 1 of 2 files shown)


app/components/gh-file-uploader.js
```diff
@@ -21,7 +21,8 @@ export default Component.extend({
     labelText: 'Select or drag-and-drop a file',
     url: null,
     paramName: 'file',
-    accept: 'text/csv',
+    accept: ['text/csv'],
+    extensions: ['csv'],
```
kevinansfield: The main reason I went with a strings here was so that they can be expressed easily in handlebars when there are multiple accepted mime types/extensions e.g. 

```handlebars
{{gh-file-uploader
    accept="image/gif,image/jpg,image/jpeg,image/png,image/svg+xml"
    extensions="gif,jpg,jpeg,png,svg"}}
```

app/controllers/settings/labs.js
```diff
@@ -18,30 +21,67 @@ export default Controller.extend({
...
 
+    // TODO: convert to ember-concurrency task
+    _validate(file) {
+        // Windows doesn't have mime-types for json files by default, so we
+        // need to have some additional checking
+        if (file.type === '') {
+            // First check file extension so we can early return
+            let [, extension] = (/(?:\.([^.]+))?$/).exec(file.name);
+
+            if (!extension || extension !== 'json') {
```
kevinansfield: It may be a good idea for extension checks to be case-insensitive.

app/components/gh-file-uploader.js
```diff
@@ -189,7 +190,15 @@ export default Component.extend({
...
-        if (!isBlank(accept) && file && accept.indexOf(file.type) === -1) {
+        if (!isBlank(accept) && file && !isBlank(file.type) && accept.indexOf(file.type) === -1) {
+            return new UnsupportedMediaTypeError();
+        }
+
+        // First check file extension so we can early return
+        let [, extension] = (/(?:\.([^.]+))?$/).exec(file.name);
+        let extensions = this.get('extensions');
+
+        if (!extension || extensions.indexOf(extension) === -1) {
```
kevinansfield: It may be a good idea for extension checks to be case-insensitive.

app/components/gh-file-uploader.js
kevinansfield: Maybe we can use `didReceiveAttrs` to build an internal array from any supplied strings?

app/components/gh-file-uploader.js
```diff
@@ -74,6 +76,20 @@ export default Component.extend({
         }
     },
 
+    didReceiveAttrs() {
+        this._super(...arguments);
+        let accept = this.get('accept');
+        let extensions = this.get('extensions');
+
+        if (!isBlank(accept) && !isEmberArray(accept)) {
+            this.set('accept', accept.split(','));
```
kevinansfield: Hmm, this would result in updating the two-way binding. I was thinking something like `this._accept = accept.split(',')` - no need to use getters/setters as we're dealing with completely internal POJOs.

acburdine: I tried doing that and I got a lot of undefined errors - will update it to not use two-way binding but it didn't seem to like me just setting properties on the component.

acburdine: Nevermind, I'm just tired 😛

acburdine:
@kevinansfield all changes made and tests _should_ be passing :)

PR #242, opened 2016-09-01
🐛 fix duplicate server alerts + default alert bg color

kevinansfield:
closes https://github.com/TryGhost/Ghost/issues/7305

The `location` property of server-side notifications was being ignored by the client and so wasn't being used to de-duplicate alerts.

- adds `key` attribute to `Notification` model
- adds a serializer for notifications that renames the `location` key sent by the server to `key`
- set the default background color of alerts to white so that alerts with no `type` set do not inherit the background color (useful in Ghost Desktop which has a transparent background color set)

### DIFF
--- a/app/models/notification.js
+++ b/app/models/notification.js
@@ -5,5 +5,6 @@ export default Model.extend({
     dismissible: attr('boolean'),
     status: attr('string'),
     type: attr('string'),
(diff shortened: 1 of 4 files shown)


acburdine:
LGTM 👍

PR #247, opened 2016-09-05
Update Gruntfile.js

yedaodao:
add --allow-root for the ESUDO error about using 'bower install' in docker

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -104,7 +104,7 @@ module.exports = function(grunt) {
             },
 
             'bower-install': {
-                command: 'bower install'
+                command: 'bower install --allow-root'
             },
 
             ember: {


ErisDS:
Hi @yedaodao! This change was suggested & rejected previously: https://github.com/TryGhost/Ghost/pull/2351 and hasn't come up again since. Do you have a clear case for why this in fact needed? If not I would be in favour of not merging, for the same reason as last time 😉

yedaodao:
@ErisDS Thank you. When I build the custom ghost image with docker, I use `grunt init`, `grunt prod` for building the project, but the error logs remind me that there is a ESUDO error about bower install. I think I can use bower install manually to avoid this problem, bug in my opinion adding this argument is more standardized.

ErisDS:
Doing a search for `bower install` on GitHub does not turn up any other projects which do this, and therefore I am unable to find evidence to support your suggestion that this is standardized. If you have some, feel free to share 👍

PR #262, opened 2016-09-14
✨ display warnings after theme upload

kevinansfield:
refs TryGhost/Ghost#7362, requires TryGhost/Ghost#7367
- display any gscan warnings we get back from a successful upload (very useful for the downgrade of missing `{{asset}}` helpers from an error to a warning)


ErisDS:
Cool, self-assigning this one, will merge when I've dug into the other related issues we had earlier a bit more thoroughly.

PR #273, opened 2016-09-20
Revert "Update ember-one-way-controls to version 1.1.1 🚀"

kevinansfield:
Reverts TryGhost/Ghost-Admin#252

Using this version of `ember-one-way-controls` results in a bug that prevents the spacebar working when editing post titles.

I'm not sure if it's related to our own code or the addon but reverting to the known-working version for now.

### DIFF
--- a/package.json
+++ b/package.json
@@ -62,7 +62,7 @@
     "ember-light-table": "0.1.9",
     "ember-load-initializers": "0.5.1",
     "ember-myth": "0.1.1",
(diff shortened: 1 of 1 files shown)


kevinansfield:
This issue sounds similar to https://github.com/DockYard/ember-one-way-controls/issues/119 - we trim the input value using `sanitizeInput` which the update action then passes back through the passed-in value resulting in spaces at the end of the input text always being lost.

PR #274, opened 2016-09-20
Long theme names break themes table layout (#7373)

zewish:
fixed

### DIFF
--- a/app/styles/layouts/settings.css
+++ b/app/styles/layouts/settings.css
@@ -188,8 +188,10 @@
 }
 
 .theme-list-item-body .name {
-    display: inline-block;
+    display: block;
+    overflow: hidden;
     color: var(--darkgrey);
+    text-overflow: ellipsis;
     white-space: nowrap;
     font-weight: 400;
     user-select: all;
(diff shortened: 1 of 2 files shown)


kevinansfield:
Thanks @zewish, at first glance your fix looks good, I'll get around to testing it properly later 😄 In the meantime would you mind updating your commit and PR content to match our [commit message format](https://github.com/TryGhost/Ghost/wiki/Git-workflow#notes-on-writing-good-commit-messages)?

zewish:
Commit message updated.

KennethAshley:
@zewish Nice fix. I didn't even think to use `text-overflow: ellipsis;`. 

Here was my css only flexbox fix.

Desktop
image

Mobile
image

kevinansfield:
@zewish, @KennethAshley thanks to both of you for taking a stab at this! 😄 After reviewing and seeing the results of @KennethAshley's approach I think that is probably the better direction as it doesn't hide any information, especially as the version numbers are the parts likely to be hidden using ellipsis and they are what's used to differentiate duplicate theme names in the list.

@KennethAshley are you able to PR the approach that you took?

PR #277, opened 2016-09-20
Separate invites and users

kevinansfield:
issue https://github.com/TryGhost/Ghost/issues/7420, requires https://github.com/TryGhost/Ghost/pull/7422
- adds a new `Invite` model with associated serializer and test setup
- updates team screen to use invites rather than existing users with the "invited" property
- updates signup process to work with new invite model
- updates setup process to create invites instead of users
- swaps usage of `gh-select-native` for `one-way-select` in the invite modal so that attributes can be set on the `select` element
- updates resend invite process to account for server returning a new model
- rewrites the invite management tests and fixes mirage mocks for invite endpoints
- sorts invites by email address to avoid jumping invites when re-sending

TODO:
- [x] re-instate invite validation when accessing signup page
- [x] add tests for resend + revoke actions
- [x] add tests for role display
- [x] add tests for authors accessing team page (mirage needs to return 401 if author accesses invites endpoint)
- [x] 🐛 resend doesn't account for getting a new ID back from the server

### DIFF
--- a/app/components/gh-user-invited.js
+++ b/app/components/gh-user-invited.js
@@ -1,36 +1,48 @@
 import Component from 'ember-component';
 import computed from 'ember-computed';
 import service from 'ember-service/inject';
+import {isNotFoundError} from 'ember-ajax/errors';
 
 export default Component.extend({
     tagName: '',
 
-    user: null,
+    invite: null,
     isSending: false,
 
     notifications: service(),
+    store: service(),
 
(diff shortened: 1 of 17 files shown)


kevinansfield:
ready for review 😎

ErisDS:
Have tested this and all the behaviour seems to be working as far as I can tell.

PR #281, opened 2016-09-21
Refactor `gh-fullscreen-modal` component

kevinansfield:
no issue
- remove `liquid-tether` dependency
- add `liquid-wormhole` dependency (previously a dependency of liquid-tether)
- refactor `gh-fullscreen-modal` to simplify and use `liquid-wormhole` directly instead of `liquid-tether` as we don't need the positioning ability that `liquid-tether` provides
- fixes broken View/Component tree in Ember Inspector 🎉

### DIFF
--- a/app/components/gh-fullscreen-modal.js
+++ b/app/components/gh-fullscreen-modal.js
@@ -4,42 +4,34 @@ import {A as emberA} from 'ember-array/utils';
 import {isBlank} from 'ember-utils';
 import on from 'ember-evented/on';
 import run from 'ember-runloop';
-
-import LiquidTether from 'liquid-tether/components/liquid-tether';
 import {invokeAction} from 'ember-invoke-action';
+import computed from 'ember-computed';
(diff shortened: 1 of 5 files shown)


kevinansfield:
I've refactored/simplified our modal implementation a bit so that we can work with the liquid-tether recommended method of background overlays.

Unfortunately this implementation puts the overlay and the modal in two disparate containers so it's no longer possible to have both the modal scrollable and overlay clickable at the same time. I'm in discussions with liquid-tether's developer in Ember's #topic-animation slack channel to see if there's a way we can resolve this.

An alternative method may be to drop `liquid-tether` altogether and rely on `liquid-wormhole` and some CSS to ensure that modals are centred properly.

kevinansfield:
Ready for review

disordinary:
LGTM and may I add a 🍾 + 🍰  + 👯‍♂️ + 🎆

PR #283, opened 2016-09-21
use ember-ajax in place of ember-data's networking

kevinansfield:
closes TryGhost/Ghost#7014
- uses the `AjaxServiceSupport` mixin from `ember-ajax` to replace Ember Data's internal `ajax` method with our own ajax service
- normalizes all error handling to use `ember-ajax` style errors
- default to the `application/vnd.api+json` content-type so that we don't have a mix of urlencoded and plain JSON content

TODO:
- [x] error responses are being passed through the standard normalisation methods in the serialisers resulting in a lot of console logs containing: `WARNING: Encountered "errors" in payload, but no model was found for model name "error" (resolved model name using ghost-admin@serializer:user:.modelNameFromPayloadKey("errors"))`
- [x] change Content-Type back to `application/json` to appease the server and fix url-encoding of JSON data


kevinansfield:
Ready for review 🎱

acburdine:
LGTM 👍

PR #286, opened 2016-09-23
replace gh-select-native with one-way-select

kevinansfield:
no issue
- removes `gh-select-native` component as ember-one-way-control's `one-way-select` component offers the same one-way-binding but also allows element attributes to be set


app/templates/components/modals/invite-new-user.hbs
```diff
@@ -27,12 +27,13 @@
...
-                    content=roles
+                {{one-way-select
+                    id="new-user-role"
+                    options=roles
                     optionValuePath="id"
                     optionLabelPath="name"
-                    selection=role
-                    action="setRole"
+                    value=role
+                    action=(action "setRole")
```
acburdine: needs to be `update=(action "setRole")`

app/templates/team/user.hbs
```diff
@@ -115,12 +115,13 @@
...
-                                content=roles
+                            {{one-way-select
+                                id="new-user-role"
+                                options=roles
                                 optionValuePath="id"
                                 optionLabelPath="name"
-                                selection=model.role
-                                action="changeRole"
+                                value=model.role
+                                action=(action "changeRole")
```
acburdine: same here

app/templates/components/modals/invite-new-user.hbs
kevinansfield: Interesting, I wonder why that didn't cause tests to fail

kevinansfield:
Typos fixed, I think those areas weren't tested particularly well. I think if this was applied on top of #277 then the new tests would have at least picked up the invite modal typo.

PR #288, opened 2016-09-23
Safari doesn't show dragged-over state for file uploads

KennethAshley:
# Why?
In reference to [#7310](https://github.com/TryGhost/Ghost/issues/7310)

This is not an Ember issue. Safari ignores classes that begin with `--`. I believe that none of the IEs allow these types of classes as well.

Example:
- https://jsfiddle.net/oy75f0L3/2/

Sources: 
- https://twitter.com/logicoder/status/654897918357434368
- https://www.reddit.com/r/web_design/comments/4ji5p7/safari_completely_ignores_classnames_starting/

# What Changed?

Globally searched for modifier classes that start with `--` and removed the first dash. 

# Example

safari-drag

closes #7310
- Removes double dashes from modifier classes as safari won’t allow
this.

### DIFF
--- a/app/components/gh-file-uploader.js
+++ b/app/components/gh-file-uploader.js
@@ -108,7 +108,7 @@ export default Component.extend({
         event.stopPropagation();
         event.preventDefault();
 
-        this.set('dragClass', '--drag-over');
+        this.set('dragClass', '-drag-over');
     },
 
     dragLeave(event) {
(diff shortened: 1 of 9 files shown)


kevinansfield:
@KennethAshley thanks for this! TIL 😁 Never knew Safari ignored classes prefixed with `--`

PR #291, opened 2016-09-25
Mobile-Doc based editor

disordinary:
[WIP] Mobile-Doc based renderer

Refs #7429

Depends On [#7437](https://github.com/TryGhost/Ghost/pull/7437)

Added mobiledoc renderer

-------------------

- Added generic mobiledoc-renderer
- Kept the existing showdown renderer for legacy mode

This is the first integration of ghost-editor. It's styled enough to work, however it is not anywhere approaching something that looks remotely like what the finished thing will be.

Early ALPHA, development build. Tread cautiously.


tests/unit/components/gh-notification-test.js
```diff
@@ -7,44 +7,44 @@ import {
...
-        unit: true
-        // specify the other units that are required for this test
-        // needs: ['component:foo', 'helper:bar']
-    },
-    function () {
-        it('closes notification through notifications service', function () {
-            let component = this.subject();
-            let notifications = {};
-            let notification = {message: 'Test close', type: 'success'};
+// describeComponent(
```
kevinansfield: Does this test need to be commented out? It should be untouched by the other changes in this PR, interested to hear if it's causing issues

tests/unit/transforms/json-string-test.js
```diff
@@ -0,0 +1,27 @@
+import { expect } from 'chai';
+import { describeModule, it } from 'ember-mocha';
+import Post from 'ghost-admin/models/post';
+
+describeModule(
+    'transform:json-string',
+    'Unit | Transform | json string',
```
kevinansfield: Very minor thing but could this be updated to match the naming scheme of the other tests? `Unit: Transform: json-string` (it helps to be consistent so that test filters pick up everything that's expected)

ember-cli-build.js
```diff
@@ -137,8 +137,11 @@ module.exports = function (defaults) {
     app.import('bower_components/google-caja/html-css-sanitizer-bundle.js');
     app.import('bower_components/jqueryui-touch-punch/jquery.ui.touch-punch.js');
 
+
+
     if (app.env === 'test') {
         app.import(app.bowerDirectory + '/jquery.simulate.drag-sortable/jquery.simulate.drag-sortable.js', {type: 'test'});
+        app.import('ghost-admin/transforms/json-string.js');
```
kevinansfield: This import shouldn't be needed as Ember's resolver should pick it up automatically, if it was causing issues in testing perhaps it was missing from the `needs` array in unit tests?

package.json
```diff
@@ -30,6 +30,7 @@
     "broccoli-asset-rev": "2.4.6",
     "broccoli-clean-css": "1.1.0",
     "broccoli-concat": "3.0.2",
+    "broccoli-funnel": "^1.0.7",
```
kevinansfield: Does this dependency need to be added here or is it a sub-dependency of the `ghost-editor` package?

disordinary: Yes, removed in latest PR.

tests/unit/components/gh-editor-save-button-test.js
```diff
@@ -29,4 +29,4 @@ describeComponent(
             expect(component._state).to.equal('inDOM');
         });
     }
-);
+);
```
disordinary: Done.

disordinary:
Thanks for the feedback Kevin,

I think I've addressed all of them.

Tabbing off the title will return, I'd also like to be able to navigate from the main body to the title with up and down and left and right key presses. Originally I wanted to have the title and body within the one undo queue but that's not possible with mobiledoc as it stands right now.

PR #300, opened 2016-10-03
Replace gh-select-native with one-way-select

acburdine:
This fixes the issue introduced with the switch to one-way-select, and essentially reverts the reversion that was done before the 0.11.1 release


acburdine:
@kevinansfield do we still need this in LTS? If not feel free to close this 😄

PR #307, opened 2016-10-05
deps: ember-light-table@1.4.0

kevinansfield:
no issue
- update `ember-light-table` to 1.4.0
- remove `gh-light-table` override component as scrolling has been fixed
- disable the infinite scroll acceptance test for now because Chrome has issues detecting the position of the scroll trigger when the app is rendered into the zoomed container during testing

I have opened [an issue](https://github.com/offirgolan/ember-light-table/issues/201) regarding the scroll behaviour when acceptance testing, I think for now it's safe to remove the test. If no solution is forthcoming and we are concerned about the wiring we can always test the related components in isolation to ensure the correct calls are being made.


disordinary:
LGTM 👍

PR #309, opened 2016-10-06
🎨  Change asset path to /ghost/assets

ErisDS:
See TryGhost/Ghost#7503 for a hopefully clear explanation of why I'm making this change 

I have changed to having both an `admin` and an `asset` path because, although I can't find any usages of `admin`, I didn't want to hijack the term "admin" to mean "asset".

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -216,7 +216,7 @@ module.exports = function(grunt) {
         ).then(function (results) {
             var contributors = mergeContribs(results[1], results[2]),
                 contributorTemplate = '<article>\n    <a href="<%= githubUrl %>" title="<%= name %>">\n' +
-                    '        <img src="{{gh-path "admin" "/img/contributors"}}/<%= name %>" alt="<%= name %>" />\n' +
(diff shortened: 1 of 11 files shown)


ErisDS:
This passes the tests, however it was screwing up code injection because codemirror is included using the lazyLoader.

I've fixed this & pushed up my fix, however the concatenated `codemirror.js` file has a source map url at the bottom:

`//# sourceMappingURL=codemirror.map`

And this is wrong :(

I have no idea how to make the url be correct. I tried looking at various packages that seem to be used, but none of them appear to have any documentation on how to configure them! 

Any ideas?

kevinansfield:
I believe you can add this line to the [codemirror concat config](https://github.com/TryGhost/Ghost-Admin/blob/master/ember-cli-build.js#L42-L46) to specify the relative path for the sourcemap:

```
sourceMapConfig: {sourceRoot: 'proper/path/'}
```

`broccoli-concat` uses https://github.com/mozilla/source-map under the hood so any options passed in via `sourceMapConfig` are passed through to that.

kevinansfield:
Alternatively, that source map isn't particularly useful so it could be set to `sourceMapConfig: {enabled: false}`

ErisDS:
Switched to remove the sourcemap ⛵️

PR #327, opened 2016-10-12
Attempt expired session restore on network requests

kevinansfield:
issue TryGhost/Ghost#5202

We can get into a situation where the app is left open without a network connection and the token subsequently expires, this will result in the next network request returning a 401 and killing the session. This is an attempt to detect that and restore the session using the stored refresh token before continuing with the request

- wrap ajax requests in a session restore request if we detect an expired `access_token`

### DIFF
--- a/app/services/ajax.js
+++ b/app/services/ajax.js
@@ -133,10 +133,36 @@ let ajaxService = AjaxService.extend({
     // ember-ajax recognises `application/vnd.api+json` as a JSON-API request
     // and formats appropriately, we want to handle `application/json` the same
     _makeRequest(hash) {
+        let isAuthenticated = this.get('session.isAuthenticated');
+        let isGhostRequest = hash.url.indexOf('/ghost/api/') !== -1;
+        let isTokenRequest = isGhostRequest && hash.url.match(/authentication\/(?:token|ghost)/);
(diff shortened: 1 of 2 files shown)


ErisDS:
The bug/todo in the code here is due to this: https://github.com/simplabs/ember-simple-auth/blob/master/addon/authenticators/oauth2-password-grant.js#L345 am I right?
It's something that was discussed as "improvement welcome" I think: https://github.com/simplabs/ember-simple-auth/issues/980#issuecomment-233266886

kevinansfield:
It shouldn't be due to that ESA issue as I've overridden `makeRequest` to use our [own ajax service](https://github.com/TryGhost/Ghost-Admin/blob/master/app/authenticators/oauth2.js#L33-L46) so that we get consistent ajax error objects 🤔

kevinansfield:
Ignore the above, I was looking at completely the wrong thing. I've previously overridden the `authenticate` method to return a proper AjaxError object but not the `_refreshAccessToken` method.

I'm not sure if it's something we want to do or not, the more we override the trickier upgrades are. The fact that `session.invalidate()` is called when we get a 401 means that the original caller wouldn't have been able to do anything with the error response in the past so I think better to leave behaviour as-is and leave the comment in place?

PR #330, opened 2016-10-13
✍ Change contributors on about page to be manually

AileenCGN:
refs TryGhost/Ghost#7427

Removes `-contributors.hbs` and `public/assets/img/contributors` from `.gitignore`.
Removes `buildAboutPage` from `grunt init` task.
Updates `contributors.hbs` and adds images for each contributor.
Changes the `max-width` of `.gh-credits` wrapper to be 700px instead of 650px, so we'll fit them all in two rows.


AileenCGN:
This is ready to go! 🎉

PR #331, opened 2016-10-13
🔥  Remove grunt contributor

AileenCGN:
refs TryGhost/Ghost#7427

Removes unneccessary grunt dependencies from `package.json` and deletes contributor part in `Gruntfile.js`

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -4,7 +4,6 @@ var _              = require('lodash'),
     fs             = require('fs-extra'),
     path           = require('path'),
     https          = require('https'),
-    getTopContribs = require('top-gh-contribs'),
     moment         = require('moment'),
     chalk          = require('chalk'),
     Promise        = require('bluebird');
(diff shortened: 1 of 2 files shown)


ErisDS:
Note: after this is merged, we will need a temporary quick-fix to replace `subgrunt:init` in the Ghost Gruntfile.js where (and only where) it is needed for the `grunt release` task that hasn't been rewritten yet.

Definitely want to rewrite the release build, but that's a bigger task and we don't want to block the alpha release ;)

kevinansfield:
I'm moving this to the next milestone as we still need to resolve the linting issue.

kevinansfield:
Pushed back again 😞 I'm looking at getting linting running in our test suite as part of https://github.com/TryGhost/Ghost-Admin/pull/357

kevinansfield:
@AileenCGN does this need revisiting a little? We still have a `Gruntfile.js` but no `grunt` dependencies available to run any of the commands - is that intended?

AileenCGN:
Yeah, fair point, @kevinansfield 😬

AileenCGN:
This is ready to go now! 🎉
We will need another PR to finish this task:
> we will need a temporary quick-fix to replace subgrunt:init in the Ghost Gruntfile.js where (and only where) it is needed for the grunt release task that hasn't been rewritten yet.

PR #342, opened 2016-10-17
add manual pretender shutdown to fix acceptance tests

kevinansfield:
no issue
- ensure our mirage server is shutdown during app teardown in acceptance tests
- a [recent update](https://github.com/pretenderjs/pretender/pull/178) to Pretender contained a breaking change that throws an error when multiple pretender instances exist
- using mirage in acceptance tests was [triggering the error](https://github.com/samselikoff/ember-cli-mirage/issues/915)

### DIFF
--- a/tests/helpers/destroy-app.js
+++ b/tests/helpers/destroy-app.js
@@ -1,5 +1,10 @@
 import run from 'ember-runloop';
 
 export default function destroyApp(application) {
+    // this is required to fix "second Pretender instance" warnings
+    if (server) {
+        server.shutdown();
+    }
+
     run(application, 'destroy');
 }


robneville73:
@kevinansfield tried to do the same thing and it's not fixing the issue...any idea what I might be missing?

trek:
Pretender 1.4.1 is released and only warns now. To be fair: starting several pretender servers has never been allowed and shutting down is specifically called out in the README as a requirement. We only started preventing it because people forgetting this requirement was causing other issues.

acburdine:
@kevinansfield Do we still want this? If so it might be a good idea to update the TODO comment :)

acburdine:
@kevinansfield I can either merge this now or close it as the same change is present in #37 as part of the mirage update. I'm fine with either :)

kevinansfield:
@acburdine let's merge this now as it makes the test logs on Travis rather noisy. I've updated the comment to remove the `TODO` as the upstream "fix" was to make it a warning rather than an error meaning this is still necessary

PR #355, opened 2016-10-21
New editor layout

disordinary:
This bumps the editor to 14 (13 wasn't so lucky).

Requires: https://github.com/TryGhost/Ghost/pull/7619
Refs: https://github.com/TryGhost/Ghost/issues/7429

It includes:
- The title is integrated into the content.
- The toolbar has been revamped, that annoying under paragraph toolbar has been removed in favor of a block level changed (on the left hand side of the content).
- The toolbar also has the ability to create links.
- Additional markdown support within the editor including `# h1 headings`, `## h2 headings`, `### h3 headings`, `1. lists`, `* lists`, `- lists`, `> blockquote`, `_italic_`, `*italic*`, `__bold__`, `**bold**`, `~~strikethrough~~`, `[links](http://www.ghost.org/)`, `images` (warning will replace the entire block).

Footnotes and codeblocks are on our todo, we won't be supporting inline HTML, however there are markdown and HTML cards which can be inserted into the content.

Whew, it's getting there. What you see is still placeholder styling - but we're slowly refining towards something that's useful. 🍾

### DIFF
--- a/app/mixins/editor-base-controller.js
+++ b/app/mixins/editor-base-controller.js
@@ -16,6 +16,8 @@ import PostModel from 'ghost-admin/models/post';
 import boundOneWay from 'ghost-admin/utils/bound-one-way';
 import {isVersionMismatchError} from 'ghost-admin/services/ajax';
 import {isInvalidError} from 'ember-ajax/errors';
+import $ from 'jquery';
+import ghostPaths from 'ghost-admin/utils/ghost-paths';
 
 const {resolve} = RSVP;
 
@@ -39,6 +41,12 @@ export default Mixin.create({
     notifications: injectService(),
     clock: injectService(),
     slugGenerator: injectService(),
+
+    cards: [], // for apps
+    atoms: [], // for apps
+    toolbar: [], // for apps
+    apiRoot: ghostPaths().apiRoot,
+    assetPath: ghostPaths().assetRoot,
 
     init() {
         this._super(...arguments);
@@ -543,6 +551,13 @@ export default Mixin.create({
 
         toggleReAuthenticateModal() {
             this.toggleProperty('showReAuthenticateModal');
+        },
+
+        titleKeyDown(event) {
+            if (event.keyCode === 13 || event.keyCode === 40) {
+                // if the enter key or down key are pressed then focus on the editor
+                $('.__mobiledoc-editor').focus();
+            }
         }
     }
 });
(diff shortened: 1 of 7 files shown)


app/mixins/editor-base-controller.js
```diff
@@ -331,7 +331,16 @@ export default Mixin.create({
 
         notifications.showAlert(message, {type: 'error', delayed: delay, key: 'post.save'});
     },
-
+    keyDown: task(function* (event) {
+        switch (event.keyCode) {
+            // if the enter key or down key are pressed then focus on the editor
+            case 13:
+            case 40:
+                Ember.$('.__mobiledoc-editor').focus();
```
disordinary: I should add it's triggered by the keyDown action on gh-trim-focus-input within edit.hbs

kevinansfield: If you want access to jQuery you can use `import $ from 'jQuery'` if that will help with the linting problems.

The `editor-base-controller` mixin is unfortunately a bit of a mess, in many older areas of our codebase we have controllers taking on the responsibilities of a component.

I may have missed it but what is the reasoning behind using an ember-concurrency task here? As far as I can see there's no asynchrony involved - would it make more sense for this to be a standard action?

tests/helpers/destroy-app.js
```diff
@@ -1,5 +1,6 @@
 import run from 'ember-runloop';
 
 export default function destroyApp(application) {
+    server.shutdown();
```
kevinansfield: Would you mind removing this line from the PR? I'd prefer if that change went in separately with some notes about why it's being made, especially as it's something that should be handled by `ember-cli-mirage`

app/routes/editor/new.js
```diff
@@ -36,6 +37,12 @@ export default AuthenticatedRoute.extend(base, {
         psm.send('resetPubDate');
 
         this._super(...arguments);
+
+        controller.set('cards' , []);
+        controller.set('atoms' , []);
+        controller.set('toolbar' , []);
+        controller.set('apiRoot', ghostPaths().apiRoot);
```
acburdine: it would probably be better to inject the `ghostPaths` service into the controller rather than re-set the API and asset paths each time the route is entered.

app/mixins/editor-base-controller.js
disordinary: I was using a generator mainly for consistency and to keep code grouped together, will adjust.

So yes having the mixin behave like  a component is what we expect...

Thanks.

app/routes/editor/new.js
disordinary: Yep,

As Kevin said above the whole editor is a bit of a mess and it's not 100% clear that everything is behaving as it should. I'll also adjust the edit route.

gurde:
How will the editor handle youtube embeds?

rawsh:
@gurde In the initial post (referenced at the top) they said that you can paste html code as an embed. It's unlikely that there's going to be any specific embedding.

PR #385, opened 2016-11-07
Updated ghost-editor to 0.1.4

disordinary:
For some reason greenkeeper isn't upgrading ghost-editor. This upgrades it to the latest version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,7 +74,7 @@
     "ember-wormhole": "0.5.0",
     "emberx-file-input": "1.1.0",
     "fs-extra": "1.0.0",
-    "ghost-editor": "0.1.1",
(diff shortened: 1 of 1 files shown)


kevinansfield:
> For some reason greenkeeper isn't upgrading ghost-editor.

I disabled it for `ghost-editor` as we seemed to be getting duplicate PRs (greenkeeper+manual) for each release, are we at a stage where greenkeeper makes more sense?

disordinary:
@kevinansfield yeah I think it's a good idea, the API is locked so a change in Ghost-Editor doesn't need a corresponding change in Ghost-Admin, I also try and do the release at the same time as the alpha.

PR #394, opened 2016-11-11
🎨  invites roles table into a field on the invites table

kirrg001:
refs https://github.com/TryGhost/Ghost/issues/7432

- invite.role instead of invite.roles
- be able to serialize/deserialize response from and request to server
- server returns role_id

This is my first touch with Ember, so be nice to me 😃 👻

- [x] test with Ghost PR https://github.com/TryGhost/Ghost/pull/7705 (all the general features work: create invite, resend, invoke and list invites)
- [x] check if tests are green
- [x] final testing on Monday after PR review (includes signup as invited user)

### DIFF
--- a/app/models/invite.js
+++ b/app/models/invite.js
@@ -1,3 +1,5 @@
+/* eslint camelcase: [2, {properties: "never"}] */
+/* jscs:disable requireCamelCaseOrUpperCaseIdentifiers */
 import Model from 'ember-data/model';
 import attr from 'ember-data/attr';
 import {belongsTo} from 'ember-data/relationships';
@@ -19,9 +21,10 @@ export default Model.extend({
 
     resend() {
         let fullInviteData = this.toJSON();
+
         let inviteData = {
             email: fullInviteData.email,
-            role: fullInviteData.role
+            role_id: fullInviteData.role
         };
 
         let inviteUrl = this.get('ghostPaths.url').api('invites');
--- a/app/models/invite.js
+++ b/app/models/invite.js
@@ -1,7 +1,6 @@
 import Model from 'ember-data/model';
 import attr from 'ember-data/attr';
-import {hasMany} from 'ember-data/relationships';
-import computed from 'ember-computed';
+import {belongsTo} from 'ember-data/relationships';
 import injectService from 'ember-service/inject';
 
 export default Model.extend({
@@ -13,33 +12,18 @@ export default Model.extend({
     updatedAtUTC: attr('moment-utc'),
     updatedBy: attr('number'),
     status: attr('string'),
-    roles: hasMany('role', {
-        embedded: 'always',
-        async: false
-    }),
+    role: belongsTo('role', {async: false}),
 
     ajax: injectService(),
     ghostPaths: injectService(),
-
-    role: computed('roles', {
-        get() {
-            return this.get('roles.firstObject');
-        },
-        set(key, value) {
-            // Only one role per user, so remove any old data.
-            this.get('roles').clear();
-            this.get('roles').pushObject(value);
-
-            return value;
-        }
-    }),
 
     resend() {
         let fullInviteData = this.toJSON();
         let inviteData = {
             email: fullInviteData.email,
-            roles: fullInviteData.roles
+            role: fullInviteData.role
         };
+
         let inviteUrl = this.get('ghostPaths.url').api('invites');
 
         return this.get('ajax').post(inviteUrl, {
--- a/app/routes/team/index.js
+++ b/app/routes/team/index.js
@@ -37,8 +37,8 @@ export default AuthenticatedRoute.extend(styleBody, CurrentUserSettings, Paginat
 
             // we need to load the roles into ember cache
             // invites return role_id only and we do not offer a /role/:id endpoint
-            this.get('store').query('role', {permissions: 'assign'}).then((roles) => {
-                modelPromises.roles = roles;
+            modelPromises.roles = this.get('store').query('role', {permissions: 'assign'}).then((roles) => {
+                return roles;
             });
 
             return RSVP.hash(modelPromises);
--- a/app/routes/team/index.js
+++ b/app/routes/team/index.js
@@ -37,7 +37,7 @@ export default AuthenticatedRoute.extend(styleBody, CurrentUserSettings, Paginat
 
             // we need to load the roles into ember cache
             // invites return role_id only and we do not offer a /role/:id endpoint
-            modelPromises.roles = this.get('store').query('role', {permissions: 'assign'}).then((roles) => {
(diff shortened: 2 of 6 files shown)


app/routes/team/index.js
```diff
@@ -35,6 +35,12 @@ export default AuthenticatedRoute.extend(styleBody, CurrentUserSettings, Paginat
                 });
             }
 
+            // we need to load the roles into ember cache
+            // invites return role_id only and we do not offer a /role/:id endpoint
+            this.get('store').query('role', {permissions: 'assign'}).then((roles) => {
```
kevinansfield: This should be `modelPromises.roles = this.get('store').query('role', ...` and the promise body should just `return roles;`.

This is because the `model()` hook will pause for the returned promises to be resolved before continuing so when returning `RSVP.hash` you end up with a hash containing the values returned from the promises so the controller and everything else relying on the `model` object doesn't need to worry about things being async.

kirrg001:
I have added the change you requested.
Can you give me a 👍 if this PR is ready to merge?
Then i will start with final tests.

app/serializers/invite.js
```diff
@@ -1,10 +1,24 @@
+/* jscs:disable requireCamelCaseOrUpperCaseIdentifiers */
 import ApplicationSerializer from 'ghost-admin/serializers/application';
 import EmbeddedRecordsMixin from 'ember-data/serializers/embedded-records-mixin';
 
 export default ApplicationSerializer.extend(EmbeddedRecordsMixin, {
     attrs: {
-        roles: {embedded: 'always'},
+        role: {embedded: 'always'},
```
kevinansfield: This should be removed I think, the `{embedded: 'always'}` option is used when the API returns a full embedded record object whereas now we are only dealing with an ID so there's no special behaviour needed. The `EmbeddedRecordsMixin` can also be removed.

app/routes/team/index.js
```diff
@@ -35,6 +35,12 @@ export default AuthenticatedRoute.extend(styleBody, CurrentUserSettings, Paginat
                 });
             }
 
+            // we need to load the roles into ember cache
+            // invites return role_id only and we do not offer a /role/:id endpoint
+            modelPromises.roles = this.get('store').query('role', {permissions: 'assign'}).then((roles) => {
```
kevinansfield: Is the `{permissions: 'assign'}` needed here? It would be preferable to load all roles so that the role of team members can be displayed even if the current user doesn't have permission to assign those roles to new users.

app/models/invite.js
```diff
@@ -13,33 +12,18 @@ export default Model.extend({
     updatedAtUTC: attr('moment-utc'),
     updatedBy: attr('number'),
     status: attr('string'),
-    roles: hasMany('role', {
-        embedded: 'always',
-        async: false
-    }),
+    role: belongsTo('role', {async: false}),
```
kevinansfield: Are roles retrievable from the API via an ID (e.g. `/api/roles/:id/`)? If so then we can remove the `{async: false}` option - it's required for embedded records but when we only get an ID it's safer to go with the default behaviour and allow async retrieval of the associated role model.

app/serializers/invite.js
kevinansfield: Changing it to `role: {key: 'role_id'}` removes the need for the custom `normalize` and `serialize` functions below.

app/models/invite.js
```diff
@@ -13,33 +12,18 @@ export default Model.extend({
...
-            return value;
-        }
-    }),
-
     resend() {
         let fullInviteData = this.toJSON();
         let inviteData = {
             email: fullInviteData.email,
-            roles: fullInviteData.roles
+            role: fullInviteData.role
```
kevinansfield: This needs changing to `role_id: fullInviteData.role` otherwise the server complains that "role is required"

kirrg001: As far as i understood Ember, this get's called:
https://github.com/TryGhost/Ghost-Admin/pull/394/files#diff-4988c6daa641a939ebddb019fbf5d3dcR18

That's why no need to use role_id?

app/serializers/invite.js
kirrg001: > Changing it to role: {key: 'role_id'} removes the need for the custom normalize and serialize functions below.

Wasn't working for me :/

app/models/invite.js
kevinansfield: No, here we're using ajax directly:

```
let inviteUrl = this.get('ghostPaths.url').api('invites');

        return this.get('ajax').post(inviteUrl, {
            data: JSON.stringify({invites: [inviteData]}),
            contentType: 'application/json'
        });
```

It's not going through Ember Data and so none of the models or serialisers are touched.

kirrg001:
See PR TODO's:
`final testing on Monday after PR review (includes signup as invited user)`

I only made a basic test, because i knew Ember changes get required :D

app/models/invite.js
kirrg001: Ok thanks. I need to test this behaviour again 👍 I remember i had trouble getting the test green: `tests/unit/models/invite-test.js` and this was related. Will come back

app/routes/team/index.js
kirrg001: To be honest: i copied this query over from somewhere else. I will remove it 👍

app/routes/team/index.js
```diff
@@ -35,6 +35,12 @@ export default AuthenticatedRoute.extend(styleBody, CurrentUserSettings, Paginat
                 });
             }
 
+            // we need to load the roles into ember cache
+            // invites return role_id only and we do not offer a /role/:id endpoint
+            modelPromises.roles = this.get('store').query('role', {}).then((roles) => {
```
kevinansfield: Doesn't matter in this case but for future reference `this.get('store').query('role', {})` would [normally be written](https://guides.emberjs.com/v2.9.0/models/finding-records/#toc_retrieving-multiple-records) as `this.get('store').findAll('role')`

tests/unit/models/invite-test.js
```diff
@@ -76,11 +60,9 @@ describeModel(
                 ).to.equal(1);
 
                 expect(invite.email).to.equal('resend-test@example.com');
-                expect(
-                    invite.roles.length,
-                    'number of roles in request body'
-                ).to.equal(1);
-                expect(invite.roles[0], 'role ID').to.equal('1');
+                /* jscs:disable requireCamelCaseOrUpperCaseIdentifiers */
```
kevinansfield: Ghost-Admin is using `eslint` now so this will probably fail the lint tests when merged with master. `eslint` has a "next-line" mode to avoid needing to wrap single lines in disable/enable statements:

`// eslint-disable-next-line camelcase`

ErisDS:
The server side version of this is merged :)

kevinansfield:
👍 Testing this at the moment

PR #397, opened 2016-11-14
✨  Blog icon upload

AileenCGN:
refs TryGhost/Ghost#7688
needs TryGhost/Ghost#7893

Adds new upload functionality for a blog icon in general settings.
Icons will be uploaded to a new endpoint `uploads/icons` to trigger different validations.

Todo:
- [X] add client side validation
- [X] remove form switch for URL upload (only for blog icon)
- [ ] get feedback/spec for design & wording
- [X] write tests

### DIFF
--- a/app/components/gh-image-uploader.js
+++ b/app/components/gh-image-uploader.js
@@ -24,8 +24,10 @@ export default Component.extend({
     text: '',
     altText: '',
     saveButton: true,
-    accept: 'image/gif,image/jpg,image/jpeg,image/png,image/svg+xml',
-    extensions: ['gif', 'jpg', 'jpeg', 'png', 'svg'],
+    accept: null,
+    extensions: null,
+    uploadUrl: null,
+    allowUrlInput: true,
     validate: null,
 
     dragClass: null,
@@ -38,6 +40,10 @@ export default Component.extend({
     ajax: injectService(),
     config: injectService(),
     notifications: injectService(),
+
+    _defaultAccept: 'image/gif,image/jpg,image/jpeg,image/png,image/svg+xml',
+    _defaultExtensions: ['gif', 'jpg', 'jpeg', 'png', 'svg'],
+    _defaultUploadUrl: '/uploads/',
 
     // TODO: this wouldn't be necessary if the server could accept direct
     // file uploads
@@ -83,6 +89,16 @@ export default Component.extend({
     didReceiveAttrs() {
         let image = this.get('image');
         this.set('url', image);
+
+        if (!this.get('accept')) {
+            this.set('accept', this.get('_defaultAccept'));
+        }
+        if (!this.get('extensions')) {
+            this.set('extensions', this.get('_defaultExtensions'));
+        }
+        if (!this.get('uploadUrl')) {
+            this.set('uploadUrl', this.get('_defaultUploadUrl'));
+        }
     },
 
     dragOver(event) {
@@ -161,7 +177,10 @@ export default Component.extend({
         }
 
         if (isUnsupportedMediaTypeError(error)) {
-            message = 'The image type you uploaded is not supported. Please use .PNG, .JPG, .GIF, .SVG.';
+            let validExtensions = this.get('extensions').join(', .').toUpperCase();
+            validExtensions = `.${validExtensions}`;
+
+            message = `The image type you uploaded is not supported. Please use ${validExtensions}`;
         } else if (isRequestEntityTooLargeError(error)) {
             message = 'The image you uploaded was larger than the maximum file size your server allows.';
         } else if (error.errors && !isBlank(error.errors[0].message)) {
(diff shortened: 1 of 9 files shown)


app/controllers/settings/general.js
```diff
@@ -24,10 +25,19 @@ export default Controller.extend(SettingsSaveMixin, {
     _scratchFacebook: null,
     _scratchTwitter: null,
 
+    imgMimeTypes: 'image/gif,image/jpg,image/jpeg,image/png,image/svg+xml',
```
kevinansfield: These shouldn't need to be duplicated, they are the same as the defaults in `gh-image-uploader`.

app/templates/components/gh-image-uploader.hbs
```diff
@@ -19,10 +19,11 @@
                 <div class="description">{{description}}</div>
             {{/gh-file-input}}
         </div>
-
-        <a class="image-url" {{action "switchForm" "url-input"}}>
-            <i class="icon-link"><span class="hidden">URL</span></i>
-        </a>
+        {{#if allowUrlInput}}
```
kevinansfield: This may need testing with `fileStorage` being set to `false` in the config. If file storage is disabled we'll always display the url input form rather than the upload form - is that the desired behaviour?

app/templates/settings/general.hbs
```diff
@@ -39,13 +39,30 @@
 
                 {{#if showUploadLogoModal}}
                     {{gh-fullscreen-modal "upload-image"
-                            model=(hash model=model imageProperty="logo")
+                            model=(hash model=model imageProperty="logo" accept=imgMimeTypes extensions=imgExtensions allowUrlInput=true)
```
kevinansfield: Is it necessary to set the `accept` and `extensions` properties here when they are the same as the component default values?

app/controllers/settings/general.js
AileenCGN: That's what I thought, too.. But the normal image upload didn't work anymore without setting them explicitly... Maybe there's another way to fix that?

app/templates/components/gh-image-uploader.hbs
AileenCGN: No, I wasn't aware of that. Thanks!

app/templates/settings/general.hbs
AileenCGN: Same problem as mentioned above. It didn't work without...

kevinansfield: Taking a deeper look at this I think it's because inside `gh-fullscreen-modal` the attributes are always passed on to the `gh-image-uploader` component so if nothing is passed in to the modal component the defaults on the uploader component will be overridden with `undefined`.

One way around it would be to use the `didReceiveAttrs` hook in the `gh-image-uploader` to check if the provided attributes are null or undefined then set them to the defaults, e.g.:

```
accept: null,
extensions: null,

_defaultAccept: 'image/gif,image/jpg,image/jpeg,image/png,image/svg+xml',
_defaultExtensions: ['gif', 'jpg', 'jpeg', 'png', 'svg'],

didReceiveAttrs() {
    if (!this.get('accept') { this.set('accept', this.get('_defaultAccept')); }
    if (!this.get('extensions') { this.set('extensions', this.get('_defaultExtensions')); }
}
```

AileenCGN: Great! That worked! Did the same for `allowUrlInput`, tho.

app/templates/components/gh-image-uploader.hbs
AileenCGN: I disabled the blog icon upload form in general settings now, if `fileStorage = false`.

app/components/gh-image-uploader.js
```diff
@@ -177,7 +192,9 @@ export default Component.extend({
     generateRequest() {
         let ajax = this.get('ajax');
         let formData = this.get('formData');
-        let url = `${ghostPaths().apiRoot}/uploads/`;
+        let isIcon = this.get('isIcon');
+        // CASE: we want to upload an icon and we have to POST it to a different endpoint, expecially for icons
+        let url = isIcon ? `${ghostPaths().apiRoot}/uploads/icons/` : `${ghostPaths().apiRoot}/uploads/`;
```
AileenCGN: After going through the existing code in core and having a conversation with @kirrg001, we decided to post icons to a new endpoint. Therefore, I created the property `isIcon` which checks in the `_defaultValidator` if the extensions with which `gh-image-uploader` is called include the `ico` extension. In this case, we have an icon for sure and I set the correlated property to `true`.

AileenCGN:
This one is ready for merging 🎉

app/components/gh-image-uploader.js
kevinansfield: Rather than build domain knowledge into a generic component it would be better to add an `uploadURL` attribute so that it can be re-used anywhere in the app, keeping the domain knowledge in the higher level components/controllers. If you follow the same `didReceiveAttrs` pattern as the accept/extensions attributes that would be 👍

PR #411, opened 2016-11-23
Better active link handling on content page

acburdine:
Whilst doing testing with Glimmer 2, I was getting an infinite render issue, which was being caused by the `active-link-wrapper` mixin - I updated it to depend on the `ember-cli-active-link-wrapper` because it has a mixin.

Also, I moved the template content for the post item into the component itself - it seems like better practice then just doing `{{yield this}}` in the template.

~~With the release of Ember 2.10, this PR also introduces Glimmer2 support :tada: I did a cursory overview of all of the app's functionality and it all seemed to work for the most part - there were a couple of deprecations thrown in the editor but that's probably an editor issue and will debug it further later.~~ will be done in a future PR


acburdine:
@kevinansfield Initially this PR was going to add glimmer2 support, but as that is dependent on the validations refactor it's probably best to just merge this as-is and then worry about glimmer2 support later.

PR #412, opened 2016-11-23
[do not merge] slack app => ember engine as proof of concept

acburdine:
This PR is a proof of concept for implementing lazy-loading engines into Ghost.
It converts the apps/slack portion of the app to a lazily-loaded engine as part of an in-repo addon.

Some notes:
- it only works with Glimmer2 (hence ember 2.10.0-beta)
- it only works with the current beta of ember-engines
- depends on ghost upstream changes (todo: look into this more and see if they can't be combined better)
- it has to do a bunch of importing and re-exporting of current Ghost components - this probably needs to be re-evaluated at some point

todo:
- rework the asset-manifest config so that Ghost upstream changes aren't needed

also todo:
- consider doing this as a route-less engine since there's only one route?


acburdine:
Closing this for now to cleanup the PR noise, but can reopen later if there's interest / once glimmer2 support gets merged

PR #421, opened 2016-11-28
Improve test timing with ember-exam

acburdine:
ember-exam now supports mocha, so this updates Travis testing to use ember-exam.

TODO: play around with parallel/split numbers to maximize efficiency


acburdine:
This didn't seem to actually work, so going to close it for now, will reopen if there's actually a need for it in the future.

PR #454, opened 2016-12-19
🐛 fix session restoration on setup/three with temp token

kevinansfield:
no issue
- fixes an issue where landing on setup/three with a temporary token would appear to work until you clicked the invite or skip buttons at which point you landed back on the signin screen
- ESA requires data to be persisted to the store (LocalStorage/cookie) before a session can be "restored"


kirrg001:
works 👍

PR #469, opened 2017-01-04
add extra escaping when using htmlSafe on user input

kevinansfield:
no issue
- ensure that we always pre-escape user input when it's used within `htmlSafe` marked output

I tried to exploit areas we use `htmlSafe` pretty heavily and was unable to (Handlebars/HTMLbars does a good job of preventing this sort of thing) so this PR isn't strictly necessary but I think it's good practice and makes sure that we are demonstrating the best use of `htmlSafe` throughout the codebase.


acburdine:
LGTM 👍

PR #478, opened 2017-01-05
fix tests on LTS

kevinansfield:
no issue
- `liquid-tether` which we use for modals leaves container elements behind when it's used within acceptance tests which then interfere with the integration tests because integration tests expect the first element in the testing container to be the rendered component


kevinansfield:
💥 tests are ✅ again 😄

acburdine:
Awesome, thanks @kevinansfield!

PR #486, opened 2017-01-09
Add code coverage badge to README

acburdine:


kevinansfield:
I'll merge this for the moment but I'm wondering if we should align with the Ghost repo and use codeclimate for coverage reports rather coveralls?

PR #487, opened 2017-01-11
🐛 correctly count multibyte chars in character counters

kevinansfield:
closes https://github.com/TryGhost/Ghost/issues/7739
- use es6's `Array.from` to convert the string to an array to get a symbol count rather than a byte count


kevinansfield:
@kirrg001 this should cherry-pick cleanly for the `lts` branch

kirrg001:
> this should cherry-pick cleanly for the lts branch, let me know if you'd rather I open a separate PR

chery-picking sounds good to me 👍

PR #491, opened 2017-01-13
display a spinner whilst the app is loading

kevinansfield:
no issue
- use the same loading indicator as used on my.ghost.org
- will display whilst the JS is loading and initializing rather than showing a completely blank screen

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -37,6 +37,13 @@
     {{content-for "head-footer"}}
 </head>
 <body>
+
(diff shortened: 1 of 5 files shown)


kevinansfield:
Strange, looks like the files got missed in my commit 😕

kevinansfield:
@acburdine added the missing files 👍

acburdine:
LGTM 👍

PR #494, opened 2017-01-16
✨add yarn.lock file and update travis

acburdine:
no issue
- add yarn.lock file with latest dependencies
- update travis to install via yarn


kevinansfield:
I think the main issue with introducing Yarn at the moment is that `ember install` will still use `npm` so there's some potential problems when adding new dependencies especially when older versions of `npm` are used that have a deeper nesting structure.

I'm fine to merge and see how it goes though 👍

PR #495, opened 2017-01-16
🏃 improve navigation responsiveness with loading substates

kevinansfield:
no issue
- add loading substates for all routes that previously blocked transitions until their model had finished loading
  - enables immediate response on navigation click
  - loading templates include the title bar to provide immediate indication of which page is loading
  - loading templates include a new `{{gh-loading-spinner}}` component that will only show the spinner after 200ms to avoid flashing a spinner for users on fast connections

### DIFF
--- a/app/components/gh-loading-spinner.js
+++ b/app/components/gh-loading-spinner.js
@@ -1,0 +1,20 @@
+import Component from 'ember-component';
+import {task, timeout} from 'ember-concurrency';
+
+export default Component.extend({
+    tagName: '',
+
+    showSpinner: false,
+    // ms until the loader is displayed,
+    // prevents unnecessary flash of spinner
+    slowLoadTimeout: 200,
+
+    startSpinnerTimeout: task(function* () {
+        yield timeout(this.get('slowLoadTimeout'));
+        this.set('showSpinner', true);
+    }),
+
+    didInsertElement() {
+        this.get('startSpinnerTimeout').perform();
+    }
+});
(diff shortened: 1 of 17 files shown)


kevinansfield:
Looks like I have some tests to update here, I'll re-visit those in the morning.

app/routes/application.js
```diff
@@ -3,6 +3,7 @@ import {htmlSafe} from 'ember-string';
 import injectService from 'ember-service/inject';
 import run from 'ember-runloop';
 import {isEmberArray} from 'ember-array/utils';
+import $ from 'jquery';
```
acburdine: does this need to be here? It's throwing lint errors

kevinansfield:
The test errors here are actually showing a change in behaviour with how version mismatch errors are handled. Previously:

1. User clicks on a navigation link
2. New screen models are loaded in the background
3. Model loads fail with a VersionMismatch error
4. Alert is shown and user is still on the same page so they can copy unsaved data if needed

With loading substates:

1. User clicks a navigation link
2. Immediate transition to newscreen-loading state
3. Model loads are started in the background
4. Model loads fail with a VersionMismatch error
5. Alert is shown but user is now on a loading screen so any data is gone, the loading substate doesn't push anything into the history so the back button won't take the user to the screen they just navigated from

Thinking about this a bit more I'm not sure this is too much of an issue - it will only occur when a user is navigating in which case any unsaved data is gone anyway. A version mismatch error whilst saving will remain on the same page as no transition is kicked off. VersionMismatch errors should also be an extreme edge case.

I'll update the tests to reflect the new behaviour.

kevinansfield:
@acburdine updated, should be good to go now 😄

acburdine:
LGTM 👍

PR #496, opened 2017-01-19
Start Ember Concurrency refactor

kevinansfield:
refs https://github.com/TryGhost/Ghost/issues/7865

- de-couple gh-task-button from gh-spin-button
- switch all modals over to ember-concurrency and gh-task-button

TODO:
- [x] `gh-task-button` tests


kevinansfield:
@acburdine of course Firefox has a different way of computing inline-styles compared to chrome 😞 Hold off on the review for now, I'll see if I can find a fix

kevinansfield:
@acburdine this should actually be ready to review now 😁

PR #499, opened 2017-01-20
⚡❓ AMP app integration

AileenCGN:
refs TryGhost/Ghost#7769

Removes Google AMP option from general settings and adds it to apps instead.


AileenCGN:
@kevinansfield: No idea how to fix this for Travis here. Local tests for it are totally fine.

PR #500, opened 2017-01-21
[WIP] Validations

acburdine:
Third time's the charm.

This PR is the beginning of the inline validation system refactor. It won't replace all of the validation system at once, but the idea is to add the new components in place and then work on getting the rest of the forms upgraded bit by bit.

TODOs for this:
- [ ] Facebook & Twitter validator cleanup
- [ ] Tests


app/components/form-field.js
```diff
@@ -0,0 +1,30 @@
+import Ember from 'ember';
+import FormField from 'ember-form-for/components/form-field';
+
+const {get, getWithDefault, set} = Ember;
+
+export default FormField.extend({
```
kevinansfield: We've had trouble with this update-on-focus-out workaround with checkboxes in Safari/Firefox. I didn't dig in too deeply but ended up simply overriding the behaviour so that checkboxes always update the underlying value instantly:

```javascript
    didInsertElement() {
        this._super(...arguments);

        // checkboxes don't work with `rawValue` in Safari/Firefox and shouldn't
        // have on-focus-out validation attached to them anyway
        if (get(this, 'control') === 'one-way-checkbox') {
            set(this, 'updateOnFocusOut', false);
        }
    },
```

kevinansfield:
👍 looks good so far, should hopefully clean things up a lot 😁 

One thing I I'm not too sure about is the util for creating a mixin. I'm wondering if it would be simpler/cleaner to use the standard `changeset=(changeset ...)` helper and ensure that we pass the changeset around as an argument to actions/tasks. From what I can see the mixin is being used so that  the `generatePassword` observer can still function but it seems like it may be better to drop the observer and instead have an `update` action on the `isPrivate` checkbox control that generates the password if necessary?

acburdine:
@kevinansfield yeah - that was a bit of a weird approach. The problem I was running into is that by default the changeset object is passed to the submit action from `ember-form-for` when you click the submit button, but if you just press "cmd+s" then the save action is called with no changeset which leads to errors. There is probably a need to refactor the keyboard shortcut handling across the app (at some point) but didn't want to do that with this PR. The util for creating a mixin was intended to avoid the same code in each controller that needs a changeset.

The other approach to it would be what I did in previous refactor attempts, which was to define a changeset object on the model itself - however it's probably better performance-wise to have it be done in the controller/template like I did here.

I'm definitely all for removing the util - just didn't see another way as I was doing it and am completely open to suggestions 😄

app/components/form-field.js
acburdine: 👍

kevinansfield:
> The problem I was running into is that by default the changeset object is passed to the submit action from ember-form-for when you click the submit button, but if you just press "cmd+s" (behavior defined in Ghost itself) then the save action is called with no changeset which leads to errors.

Good point, I'm not sure there's a way around that at the moment without updating the keyboard shortcut to trigger the form's `onsubmit` handler.

acburdine:
> Good point, I'm not sure there's a way around that at the moment without updating the keyboard shortcut to trigger the form's onsubmit handler.

Which I initially tried to do, but it didn't work because the cmd+s shortcut is set to 'all' scopes - making it impossible to override (I tried lots of ways 😕 ). I'd say the best thing for this PR at least is to maybe find a better way than using the util, then work on refactoring that little bit out once we've addressed the keyboard shortcuts issue.

kevinansfield:
Don't spend too much time trying to make things perfect, I'd say keep the util for now, it's the validations side of things which is more important and has more effect on other code. We can circle back and revisit the util and keyboard shortcuts later on.

kevinansfield:
Hey @acburdine 👋 I'm going to close this as it's now quite out of date 😬

I'm of two minds at the moment for how to handle this in the future. On one side we could push forward with moving some validation and maybe trickier bits of "unsaved change" handling to `ember-changeset` now, on the other we could wait for the [Ember Data ModelData](https://github.com/igorT/rfcs/blob/17e8f97049452ae18d3b12ee23c4858a6d5e0bfd/text/0000-model-data.md) work to land which should open up store forking which would have a similar effect to `ember-changeset`.

PR #507, opened 2017-01-26
🐛  fix icon

kirrg001:
refs https://github.com/TryGhost/Ghost/issues/7688

We don't set a default value anymore.

### DIFF
--- a/app/controllers/settings/general.js
+++ b/app/controllers/settings/general.js
@@ -33,7 +33,7 @@ export default Controller.extend(SettingsSaveMixin, {
     }),
 
     iconImageSource: computed('model.icon', function () {
-        return this.get('model.icon') || '';
+        return this.get('model.icon') || '/favicon.ico';
     }),
 
     coverImageSource: computed('model.cover', function () {


app/controllers/settings/general.js
```diff
@@ -33,7 +33,7 @@ export default Controller.extend(SettingsSaveMixin, {
     }),
 
     iconImageSource: computed('model.icon', function () {
-        return this.get('model.icon') || '';
+        return this.get('model.icon') || '/favicon.ico';
```
acburdine: So I haven't tested this - but since the icon location is just directly dumped into the img _src_ attribute, won't this cause issues with Ghost blogs in a subdirectory?

If I'm misunderstanding what the purpose of the default is then let me know 😄

kirrg001: @AileenCGN will take over

AileenCGN: I will test this now, @acburdine. We build redirects `/favicon.ico` and `/favicon.png` requests, which should actually work.

AileenCGN:
This can be closed. The favicon url for admin is rendered with the `{{asset}}` helper in Ghost itself, which I fixed with TryGhost/Ghost#7700

PR #512, opened 2017-01-30
css fix for #7890

andreborud:
#7890
a css fix to keep the button on the right when modal opens


kevinansfield:
I think in this case the re-positioning of the modal within the HTML as shown in https://github.com/TryGhost/Ghost-Admin/pull/513 is preferred. The button jumping is a result of the flexbox justification of the header element, it seems like that expected CSS behaviour should be kept rather than adjusting the view-actions margin to workaround divs being inserted where they shouldn't be.

Closing in favour of https://github.com/TryGhost/Ghost-Admin/pull/513

PR #514, opened 2017-01-30
Minor dependency updates

kevinansfield:
Weekly manual dependency update whilst Greenkeeper is disabled.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -1,11 +1,9 @@
 {
   "name": "ghost-admin",
   "dependencies": {
-    "Faker": "3.1.0",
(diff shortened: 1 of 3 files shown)


acburdine:
@kevinansfield something else that might be good to go in here (specifically, the `ember-cli-mocha` update) is the removal of `ember-mocha-adapter` from bower, as it's no longer needed. Also, bower.json has Faker.js in there twice

kevinansfield:
@acburdine good catch, I think that got missed in an earlier upgrade

PR #517, opened 2017-01-31
Upgrade to Ember 2.11

kevinansfield:
no issue
- upgrade to Ember.211
- upgrade `ember-light-table` for Ember 2.11 compatibility
- fix double-render errors (they were only a deprecation before Glimmer2)
- fix broken tests that were highlighted by the upgrade
- add filter menu to the posts-loading template to avoid flickering when changing filters

TODO:

- [x] fix flashing filter menu when switching filters in new content screen
- [x] fix issue with navigation occasionally showing author-level menu items instead of administrator-level menu
- [x] test tag re-ordering via selectize (determine if it was only the test that was broken)
- [x] investigate performance impact on infinite scroll
  - Safari & Firefox are both faster (they were already crazy fast compared to Chrome)
  - Chrome still exhibits variably slow performance, often showing the spinner for a second or two before rendering the next page of results but also sometimes rendering the next page as quickly as Safari/Firefox
  - Chrome Canary had similar behaviour to Chrome
  - Possibly related to author images? 
- [x] add back assertion for invites textarea being highlighted as an error during setup (was removed as it was failing and initially the code didn't look like it should be highlighting the textarea)


kevinansfield:
Rebase and merge would be good, I think being able to track which changes relate to the specific fixes could be useful

PR #518, opened 2017-02-01
use current URI instead of configured blog URI for OAuth redirectURI

kevinansfield:
refs https://github.com/TryGhost/Ghost/issues/7907
- rather than forwarding the redirect URI as configured in Ghost we should use the URL that the admin client is currently loaded on when setting the Ghost OAuth `redirectUri` attribute
- fixes issue with the admin app loading inside of the OAuth popup window and not logging in when using https with a registered http redirect uri

### DIFF
--- a/app/torii-providers/ghost-oauth2.js
+++ b/app/torii-providers/ghost-oauth2.js
@@ -9,7 +9,7 @@ let GhostOauth2 = Oauth2.extend({
 
     name:    'ghost-oauth2',
     baseUrl: computed(function () {
-        return `${this.get('config.ghostAuthUrl')}/oauth2/authorize`;
+        return `${this.get('config.ghostAuthUrl')}/oauth2/authorize/`;
     }),
     apiKey: computed(function () {
         return this.get('config.ghostAuthId');
--- a/app/torii-providers/ghost-oauth2.js
+++ b/app/torii-providers/ghost-oauth2.js
@@ -1,6 +1,7 @@
 import Oauth2 from 'torii/providers/oauth2-code';
 import injectService from 'ember-service/inject';
 import computed from 'ember-computed';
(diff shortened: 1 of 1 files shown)


app/torii-providers/ghost-oauth2.js
```diff
@@ -21,7 +23,16 @@ let GhostOauth2 = Oauth2.extend({
     // we want to redirect to the ghost admin app by default
     init() {
         this._super(...arguments);
-        this.set('redirectUri', `${this.get('config.blogUrl')}/ghost/`);
```
kirrg001: Ghost would still serve `blogUrl` via public configuration endpoint - remove/keep?

app/torii-providers/ghost-oauth2.js
```diff
@@ -21,7 +23,16 @@ let GhostOauth2 = Oauth2.extend({
...
+        let adminPath = ghostPaths().adminRoot;
+        let redirectUri = `${window.location.protocol}//${window.location.host}`;
```
kirrg001:
```
window.location.host
"blog.ghost.local:2368"

window.location.port
"2368"
```

app/torii-providers/ghost-oauth2.js
kevinansfield: Keep, it's used elsewhere in the client

app/torii-providers/ghost-oauth2.js
kevinansfield: Good catch, I was testing without a port so didn't see that! I'll fix

app/torii-providers/ghost-oauth2.js
kirrg001: 👍

PR #520, opened 2017-02-06
Dependency updates

kevinansfield:
Another weekly rollup, also bumps our sub-dependencies in yarn.lock as that's not done automatically when individual packages are upgraded.


acburdine:
Note: After this PR is merged #44 should be able to be rebased and merged

PR #523, opened 2017-02-09
Offline Support via ServiceWorker

felixrieseberg:
I'm so very excited to be sharing this with you - I spend a bit of time adding offline support to Ghost Admin. This is, of course, done with the super powers of Service Workers. To explain this in a few words, when you load Ghost Admin, we register a `ServiceWorker` with the browser. That service worker is able to intercept and cache all requests originating from `https://blog.com/ghost/`. As soon as you load the admin page, I aggressively cache *everything* inside Ghost Admin, but as you keep clicking around, we also cache _all_ API requests. Thus, even while offline, users can load the Ghost Admin interface, check out their existing posts, and even edit them. It's pretty bonkers.

:rocket: Read operations: :100:% coverage
:books: Write operations: :zero:% coverage, but easy to implement.

The heavy lifting happens in modules, in detail:
* ember-service-worker (SW infrastructure)
* ember-service-worker-ghost-cache (Caches static assets and the main application)
* ember-service-worker-cache-fallback (Caches API requests and falls back to the cache if the server can’t be reached)
* ember-service-worker-index (Caches main html)


ember-cli-build.js
```diff
@@ -68,6 +68,14 @@ module.exports = function (defaults) {
             optional: ['es6.spec.symbols'],
             includePolyfill: true
         },
+        'ember-service-worker': {
+            rootUrl: '/ghost/'
```
acburdine: Would this break things if Ghost is run in a subdirectory or no?

felixrieseberg: Offline wouldn't work, but it wouldn't break things. We (for now) need to know ahead of compile-time where we're running.

PR #527, opened 2017-02-13
✨ welcome tour

kevinansfield:
refs TryGhost/Ghost#5168
- adds a `tour` service that handles syncing and management of tour throbbers & content
- adds a `gh-tour-item` component that handles the display of a throbber and it's associated popover when clicked
- populates viewed tour items on app boot/signin

TODO:
- [ ] add tests
- [x] review `user.tour` JSON format - do we need more than an array of IDs?
  - for the current tour items an array of IDs is sufficient, if we need deeper customisation for recording tour item state later on then the client should be able to modify the `user.tour` JSON itself
- [ ] add usage instructions
- [x] add content and hook up throbbers
- [x] fix bug with throbber positioning not working with PSM's CSS based positioning/animation
- [x] ensure that all on-screen throbbers are removed when the "skip these tips" button is clicked
- [ ] update styles to match new design, see https://github.com/TryGhost/Ghost/issues/5168#issuecomment-305316885
- [x] add mechanism to allow dynamic content to be inserted into tour content, e.g. `{{blogUrl}}`
  - partially fixed, using semi-hardcoded values for now, we can look at using variable replacement later on if required
- [ ] fix bug with background animation not working too well with the PSM (may be possible to specify a liquid destination stack to avoid the PSM sitting on top of the background whilst it's animating)

### DIFF
--- a/app/components/gh-post-settings-menu.js
+++ b/app/components/gh-post-settings-menu.js
@@ -10,8 +10,11 @@ import moment from 'moment';
 import {guidFor} from 'ember-metal/utils';
 import {htmlSafe} from 'ember-string';
 import {invokeAction} from 'ember-invoke-action';
+import {task, timeout} from 'ember-concurrency';
 
 const {Handlebars} = Ember;
+
+const PSM_ANIMATION_LENGTH = 400;
 
 export default Component.extend(SettingsMenuMixin, {
     selectedAuthor: null,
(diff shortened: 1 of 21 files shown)


kevinansfield:
@kirrg001 this is ready to merge, the final style tweaks can be handled during the beta period and separate issues can be opened for the tests/instructions and the minor background animation bug for the PSM tour items

kevinansfield:
Noticed another bug 😞 Split screen in the editor should probably disable the nav throbber...

screen shot 2017-06-08 at 10 33 47

PR #528, opened 2017-02-14
WIP: New content

JohnONolan:
Can be merged as-is or left open, doesn't really matter.


kevinansfield:
Going to merge as-is so that it's easier to avoid conflicts. I have a quick-fix for the flash when switching filters I can push separately as I don't seem to have permission to push updates to your branch.

PR #540, opened 2017-02-21
New settings/general styles

JohnONolan:
Dem styles


kevinansfield:
Incorporated into https://github.com/TryGhost/Ghost-Admin/pull/541 with test fixes and accordion toggles

PR #551, opened 2017-02-27
Dependency updates

kevinansfield:
Another dep update rollup, also bumps our sub-dependencies in yarn.lock as that's not done automatically when individual packages are upgraded.


kevinansfield:
`ember-sortable@1.9.3` update was removed as it was breaking tests, from an initial quick investigation it seems that it interacts with the `wait()` helper in some way that causes the tests to timeout. /cc @acburdine

acburdine:
Hmm...ok - the only thing that changed was a fix for Chrome touch events.....but I'll look at that and see.

The update isn't really necessary as we don't use touch events anyways...but that's a good catch as I imagine that breaks things in other ember apps that use `ember-sortable` as well.

PR #555, opened 2017-02-28
Synchronous token refresh on app boot (LTS)

kevinansfield:
refs https://github.com/TryGhost/Ghost/pull/8065
- old access tokens are now deleted when we get a new one using the refresh token so we need to ensure that we don't fire multiple requests in parallel during app boot in case the token refresh is handled by the server first therefore causing the other requests with the old credentials to fail


kevinansfield:
Closing as this is no longer needed. Deleting tokens caused a lot of other issues so we're going with a softer approach that also won't break a background token refresh.

PR #559, opened 2017-03-03
🔥 remove URL input option from image upload components

kevinansfield:
refs https://github.com/TryGhost/Ghost/issues/8032
- `fileStorage: false` config is going away, it predates storage engines and will simplify future image optimisation work
- simplifies UI, it can be brought back in the future in a more robust fashion if required

### DIFF
--- a/app/components/gh-image-uploader-with-preview.js
+++ b/app/components/gh-image-uploader-with-preview.js
@@ -9,12 +9,6 @@ export default Component.extend({
             }
         },
 
-        onInput() {
-            if (typeof this.attrs.onInput === 'function') {
-                this.attrs.onInput(...arguments);
-            }
-        },
-
         uploadStarted() {
             if (typeof this.attrs.uploadStarted === 'function') {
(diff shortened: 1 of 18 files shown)


ReidWeb:
@kevinansfield @acburdine 

Could you clarify for me which option this removes:
A) Option to upload an image directly?
B) Option to link to an image on external storage e.g. S3/Cloudfront?

I use the latter a lot to optimize page load times, and personally I don't think it should be removed.

acburdine:
@ReidWeb this removes option b - however with the new editor it should be much easier to add linking to image urls manually, without going through the entire image upload ui.

As is mentioned in the comment above as well, this is a drive to simplify the image upload UI - and depending on some stylistic changes it may be brought back in a different form before Ghost 1.0 is released.

ReidWeb:
Thanks for the explanation @acburdine, this caused me to discover storage adapters! Didn't realise they were a feature now. Trying them out now, way better than manually uploading to S3 all the time!

PR #560, opened 2017-03-03
remove email field when using OAuth, direct users to my.ghost.org

kevinansfield:
no issue
- removes user email field when using Ghost OAuth because email addresses are synced from the central identity management system
- adds a link to my.ghost.org account management when using Ghost OAuth and viewing your own user

Link to my.ghost.org currently looks like this, we probably want to make this more prominent as part of https://github.com/TryGhost/Ghost/issues/8021

screen shot 2017-03-03 at 09 19 54

### DIFF
--- a/app/controllers/team/user.js
+++ b/app/controllers/team/user.js
@@ -43,8 +43,17 @@ export default Controller.extend({
     rolesDropdownIsVisible: and('isNotOwnProfile', 'canAssignRoles', 'isNotOwnersProfile'),
     userActionsAreVisible: or('deleteUserActionIsVisible', 'canMakeOwner'),
 
-    isNotOwnProfile: computed('user.id', 'currentUser.id', function () {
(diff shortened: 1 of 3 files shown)


kirrg001:
LGTM 👍 I wonder if the user is able to see his email address in this blog anywhere?

kevinansfield:
> I wonder if the user is able to see his email address in this blog anywhere?

Hmm, good point. The only place the email is shown anywhere is possibly on the author page in the front-end but that's theme-dependent 🤔

There are two use cases I can think of:
1) knowing your own e-mail address on the site
2) owners/admins/editors looking up a users e-mail address to contact them

I'll push a change that always shows the e-mail field but if using Ghost OAuth it renders the email address as text in the same way it does when admins view an owner account.

kevinansfield:
Updated

PR #563, opened 2017-03-04
🎨  simplify travis test setup

acburdine:
no issue
- travis will install yarn by default, we don't need to install it
- we can also rely upon the local installation of bower


acburdine:
hmm....apparently that didn't work - I think it has to do with the fact that we're using sudo images on the ubuntu trusty. Anyways, closing this for now.

PR #564, opened 2017-03-04
Koenig tests.

disordinary:
Refs: https://github.com/TryGhost/Ghost/issues/7429

This PR adds acceptance tests for the koenig editor into ghost-admin, it also changes a couple of things within koenig for consistency.
- Add acceptance tests for text input and markdown support in koenig.
- Added some component tests (which are located in the `test-support` directory of `lib/koenig/`) (edit, these are skipped as they're failing ATM, will continue to work on them and flesh them out).
- Moved `gh-koenig/libs` to `gh-koenig/lib` for consistency.


lib/gh-koenig/.gitignore
```diff
@@ -0,0 +1,55 @@
+b-cov
```
kevinansfield: Do we need the additional `.gitignore` file? All of these ignores should already be in the root `.gitignore`

tests/test-helper.js
```diff
@@ -5,7 +5,7 @@ setResolver(resolver);
 
 /* jshint ignore:start */
 mocha.setup({
-    timeout: 15000,
+    timeout: 25000,
```
kevinansfield: Were you having trouble with timeouts in general acceptance tests or only the editor test? It's possible to set timeouts on a per-test basis by using `this.timeout(25000)` within the `it()` function.

I'd like to drop the global timeout in the future if possible rather than increase it - in some circumstances a failing test can have a knock-on effect which makes all later tests timeout so Travis can get hung up for a long time if no one notices and cancels the build.

lib/gh-koenig/.gitignore
disordinary: For some reason it commited all of the ember stuff for me and I had to reverse my commit. Maybe it's not required.

tests/test-helper.js
disordinary: Only the editor tests, perhaps we can do it within the describe rather than the it?

kevinansfield: Yep, `this.timeout()` inside the `describe()` should also work 👍

PR #570, opened 2017-03-08
various test deprecation cleanup

acburdine:
no issue
- cleans up a lot of the deprecation noise in tests

### DIFF
--- a/config/deprecation-workflow.js
+++ b/config/deprecation-workflow.js
@@ -1,6 +1,0 @@
-window.deprecationWorkflow = window.deprecationWorkflow || {};
-window.deprecationWorkflow.config = {
-    workflow: [
-        {handler: 'silence', matchMessage: 'You modified (-join-classes "ember-view" "form-group" (-normalize-class "errorClass" errorClass activeClass=undefined inactiveClass=undefined)) twice in a single render. This was unreliable in Ember 1.x and will be removed in Ember 3.0'}
-    ]
-};
--- a/lib/gh-koenig/addon/components/gh-koenig.js
+++ b/lib/gh-koenig/addon/components/gh-koenig.js
@@ -24,7 +24,6 @@ export default Component.extend({
 
     init() {
         this._super(...arguments);
-        this.container = document.querySelector('.gh-editor-container')[0];
 
         let mobiledoc = this.get('value') || BLANK_DOC;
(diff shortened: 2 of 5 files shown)


lib/gh-koenig/addon/components/gh-koenig.js
```diff
@@ -24,7 +24,6 @@ export default Component.extend({
 
     init() {
         this._super(...arguments);
-        this.container = document.querySelector('.gh-editor-container')[0];
```
acburdine: I looked and couldn't find any usage of `container` in any of the editor component files (js or hbs), so I assume this was used once and didn't get removed?

config/deprecation-workflow.js
```diff
@@ -1,6 +1,6 @@
 window.deprecationWorkflow = window.deprecationWorkflow || {};
 window.deprecationWorkflow.config = {
     workflow: [
-        {handler: 'silence', matchMessage: 'You modified (-join-classes "ember-view" "form-group" (-normalize-class "errorClass" errorClass activeClass=undefined inactiveClass=undefined)) twice in a single render. This was unreliable in Ember 1.x and will be removed in Ember 3.0'}
+        {handler: 'silence', matchId: 'ember-runtime.action-handler-_actions'}
```
kevinansfield: I think this is raised by `ember-invoke-action`, it might be worth looking at removing usage of that addon? It doesn't seem to be maintained any longer as far as I can see.

acburdine: Ah, that would make more sense, I'll add that to this PR.

acburdine: Actually now that you mention it, I know what's causing these deprecations. In the file uploader we're using `invoke` instead of just `send`. That's what's causing the deprecations.

acburdine:
@kevinansfield I think if we want to remove `ember-invoke-action` we should do that as a separate PR, as it's used in a large part of the codebase. The specific deprecations that were being thrown were due to a specific usage of `ember-invoke-action` that needed to be removed.

Also, due to the fact that there's not a huge use for it at the moment, I dropped the ember-cli-deprecation-workflow dependency. I feel like it may have more use added locally if it's needed temporarily.

kevinansfield:
👍 agree on both points, everything LGTM, will merge once Travis is green

PR #574, opened 2017-03-09
🎨 remove usage of ghost's {{asset}} helper in index.html

kevinansfield:
requires https://github.com/TryGhost/Ghost/pull/8142
- switch to hash-location rather than history-location
- remove usage of Ghost's `{{asset}}` helper in index.html
- add `content-for` helpers to `asset-delivery` addon that switch asset urls in index.html to `.min` files in production
- update the `asset-delivery` addon to copy the production `index.min.html` to `default-prod.hbs` so Ghost can serve production assets when in production mode
- enable asset fingerprinting
- add `ember-cli-inject-live-reload` dependency
- remove `ember-cli-sri` dependency - we weren't using it but now that ember is handling assets it was used automatically and could potentially create issues if users have proxy servers that attempt to compress or otherwise modify asset files

This cuts the final tie that requires Ghost to render+parse the index.html as a handlebars file which means that `ember serve` will finally be usable and the associated livereload functionality can be re-enabled.

To use ember serve and livereload:
1. In first tab, run `npm start` in Ghost root
2. In second tab, `cd core/client` then run `ember serve --proxy=http://localhost:2368`
3. Admin area is now available on http://localhost:4200 with livereload enabled (necessary to use password auth at the moment due to mismatch between ember URL and registered ghost url in auth server)

TODO:
- [x] server-side component that switches between default.hbs and default-prod.hbs depending on environment
- [x] `favicon.ico` handling - ghost-admin doesn't have a favicon.ico
- [ ] look into how the `core/built` folder can be kept clean - asset files now have hashes in the filename so old files will be kept around rather than being overwritten - add a gulp/grunt task and ensure any release task calls it so releases have a clean built/assets folder?
- [x] remove `minified-in-production` and combine with the `asset-delivery` addon
- [x] check whether `ghost-dark.css` is rewritten correctly to match the hashed asset
- [ ] if using `ember serve --proxy`, how should Ghost OAuth be handled? (redirect URI mismatch)

### DIFF
--- a/ember-cli-build.js
+++ b/ember-cli-build.js
@@ -95,6 +95,7 @@ module.exports = function (defaults) {
         },
         outputPaths: {
             app: {
+                html: isProduction ? 'index.min.html' : 'index.html',
                 js: assetLocation('ghost.js'),
                 css: {
                     app: assetLocation('ghost.css'),
@@ -112,7 +113,7 @@ module.exports = function (defaults) {
                 plugins: postcssPlugins()
             }
         },
-        fingerprint: disabled,
+        fingerprint: {enabled: true},
         nodeAssets: {
             'blueimp-md5': {
                 import: ['js/md5.js']
--- a/app/index.html
+++ b/app/index.html
@@ -19,20 +19,19 @@
     <meta name="apple-mobile-web-app-status-bar-style" content="black" />
     <meta name="apple-mobile-web-app-title" content="Ghost" />
 
-    <link rel="shortcut icon" href="{{asset "favicon.ico"}}" />
-    <link rel="apple-touch-icon-precomposed" href="{{asset "img/touch-icon-iphone.png" ghost="true"}}" />
-    <link rel="apple-touch-icon-precomposed" sizes="76x76" href="{{asset "img/touch-icon-ipad.png" ghost="true"}}" />
-    <link rel="apple-touch-icon-precomposed" sizes="120x120" href="{{asset "img/small.png" ghost="true"}}" />
-    <link rel="apple-touch-icon-precomposed" sizes="152x152" href="{{asset "img/medium.png" ghost="true"}}" />
+    <link rel="shortcut icon" href="../favicon.ico" />
+    <link rel="apple-touch-icon-precomposed" href="assets/img/touch-icon-iphone.png" />
+    <link rel="apple-touch-icon-precomposed" sizes="76x76" href="assets/img/touch-icon-ipad.png" />
+    <link rel="apple-touch-icon-precomposed" sizes="120x120" href="assets/img/small.png" />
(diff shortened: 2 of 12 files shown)


ember-cli-build.js
```diff
@@ -112,7 +113,7 @@ module.exports = function (defaults) {
                 plugins: postcssPlugins()
             }
         },
-        fingerprint: disabled,
+        fingerprint: {enabled: true},
```
acburdine: Isn't fingerprinting enabled by default? (meaning this line would no longer be necessary)

kevinansfield: Fingerprinting is only enabled in production by default, enabling it for development too should hopefully help catch those edge cases where we expect a non-fingerprinted asset

kevinansfield:
@acburdine interesting this actually fixes the night-shift mode in production, the asset fingerprinting knows how to convert `ghost-dark.css` to `ghost-dark{.min}-abcef.css`

ErisDS:
I had a quick look at doing the server side component of this:

- [x] switch between default.hbs & default-prod.hbs
- [ ] updating any direct links to the admin to `/#/whatever`
- [ ] adding a redirect for `/ghost/whatever` to `/ghost/#/whatever` 

I've run into a problem though, with the logic that does a redirect to `/ghost/setup/`.

If Ghost is not setup, then we have this middleware: https://github.com/TryGhost/Ghost/blob/master/core/server/middleware/redirect-to-setup.js

It checks if this is request where:

1. the blog is not a setup
2. this is not a request for ghost/setup
3. we're not doing oauth authorization

The problem is we can no longer detect if rule 2 is true or false on the server side, because only the client will know if we are on #/setup or some other URL. 

Therefore, somehow this logic needs to be moved the admin panel?

kevinansfield:
@ErisDS I've added another commit to this PR to [handle the setup redirect](https://github.com/TryGhost/Ghost-Admin/pull/574/commits/4703f9be031e4111e659758d6924e4556fb35c99), I think that means we can remove the `redirect-to-setup` middleware from Ghost itself?

ErisDS:
Cool, will check that out asap.

Can we also move the location of the templates from server/views to server/admin/views as part of this PR? I'll do the same in my server side branch. 

Sorry, I thought I already added a note here about that but must have not hit send :(

kevinansfield:
No problem on the templates, looking at the test failure atm, will push the templates change with the fix shortly.

kevinansfield:
Ugh, 2 hours debugging a really random issue with the setup redirect but should all 👍 now, template output dir has been updated too.

ErisDS:
Took this for a spin and it all seems to be working 🎉  I need to fix the server-side tests, but otherwise my PR TryGhost/Ghost#8142 is up-to-date and working well with this one.

PR #576, opened 2017-03-13
✨ show theme warnings/errors when activating a theme

kevinansfield:
closes https://github.com/TryGhost/Ghost/issues/8127
- update theme activation to manually set other themes to `active: false` in the store now that we only get the active theme back from `/themes/:name/activate` endpoint
- move theme warning list item rendering into `{{gh-theme-error-li error=x}}`
- add `theme-warnings` modal that accepts a warnings list, title, and optional message
- after activating a theme, check if the theme has any warnings or errors and display an appropriate modal informing the user

theme-activation-warnings

### DIFF
--- a/app/templates/components/gh-theme-error-li.hbs
+++ b/app/templates/components/gh-theme-error-li.hbs
@@ -1,0 +1,13 @@
+<li>
+    {{#if error.details}}
+        {{{error.details}}}
+    {{else}}
+        {{{error.rule}}}
+    {{/if}}
+
+    <ul>
+        {{#each error.failures as |failure|}}
+            <li><code>{{failure.ref}}</code>{{#if failure.message}}: {{failure.message}}{{/if}}</li>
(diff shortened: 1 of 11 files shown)


app/templates/components/gh-theme-error-li.hbs
```diff
@@ -0,0 +1,13 @@
+<li>
```
acburdine: Not to be nitpicky -> but it seems like if we set `tagName = 'li'` in the component it would remove the need for this wrapping `li` element?

kevinansfield: True but for purely presentational components like this I prefer to keep all of the markup inside the template, that way there's no need to look in both .js and .hbs files if looking to change things

acburdine: Makes sense 👍

PR #582, opened 2017-03-14
🔥  remove bg shell from Gruntfile

kirrg001:
WIP!

refs TryGhost/Ghost#7427

- both Ghost and Ghost-Admin ran the watch command in background, that didn't work
- the caller can decide if running the watch command in bg or not
- furthermore, the caller can decide if printing stdout/stderr of the background process

see https://github.com/TryGhost/Ghost/pull/8158

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -24,15 +24,6 @@ module.exports = function(grunt) {
             }
         },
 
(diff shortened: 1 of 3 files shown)


kirrg001:
Rebased 👍

PR #584, opened 2017-03-15
🐜  ⬛️   Card menu fixes:

disordinary:
Closes https://github.com/TryGhost/Ghost/issues/8164

- When a user clicks on a card within a '/' menu it now executes that tool.
- If a user engages a card with the / menu the content of the paragraph is now inserted into the card.


kevinansfield:
Whilst testing this I still can't seem to click on the the / menu items, using the keyboard works and I can see the paragraph content inserted into the card. Clicking is failing for me in Chrome+FF

disordinary:
Should be good to go now.

PR #587, opened 2017-03-15
Fix parentheses in image alt text

nglgzz:
Fixes TryGhost/Ghost#8006


kevinansfield:
I've cherry-picked this into a new PR that is based on the `lts` rather than `master` branch.

PR #588, opened 2017-03-15
✨ new publish menu

kevinansfield:
closes https://github.com/TryGhost/Ghost/issues/8249
- replaces the old split-button publish/schedule/update button with a less confusing menu system
- adds a `{{gh-date-time-picker}}` component that contains a datepicker with separate time input
- replaces the date text input in the post settings menu with `{{gh-date-time-picker}}`
  - disabled when post is scheduled, only way to update a scheduled time is via the publish menu
  - validates date is in the past when draft/published so there's no confusion with scheduling
- displays saving status in top-left of editor screen
- refactor editor (auto)saving processes to use ember-concurrency

Other minor changes:
- adds `post.publishedAtBlog{TZ,Date,Time}` properties to Post model to allow working with `publishedAt` datetime in the selected blog timezone rather than UTC
- adds a `beforeSave` hook to `validation-engine` that is called after successful validation and before the Ember Data save call is made
- adds validation of `publishedAtBlog{Date,Time}` to post validator
- prevent gh-task-button showing last task state on first render
- fixes bug where clicking into and out of the published date input in the PSM without making any changes saves a published date for draft posts

TODO:
- [x] refactor post saving to use ember-concurrency for save-state-buttons (see 
https://github.com/TryGhost/Ghost/issues/7865)
- [x] draft pane
  - [x] "radio" selects
  - [x] time input
  - [x] hook up date picker
  - [x] hook up save button
    - [x] set post publish date on save when scheduling
  - [x] scheduled date must be at least 2 mins in the future
- [x] scheduled pane
- [x] published pane
- [x] move date/time picker inputs + logic into a re-usable component
- [x] saving must trigger date validation
- [x] update PSM date input to use same date/time selector as publish pane (but only allow past dates)
  - [x] disable when post is scheduled?
- [x] display "Saving..." for a set period in top left of editor screen each time the post is saved
- [x] update tests
- [x] remove old `gh-editor-save-button` component 
- [x] BUG: flash when editor screen loads
- [x] BUG: <kbd>Cmd+S</kbd> shows error if scheduled time is less than 2 mins away
- [x] BUG: schedule countdown doesn't disappear after scheduled time passes
- [x] BUG: button does not switch to "published" state after scheduled time passes
- [x] BUG: time across DST changes is not observed correctly
  - when blog timezone is not observing DST but you specify a future date where DST is observed, hitting the Publish button will show the "scheduled" pane with a date 1 hour later than you selected
  - every save via the "scheduled" menu will add another hour to the publish time


kevinansfield:
🎉 Ready for review/merge

PR #595, opened 2017-03-20
Multi line title.

disordinary:
The title is now a contenteditable div which stretches and wraps to behave like the editor.

It also tries to seamlessly move the cursor between the editor and title to make one coherent editing experience.

Refs: https://github.com/TryGhost/Ghost/issues/7754

### DIFF
--- a/app/components/gh-editor-title.js
+++ b/app/components/gh-editor-title.js
@@ -73,12 +73,12 @@ export default Component.extend({
                 title.addClass('no-content');
             }
 
-            // sanity check if there is formatting reset it.
             let {textContent} = title[0]; // eslint-disable-line
-            if (title[0].innerHTML !== textContent && title[0].innerHTML) {
-                title[0].innerHTML = textContent;
-                // todo: retain the range position.
-            }
+            // // sanity check if there is formatting reset it.
+            // if (title[0].innerHTML !== textContent && title[0].innerHTML) {
+            //     title[0].innerHTML = textContent;
+            //     // todo: retain the range position.
+            // }
 
             if (this.get('val') !== textContent) {
                 this.set('val', textContent);
--- a/app/components/gh-editor-title.js
+++ b/app/components/gh-editor-title.js
@@ -9,7 +9,7 @@ export default Component.extend({
             return;
         }
 
-        let title = this.$('.gh-title');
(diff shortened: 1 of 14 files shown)


kevinansfield:
I can merge as it'll be good to get more people using it but I think there might still be some tweaks needed?

- At least in Chrome the up/down cursor positioning jumps doesn't seem to work but it is possible to use up/down to move between the two elements
- `{{gh-title}}` should probably be an addon `{{kg-title}}` component
- I'm not sure we need to be using `{{gh-view-title}}` for the editor title - it means we now have nested `<h2>` elements and I believe the old mobile menu button that it includes should be going away now that we have the buttons at the bottom of the screen on mobile

disordinary:
There is an issue around the up and down positioning, Up doesn't work because for some reason the content editable no longer responds to text input in Firefox after the range is manually created in it. We decided simply to merge it and spend some time on it when more important tasks were finished. There is an issue relating to it here: https://github.com/TryGhost/Ghost/issues/8191

The tagname on {{gh-title}} should definitely be a div, I'll fix that, re-name the component, and re-commit.

kevinansfield:
@disordinary I'm hitting infinite loops when editing the title 😕 Chrome becomes unresponsive pretty quickly but what I could see was pointing to the [MutationObserver](https://github.com/TryGhost/Ghost-Admin/pull/595/files#diff-5a33961d0f7d93199e80e7817522d03fR68)

disordinary:
I'm not seeing them at all, what are you doing trigger it?

disordinary:
Hmmm, maybe extensions like the gramarly extension could do it?  It creates HTML mutations that the editor will try and remove because it only wants a plain text title.

Do you have any extensions that might cause mutations?

I'll fix that possibility.

kevinansfield:
No extensions that I know of but I've just tried again without your latest commit and I'm not able to reproduce, no idea what Chrome was doing earlier but even through multiple restarts it was freezing up 😕 

Is there a chance that the latest commit has broken the title event handling in Chrome or is the Travis failure a bit of a random error?

disordinary:
Visually it's working, I ran the tests again and it failed somewhere else. I'm running them a third time so we'll see.

PR #605, opened 2017-03-28
No more icon font: `{{inline-svg}}` - Part 1

AileenCGN:
refs TryGhost/Ghost#8107

Replaces all icons in Ghost-Admin with SVGs by using our new helper `{{inline-svg}}`.

This is the first batch of the refactor. I will split it up to keep it at least a bit clear.


ErisDS:
Is there a reason that this is marked WIP? The PR description doesn't say.

AileenCGN:
It's just WIP, as there are still some more icons that need to be converted.

PR #611, opened 2017-04-04
[WIP] 🎁 Added wordcount to editor.

disordinary:
Closes: https://github.com/TryGhost/Ghost/issues/8202

Adds the `{{wordCount}}` variable to the editor screen and positions it in the bottom left hand side of the screen.

Also adds a `word-pluralize` helper that does basic pluralisation of words.

### DIFF
--- a/app/helpers/word-pluralize.js
+++ b/app/helpers/word-pluralize.js
@@ -1,0 +1,12 @@
+import {helper} from 'ember-helper';
+
+export function wordPluralize(params) {
+    let [count, word] = params;
+
+    if (count > 1) {
+        word += 's';
+    }
+    return word;
+}
+
+export default helper(wordPluralize);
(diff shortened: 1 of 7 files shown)


app/helpers/word-pluralize.js
```diff
@@ -0,0 +1,12 @@
+import {helper} from 'ember-helper';
+
+export function wordPluralize(params) {
```
kevinansfield: Ember Data brings [Ember Inflector](https://emberjs.com/api/data/classes/Ember.Inflector.html) in as a dependency so we can use proper inflection to pluralize words, it also extends the Ember.String prototype so you can do something like...

```
import Ember from 'ember';
const {String: {pluralize}} = Ember;

pluralize(word);
```

It also gives us a [`{{pluralize}}`](https://emberjs.com/api/data/classes/DS.Ember.HTMLBars.helpers.html#method_pluralize) helper that is used elsewhere in the app.

disordinary: Oh nice! I spent a bit of time googling because I thought that would be the case but my google foo failed me in this case.

disordinary:
Closed in favour of: https://github.com/TryGhost/Ghost-Admin/pull/668

PR #626, opened 2017-04-06
🐝  Refresh publication icon after upload

AileenCGN:
refs TryGhost/Ghost#7688

Instead of using a hardcoded `favicon.ico` URL to request the blog/publication icon, we read the settings which get updated as soon as a icon is uploaded.

### DIFF
--- a/app/components/gh-nav-menu.js
+++ b/app/components/gh-nav-menu.js
@@ -11,13 +11,15 @@ export default Component.extend({
 
     open: false,
 
-    navMenuIcon: computed('config.blogUrl', function () {
-        let url = `${this.get('config.blogUrl')}/favicon.png`;
+    navMenuIcon: computed('config.blogUrl', 'settings.icon', function () {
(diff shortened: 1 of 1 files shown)


app/components/gh-nav-menu.js
```diff
@@ -11,13 +11,15 @@ export default Component.extend({
 
     open: false,
 
-    navMenuIcon: computed('config.blogUrl', function () {
-        let url = `${this.get('config.blogUrl')}/favicon.png`;
+    navMenuIcon: computed('config.blogUrl', 'settings.icon', function () {
+        let blogIcon = this.get('settings.icon') ? this.get('settings.icon') : 'favicon.ico';
```
kevinansfield: is `settings.icon` always null/undefined when there's no icon or does is it get set to a blank string? It might be worth wrapping it in an `isBlank()` anyway

kevinansfield: Ignore me, blank strings are falsey in JS. Clearly not had enough caffeine this afternoon 😬

PR #644, opened 2017-04-15
💄 refactor setup screens to use ember-concurrency 

kevinansfield:
refs TryGhost/Ghost#7865
- moves authentication actions from `setup/two` route to controller
- refactors authentication and setup logic into EC tasks
- replaces use of `gh-spin-button` with `gh-task-button`
- fixes some styling issues with the new SVG icons
- adds `app/styles/patterns/icons.css` back to contain per-icon overrides and animations (some SVGs use fills and others use strokes so we sometimes have conflicting styles)

TODO:
- [x] convert setup/three to EC and `gh-task-button`


kevinansfield:
Ready for review 😄

PR #652, opened 2017-04-18
match 0.11.x's auto-save behaviour

kevinansfield:
refs https://github.com/TryGhost/Ghost/issues/8331
- rather than always saving every 60 seconds, the 60 seconds timer is a fallback for when the user has made changes to the post content for 60 seconds without triggering the 3 second debounced timer
- don't perform a save if there have been no changes to the post - this should prevent unnecessary collision errors when multiple people are viewing a post
- adds guards to `gh-publishmenu` so that we don't try to call actions that weren't passed in


kirrg001:
I would like to merge the post update collision into Ghost master first, then i will test this PR 🙃

PR #655, opened 2017-04-19
🎨  Don't scroll to top after toggle checkbox

AileenCGN:
closes TryGhost/Ghost#8335

Use `display: none` instead of positioning it out of the viewport.


AileenCGN:
This is ready for merging. Tested in Firefox, Chrome and Safari.

PR #657, opened 2017-04-19
🐛 clear date error when closing the PSM 

kevinansfield:
closes TryGhost/Ghost#8359
- if the date/time picker is in an error state when the PSM is closed, reset it to the currently saved time
- resolves the problem with the flash of an error as the publish menu is closed as it's no longer possible to be in a hidden error state

### DIFF
--- a/app/components/gh-post-settings-menu.js
+++ b/app/components/gh-post-settings-menu.js
@@ -24,9 +24,12 @@ export default Component.extend(SettingsMenuMixin, {
     session: injectService(),
     settings: injectService(),
 
+    model: null,
     slugValue: boundOneWay('model.slug'),
     metaTitleScratch: alias('model.metaTitleScratch'),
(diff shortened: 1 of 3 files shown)


kirrg001:
Forgot to say: No errors in the debug console.

kevinansfield:
@kirrg001 are there any particular circumstances in which you aren't able to open the PSM or is it everywhere? New/old posts, only certain posts? I'm not able to replicate to investigate unfortunately

kirrg001:
Will try again and see if i can share steps to reproduce.

kirrg001:
I just got this behaviour on my branch #661 as well. So it's not related to this branch.

kevinansfield:
@kirrg001 are you able to test with this PR merged on top? https://github.com/TryGhost/Ghost-Admin/pull/662

I think that is likely to be the problem rather than anything in this PR.

PR #658, opened 2017-04-19
🐛 fix broken sidebar after successful import

kevinansfield:
closes https://github.com/TryGhost/Ghost/issues/8307
- unloading the store and refreshing the `session.user` attribute after an import was triggering a rendering edge case where the style was re-computed and a re-render was attempted after the sidebar has been destroyed
- rather than binding a style attribute directly to a CP in `gh-nav-menu` we pass the menu icon in (using `settings.settledIcon` - see below) and manually set the style attribute via the `didReceiveAttrs` hook so that outside changes don't trigger re-computations when we don't expect them and so we can still react to icons being uploaded or removed
- our usage of `settings.icon` is a bit of an odd situation because it's a link to an external resource that will only resolve correctly after a successful save - if we change `settings.icon` in the local store and the nav menu icon style updates before the save has been completed then the server will give us the old icon. To work around this a `settings.settledIcon` attribute has been added that is only updated when we receive data from the store ensuring that our cache-busting technique works correctly

### DIFF
--- a/app/components/gh-nav-menu.js
+++ b/app/components/gh-nav-menu.js
@@ -1,35 +1,29 @@
 import Component from 'ember-component';
-import {htmlSafe} from 'ember-string';
 import injectService from 'ember-service/inject';
-import computed from 'ember-computed';
+import {htmlSafe} from 'ember-string';
 import calculatePosition from 'ember-basic-dropdown/utils/calculate-position';
 
 export default Component.extend({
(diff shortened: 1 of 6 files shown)


kirrg001:
> Yeah that's normal unfortunately, the import removes all local data before reloading so some parts can't be rendered in-between.

PR #660, opened 2017-04-20
Places no-content placeholder in the correct position in firefox.

disordinary:
Closes: https://github.com/TryGhost/Ghost/issues/8322

Previously in firefox the `no-content` psuedo element was placed below the actual on first load (but not subsequent reloads), this update positions it in the correct place in both scenarios.

posttitle


disordinary:
Wow that gif came out odd. It works in reality 😄

PR #661, opened 2017-04-20
🔥✨ Adapt image property changes

kirrg001:
refs https://github.com/TryGhost/Ghost/issues/8348

- [x] test signup
- [x] test all image cases again


kevinansfield:
@kirrg001 there shouldn't be any underscored names in the client, all properties/attributes/methods are camelCase so for example `feature_image` should be `featureImage`

kirrg001:
@kevinansfield 👍 
[refs https://github.com/TryGhost/Ghost-Admin/blob/master/app/serializers/application.js#L20-L22]

kirrg001:
WIP again, testing something with image deletion.

kirrg001:
Ok this is ready to review now.

app/models/setting.js
```diff
@@ -9,7 +9,7 @@ export default Model.extend(ValidationEngine, {
     title: attr('string'),
     description: attr('string'),
     logo: attr('string'),
-    cover: attr('string'),
+    cover_image: attr('string'),
```
kevinansfield: Is there a reason the API is inconsistent here? There's an odd mix of camelCase vs underscore_case...

```
ghost_head
ghost_foot
cover_image
```


```
defaultLang
activeTimezone
forceI18n
activeTheme
isPrivate
```

kirrg001: Good question. There is no reason, it's just inconsistent. Server uses camel case for our model schema and settings is a model. We should actually make this consistent before releasing the beta.

kevinansfield: 👍 shall we tackle that in a separate issue?

kirrg001: I will try to make it consistent now, if i can't make it today i will raise an issue for later.

kirrg001:
@kevinansfield Thanks!

I have pushed it as separate commit https://github.com/TryGhost/Ghost-Admin/pull/661/commits/d588f21aef5e6cc55aee42c911695527b0dbfdea

PR #663, opened 2017-04-21
💄 refactor acceptance tests to use async/await

kevinansfield:
no issue
- replaces `andThen(() => {}` blocks with `async/await` in acceptance tests

https://www.youtube.com/watch?v=oqwzuiSy9y0


disordinary:
Looks good to me, one thing that I've noticed is that Babel balks sometimes randomly when `await` is used with an invalid character error.

PR #665, opened 2017-04-24
🐝 Strip HTML from title.

disordinary:
Closes: https://github.com/TryGhost/Ghost/issues/8353

Currently when you paste HTML into the title (or modify the title in some other way that inserts HTML) the title keeps the styling which is not saved.

This update detects if HTML is included into the title and strips it if it exists. 

pastehtmlintotitle

### DIFF
--- a/tests/helpers/editor-helpers.js
+++ b/tests/helpers/editor-helpers.js
@@ -29,6 +29,29 @@ export function titleRendered() {
     });
 }
 
+// replaces the title text content with HTML and returns once the HTML has been placed.
+// takes into account converting to plaintext.
+export function replaceTitleHTML(HTML) {
+        return Ember.Test.promise(function (resolve, reject) { // eslint-disable-line
+            let startTimestamp = Date.now();
+            let title = $('#gh-editor-title div');
+            title.html(HTML);
+            function checkTitle() {
+                if (!title[0].innerHTML.includes('<')) {
+                    return resolve();
+                } else {
+                    // timeout after 200ms
+                    if (Date.now() + 200 < startTimestamp) {
+                        return reject();
+                    } else {
+                        window.requestAnimationFrame(checkTitle);
+                    }
+                }
+            }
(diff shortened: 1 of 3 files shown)


tests/helpers/editor-helpers.js
```diff
@@ -29,6 +29,29 @@ export function titleRendered() {
     });
 }
 
+// replaces the title text content with HTML and returns once the HTML has been placed.
+// takes into account converting to plaintext.
+export function replaceTitleHTML(HTML) {
```
kevinansfield: Rather than re-implenting loops within promises we can rely on Ember's built-in helpers, if we couple those with the `ember-native-dom-helpers` library that we're recommending in integration tests we can simplify this helper quite a bit:

```javascript
import wait from 'ember-test-helpers/wait';
import {findWithAssert, triggerEvent, focus} from 'ember-native-dom-helpers';

// replaces the title text content with HTML and returns once the HTML has been placed.
// takes into account converting to plaintext.
export function replaceTitleHTML(HTML) {
    let el = findWithAssert('#gh-editor-title div');

    run(() => focus(el));
    run(() => el.innerHTML = HTML);
    run(() => triggerEvent(el, 'input'));
    run(() => triggerEvent(el, 'change'));
    return (window.wait || wait)();
}
```

kevinansfield: We could also generalise this a bit so it can be used with other elements, e.g. `fillInHTML('#gh-editor-title div', '<div>pasted HTML</div>');

disordinary: Thanks @kevinansfield, using this technique we can simplify it even further.

disordinary: Done, I'm working on a PR which fixes the issues that testing has with some newly introduced side-effects and then to re-enable the skipped tests and expand the test coverage.

I'll include these lessons / make a more generalised helper in that.

PR #668, opened 2017-04-24
🔢  Add wordcount feature.

disordinary:
Closes: https://github.com/TryGhost/Ghost/issues/8202

Implements the wordcount feature, on update it iterates through every section of the document and does a count. The exception is with cards where the count is cached in the card payload and is refreshed upon save.

wordcount

Note: _The count is only displayed if the screen is big enough. Markdown and HTML cards are now counted but the wordcount only updates when the card closes (although a 'word' is something which is pretty abstract in HTML), the HR and image cards are not counted for obvious reasons. A cache of the wordcount is saved with the card in the document, cards are responsible for counting themselves._


disordinary:
I've removed the throttling to get this into the beta, the throttling relies on a bit of a refactor so that it doesn't break tests.

Besides, autosave on change slows us down earlier than wordcount does. I'm not seeing any issues below 70,000 words that will of course differ based on the power of the machine.

If this is an issue, and people are writing novels in Ghost then updating both the autosave and the wordcount on a timer whenever the content has changed is probably the easiest solution.

app/templates/editor/edit.hbs
```diff
@@ -50,9 +50,11 @@
                 setEditor=(action "setEditor")
                 menuIsOpen=(action "editorMenuIsOpen")
                 menuIsClosed=(action "editorMenuIsClosed")
+                wordCount=wordCount
```
kevinansfield: I think we should be sticking to DDAU rather than relying on implicitly bound immutable data. There's no need to pass a word count into the `gh-koenig` component so it should be possible to pass an action in here instead.

tests/integration/components/gh-koenig-test.js
```diff
@@ -0,0 +1,51 @@
...
+            .then(() => {
+                let {editor} = window;
+                editor.element.focus();
+                inputText(editor, 'this is a test.');
+                window.setTimeout(() => {
+                    expect(this.get('wordCount')).to.equal(5);
+                    expect(this.get('changeFired')).to.be.true;
+                    expect(this.get('firstChangeFired')).to.be.true;
+                    done();
+                }, 1100);
```
disordinary: It's a throwback from an earlier iteration where the updates were throttled.

app/styles/layouts/editor.css
```diff
@@ -251,7 +251,19 @@
...
     padding: 13px 15px;
 }
+
+.gh-editor-wordcount {
+     position: fixed;
+     bottom: 0px;
+     padding:10px;
+ }
+
+ @media (max-width: 1200px) {
```
disordinary: The problem is that .gh-editor-header-small kicks in too late, the wordcount is already obscured by text at that point.

resize over wordcount

app/mixins/editor-base-controller.js
```diff
@@ -41,7 +41,7 @@ export default Mixin.create({
     notifications: injectService(),
     clock: injectService(),
     slugGenerator: injectService(),
-
+    wordCount: 0,
```
disordinary: The actual logic is in the editor, the problem is that the wordcount itself should be outside of the editor element as we may want to place it elsewhere in the UI on smaller displays, etc.

tests/integration/components/gh-koenig-test.js
```diff
@@ -0,0 +1,51 @@
...
+import {editorRendered, inputText} from '../../helpers/editor-helpers';
+
+describe.skip('Integration: Component: gh-koenig - General Editor Tests.', function () {
+    setupComponentTest('gh-koenig', {
+        integration: true
+    });
+
+    beforeEach(function () {
+        // set defaults
+        this.set('changeFired', false);
```
disordinary: Done.

app/styles/layouts/editor.css
kevinansfield: The resize is in JS so it's heavily debounced - it doesn't look like you're giving it time to settle?

app/mixins/editor-base-controller.js
kevinansfield: The `gh-editor` component is the wrapper for the editor pane, it contains the header, title, and editor and seems to be the perfect place for this. The editor base controller will be going away with much of the functionality in it moved to `gh-editor` or into more appropriate places .

PR #675, opened 2017-04-25
[WIP] Only execute a slashmenu tool on enter if it exists.

disordinary:
Prevents the slash menu from deleting the content from the '/' to the end of the range if no command is entered.

issue https://github.com/TryGhost/Ghost/issues/8387
slashmenu-fix

### DIFF
--- a/lib/gh-koenig/addon/components/koenig-slash-menu.js
+++ b/lib/gh-koenig/addon/components/koenig-slash-menu.js
@@ -184,9 +184,12 @@ export default Component.extend({
 
                     let {range} = postEditor;
 
-                    range.head.offset = self.get('range').startOffset - 1;
-                    postEditor.deleteRange(range);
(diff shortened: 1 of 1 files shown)


kevinansfield:
Whilst testing this, I'm not exactly sure how it happened but I got into a state where I wasn't able to type anything with this error:

screen shot 2017-04-25 at 13 16 44

It was after deleting a card but I'm not sure if that was the cause.

disordinary:
Hmm, I can't replicate that (trying various combinations of deleting cards and doing other things).

This PR needs a bit of work.
1. If a user hit's the / and then enter then it still delete the / because there are tools available.
2. Is there are no tools available but the menu is still open they need to press enter twice.

I'll move this to a WIP unless we want to merge as is?

kevinansfield:
WIP is fine, I'm don't think it's blocker for beta readiness.

PR #678, opened 2017-04-26
[WIP] Run-loop improvements

disordinary:
No Issue

1. Removes the runloop issue that was introduced and prevented tests from executing.
2. Prefixes code based selectors with the `js-` prefix to conform to Ghost coding standards.
3. Re-enables tests which were skipped because of 1.

### DIFF
--- a/lib/gh-koenig/addon/components/gh-koenig.js
+++ b/lib/gh-koenig/addon/components/gh-koenig.js
@@ -27,6 +27,7 @@ export default Component.extend({
     editedCard: null,
     keyDownHandler: [],
     resizeEvent: 0,
+    inRunLoop: false,
     init() {
         this._super(...arguments);
         let mobiledoc = this.get('value') || BLANK_DOC;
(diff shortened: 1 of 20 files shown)


kevinansfield:
@disordinary the `.js-*` classes were deprecated a long time ago, the only instances of them in the codebase are legacy code that hasn't been touched in a long time. Recommended way to go for exposing elements for tests is [ember-test-selectors](https://github.com/simplabs/ember-test-selectors) - the data attributes are removed in production builds.

disordinary:
We should probably amend this guide then: https://github.com/TryGhost/Ghost/wiki/Code-standards#frontend-development-standards-html--css :)

kevinansfield:
Ha, yep! I wasn't aware that page even existed 🙈

lib/gh-koenig/addon/components/gh-koenig.js
```diff
@@ -90,8 +91,18 @@ export default Component.extend({
         }
         let editor = this.get('editor');
 
-        editor.didRender(() => {
+        editor.willRender(() => {
+            if (!run.currentRunLoop) {
+                run.begin();
+                this.set('inRunLoop', true);
```
kevinansfield: For purely internal properties there's no need for get/set - you only need those when you want KVO bindings (and the additional overhead that goes with them). Also, if the property is not exposed to the outside world the convention is to prefix with an underscore. Eg.

```
if (this._inRunLoop) {
    run.end();
    this._inRunLoop = false;
}
```

It would also be useful to add a comment here explaining why manual runloop tracking is needed - I had to look it up in the [ember-mobiledoc-editor](https://github.com/bustlelabs/ember-mobiledoc-editor/blob/master/addon/components/mobiledoc-editor/component.js#L230-L235) code

PR #680, opened 2017-04-30
Fix for Spinner invisible issue

rabbihossain:
closes - TryGhost/Ghost#8412

Added the class name for running buttons in `gh-task-button.js`  . Now the running button becomes blue and makes the spinner visible.

### DIFF
--- a/app/components/gh-task-button.js
+++ b/app/components/gh-task-button.js
@@ -32,7 +32,7 @@ const GhTaskButton = Component.extend({
     buttonText: 'Save',
     runningText: reads('buttonText'),
     idleClass: '',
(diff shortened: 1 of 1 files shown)


rabbihossain:
@kevinansfield, can you please check this out? opened it a week before but no one is reviewing it.

kevinansfield:
Hey @rabbihossain, although this works it's not the right solution - we shouldn't be defaulting to different colours for the running state. The svg fill/stroke for the colours where it's not visible needs changing in the CSS instead.

JohnONolan:
This is already fixed with new save button component

PR #690, opened 2017-05-11
✨ use markdown-it for markdown previews

kevinansfield:
refs https://github.com/TryGhost/Ghost/pull/8451
- replaces SimpleMDE's default `marked` rendering with `markdown-it`
- adds `markdown-it` plugins to more closely match legacy Showdown behaviour
  - footnotes
  - highlight/mark
  - named headers
  - don't require a space after the `#` for headers
- adds `ember-browserify` so that markdown plugins that only provide CommonJS modules can be imported


app/utils/format-markdown.js
```diff
@@ -0,0 +1,32 @@
...
+    escapedhtml = md.render(markdown);
+
+    // replace script and iFrame
+    escapedhtml = escapedhtml.replace(/<script\b[^<]*(?:(?!<\/script>)<[^<]*)*<\/script>/gi,
+        '<pre class="js-embed-placeholder">Embedded JavaScript</pre>');
+    escapedhtml = escapedhtml.replace(/<iframe\b[^<]*(?:(?!<\/iframe>)<[^<]*)*<\/iframe>/gi,
+        '<pre class="iframe-embed-placeholder">Embedded iFrame</pre>');
+
+    // sanitize html
+    // jscs:disable requireCamelCaseOrUpperCaseIdentifiers
```
acburdine: not to be too picky, but do these jscs line comments actually do anything?

kevinansfield: Nope, left over from old code. Will be removed shortly, good 🎣 though!

PR #691, opened 2017-05-13
🔥 remove unused ember-responsive addon

kevinansfield:
no issue
- I couldn't find any usage of `ember-responsive` although we do have a similar feature in our custom `media-queries` service

### DIFF
--- a/package.json
+++ b/package.json
@@ -78,7 +78,6 @@
     "ember-power-datepicker": "0.3.0",
     "ember-power-select": "1.8.2",
     "ember-resolver": "4.1.0",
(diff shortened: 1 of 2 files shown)


acburdine:
@kevinansfield it's used by ember-light-table, I think removing it breaks ember-light-table usage (had a similar thing happen on another project using ember-light-table). I'll pull this down and check sometime today, but I'm fairly certain this will break general settings and subscribers

kevinansfield:
Ahh, yeah `ember-light-table` adds it as part of it's blueprint. I'll close this then, it can be removed if needed when we re-visit the subscribers screens.

PR #692, opened 2017-05-15
Fix team screen when logged in as an author

kevinansfield:
closes TryGhost/Ghost#8409
- only show current user on team screen when logged in as author
- don't show the suspended users list to authors


AileenCGN:
Tested and LGTM 👍

PR #695, opened 2017-05-16
🎨  User profile settings page layout updates

AileenCGN:
closes TryGhost/Ghost#7134

Overhaul of the user settings page to make it more consistent with other settings panels.
The hardly readable validation for user "Full Name" is redundant as well, as the input field for it now has the same styles as the other input fields.

image

### DIFF
--- a/app/helpers/gh-count-characters.js
+++ b/app/helpers/gh-count-characters.js
@@ -16,9 +16,9 @@ export function countCharacters(params) {
     el.className = 'word-count';
 
     if (length > 180) {
-        el.style.color = '#E25440';
+        el.style.color = '#f05230';
     } else {
-        el.style.color = '#9E9D95';
+        el.style.color = '#738a94';
     }
 
     el.innerHTML = 200 - length;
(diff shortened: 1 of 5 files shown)


AileenCGN:
What would you reckon it should be? Right now, it looks (afaik) the same as before:

old:
image

new:
image

kevinansfield:
I think it should be the same width as the other inputs. If we want to keep it the same width as it is now then we'd need to add a wrapper around the entire field so that the char count sits at the bottom left of the input instead of out at the side of the screen by itself.

JohnONolan:
agree w/ kevin, except

> so that the char count sits at the bottom left of the input

bottom right

kevinansfield:
> bottom right

:trollface:

/me needs more caffeine

AileenCGN:
image

@kevinansfield, @JohnONolan changes are done!

kevinansfield:
@AileenCGN looks good, only thing left is an update to the tests which are looking for specific colours on the word count

PR #697, opened 2017-05-16
✨ add spellcheck toggle

kevinansfield:
closes https://github.com/TryGhost/Ghost/issues/8461
- adds toggle spellcheck button to the `gh-markdown-editor` toolbar
- adds custom styles to fake a spellcheck icon
- updates `simplemde` fork
  - adds `spellcheck` to the list of toolbar buttons that don't have their `active` class removed based on cursor position
  - removes bundled `marked` dependency that we no longer use

TODO:
- [x] ensure button doesn't lose active state when cursor is placed in content

spellcheck-toggle

### DIFF
--- a/app/components/gh-markdown-editor.js
+++ b/app/components/gh-markdown-editor.js
@@ -107,6 +107,14 @@ export default Component.extend({
                 },
                 '|',
                 {
+                    name: 'spellcheck',
+                    action: () => {
+                        this._toggleSpellcheck();
+                    },
+                    className: 'fa fa-check',
+                    title: 'Toggle Spellcheck'
+                },
+                {
                     name: 'guide',
                     action: () => {
                         this.showMarkdownHelp();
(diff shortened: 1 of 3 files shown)


kevinansfield:
@AileenCGN that's the way CodeMirror's spell check works - it doesn't trigger the spell check until first edit and the behaviour is the same with or without this PR. If we want to fix that it's probably worth opening a separate issue.

AileenCGN:
Ah... Ok. I just found it very confusing, because it seems to be activated, but isn't until I toggle the spellcheck off and on again.

kevinansfield:
There's an open issue on the spell checker plugin here https://github.com/sparksuite/codemirror-spell-checker/issues/28 - proposed solutions (one of which is almost a complete rewrite) haven't been incorporated into the main library. We could fork the spell checker plugin and incorporate a fix into our own simplemde fork if needed.

PR #700, opened 2017-05-17
🐛  Make `cmd+s` work for all save-buttons

AileenCGN:
closes TryGhost/Ghost#8443

Fixes a bug where the keyboard shortcut `cmd+s` would cause a `Maximum call stack size` error and not save.
Wherever there is a `save` button, the keyboard shortcut to save works now.

### DIFF
--- a/app/controllers/settings/apps/amp.js
+++ b/app/controllers/settings/apps/amp.js
@@ -27,6 +27,10 @@ export default Controller.extend({
     actions: {
         update(value) {
             this.set('model', value);
+        },
+
+        save() {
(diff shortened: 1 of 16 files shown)


AileenCGN:
Awesome!! Will adapt it for the other changes as well!

AileenCGN:
Added tests @kevinansfield to wherever we have a `save` button.

AileenCGN:
Hmmm.... That's weird... The tests pass all locally... 🤔

kevinansfield:
It's ✅! 😁

PR #704, opened 2017-05-22
🐛  fix published_at is null

kirrg001:
no issue

I have to investigate how i came into the situation.
Maybe this needs a fix on the server as well.


kirrg001:
I can't reproduce this anymore. Not with importing databases, not with any UI interaction.
Still, this PR a good prevention.

PR #712, opened 2017-05-24
💄🐷 sort-imports eslint rule

kevinansfield:
no issue
- adds `eslint-plugin-sort-imports-es6-autofix` dependency
  - implements ESLint's base `sort-imports` rule but has a distinction in that `import {foo} from 'bar';` is considered `multiple` rather than `single`
  - fixes ESLint's autofix behaviour so `eslint --fix` will actually fix the sort order
- updates all unordered import rules by using `eslint --fix`

With the increased number of `import` statements since Ember+ecosystem started moving towards es6 modules I've found it frustrating at times trying to search through randomly ordered import statements. Recently I've been sorting imports manually when I've added new code or touched old code so I thought I'd add an ESLint rule to codify and automate it.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,16 +1,20 @@
+/* eslint-env node */
 module.exports = {
     root: true,
     parserOptions: {
         ecmaVersion: 2017,
         sourceType: 'module'
     },
+    env: {
+        browser: true
+    },
     extends: [
         'eslint:recommended',
         'plugin:ember-suave/recommended'
     ],
-    env: {
-        browser: true
-    },
+    plugins: [
+        'sort-imports-es6-autofix'
+    ],
     rules: {
         indent: ['error', 4],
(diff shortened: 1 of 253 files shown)


kevinansfield:
Hmm, what was the error? I already restarted it once, it looked it failed when connecting to npm 😕

acburdine:
Looks like a syntax error from the mirage config.js

kevinansfield:
Ugh, it's another node v4 incompatibility 😞 I've been running into more and more frequently lately.

I'll see if there's a quick-fix by forking the `eslint-plugin-sort-imports-es6-autofix` dependency whilst opening a PR.

kevinansfield:
Updated with a fork of the eslint plugin, [upstream PR](https://github.com/marudor/eslint-plugin-sort-imports-es6-autofix/pull/1) has been opened, hopefully it'll be merged and we can remove the custom fork.

PR #729, opened 2017-06-05
🐛  Use new user image properties for image upload

AileenCGN:
closes TryGhost/Ghost#8521

The image uploader modals in `user.hbs` used the old image properties instead of `coverImage` and `profileImage`.


kirrg001:
@kevinansfield merge?🎊

PR #735, opened 2017-06-13
⬆️  Trigger image upload when clicking on image

AileenCGN:
closes TryGhost/Ghost#8544

When clicking on an already uploaded icon, logo or cover in Settings -> General we trigger now the file upload to be able to replace it.

Todo:
- [ ] write tests

### DIFF
--- a/app/templates/settings/general.hbs
+++ b/app/templates/settings/general.hbs
@@ -71,14 +71,14 @@
             </div>
             <div class="gh-setting-action gh-setting-action-smallimg">
                 {{#if model.icon}}
-                    <img class="blog-icon" src="{{model.icon}}" onClick={{action "triggerFileDialog"}} alt="icon" data-test-icon-img>
+                    <img class="blog-icon" src="{{model.icon}}" onclick={{action "triggerFileDialog"}} alt="icon" data-test-icon-img>
                     <button type="button" class="gh-setting-action-smallimg-delete" {{action "removeImage" "icon"}} data-test-delete-image="icon">
                         <span>delete</span>
                     </button>
                 {{else if uploader.isUploading}}
                     {{uploader.progressBar}}
                 {{else}}
-                    <button type="button" class="gh-btn" onClick={{action "triggerFileDialog"}} data-test-image-upload-btn="icon">
+                    <button type="button" class="gh-btn" onclick={{action "triggerFileDialog"}} data-test-image-upload-btn="icon">
                         <span>Upload Image</span>
                     </button>
                 {{/if}}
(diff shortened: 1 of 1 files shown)


app/templates/settings/general.hbs
```diff
@@ -71,7 +71,7 @@
             </div>
             <div class="gh-setting-action gh-setting-action-smallimg">
                 {{#if model.icon}}
-                    <img class="blog-icon" src="{{model.icon}}" alt="icon" data-test-icon-img>
+                    <img class="blog-icon" src="{{model.icon}}" onClick={{action "triggerFileDialog"}} alt="icon" data-test-icon-img>
```
kevinansfield: Standard native event names (even when preceded with `on` to create event handlers) should always be lowercase, e.g. `onclick`. See https://developer.mozilla.org/en-US/docs/Web/Guide/Events/Event_handlers and https://developer.mozilla.org/en-US/docs/Web/Events. camelCase event handlers will work in some browsers but it's recommended to follow the standards.

The only time event names are camelCased in Ember are when using Ember's built-in or custom added event handler methods inside of a component, e.g.:

```
export default Ember.Component.extend({
  click() {
    return false;
  },

  doubleClick() {
     alert("DoubleClickableComponent was double clicked!");
  }
});
```

A full list of Ember event handlers is in the docs https://guides.emberjs.com/v2.13.0/components/handling-events/#toc_event-names

AileenCGN: @kevinansfield Shall we refactor the `onClick` in the `<button>` tags then as well?

Like here: https://github.com/TryGhost/Ghost-Admin/blob/master/app/templates/settings/general.hbs#L81

kevinansfield: 👍 yes please

AileenCGN:
The progress update bar works now for replacing images in general settings. Nevertheless, I wasn't able to mirror these changes in our acceptance tests, as the `fileUpload` fn is triggered without clicking on the upload button or - in this case - on the image.

app/templates/settings/general.hbs
```diff
@@ -71,14 +71,18 @@
             </div>
             <div class="gh-setting-action gh-setting-action-smallimg">
                 {{#if model.icon}}
-                    <img class="blog-icon" src="{{model.icon}}" alt="icon" data-test-icon-img>
-                    <button type="button" class="gh-setting-action-smallimg-delete" {{action "removeImage" "icon"}} data-test-delete-image="icon">
-                        <span>delete</span>
-                    </button>
+                    {{#if uploader.isUploading}}
```
kevinansfield: Rather than duplicating the `if` block and progress bar it may be simpler to re-organise:

```
if uploader.isUploading
    progress
else if model.icon
    icon
else
     upload button
```

AileenCGN: 😬 Sure thing... Guess my logic part was partly switched off

PR #743, opened 2017-06-14
Add subscribers export button class

cobbspur:
closes https://github.com/TryGhost/Ghost/issues/8553

- adds a href for pointer and hover class to subscribers export button


tests/acceptance/subscribers-test.js
```diff
@@ -198,6 +198,7 @@ describe('Acceptance: Subscribers', function() {
 
             // click the import subscribers button
             await click('.gh-btn:contains("Import CSV")');
+            expect(find('.gh-btn:contains("Import CSV")').hasClass('gh-btn-hover-green')).to.be.true;
```
kevinansfield: FYI we don't tend to check for classes in tests unless we're testing for them being added/removed dynamically as the result of some other logic, reason being is that we want the tests to be coupled to logic rather than style.

cobbspur: Okay, I never like to add something without a test :)
Do you want me tot remove the two test additions?

kevinansfield: If you could. I think we only need to keep the addition of the class in `subscribers.hbs`

cobbspur:
@kevinansfield should be good for merge once tests have run

PR #748, opened 2017-06-15
always give editor text area focus by default

kevinansfield:
closes https://github.com/TryGhost/Ghost/issues/8525
- always give focus to the editor content area by default when loading the editor
- change the editor placeholder text


cobbspur:
This seems to contain a regression bug. When editing a slug is PSM there is a collision error

PR #750, opened 2017-06-15
✨ ctrl/cmd-shift-i triggers image file dialog

kevinansfield:
closes https://github.com/TryGhost/Ghost/issues/8523
- use shortcuts mixin to register `cmd-shift-i` shortcut in `gh-markdown-editor` that triggers the image insertion dialog
- joins multiple uploaded images with new lines instead of a space for easier editing post-upload
- add `captureSelection` option to `_openImageFileDialog` so that insertion shortcut triggered when editor doesn't have focus doesn't insert at the last focused char position
- if image insertion happens with no cursor position add two newlines before the inserted image markdown for easier editing (previously it would be inserted immediately after the last char in the post)
- removes unused `editor-shortcuts.js` file


cobbspur:
Note: this works but (on a mac) using cmd+alt+I `` is inserted by simpleMDE. it can be disabled in the options using `drawImage: null` but not sure we want that or not.

cobbspur:
discussed on slack we will merge and make a separate PR to disable simpleMDE drawImage shortcut

PR #752, opened 2017-06-19
🐛 Disable Table support

ErisDS:
- this was never meant to be enabled!
- we don't ever want to support tables via markdown, instead we want to offer a different solution in future

Documentation on how to manage rules: https://github.com/markdown-it/markdown-it#manage-rules
List of rules: https://github.com/markdown-it/markdown-it/issues/289

Ghost has always deliberately not supported markdown tables because it's a compatibility nightmare, and we intend to offer something better in future.

### DIFF
--- a/app/utils/format-markdown.js
+++ b/app/utils/format-markdown.js
@@ -12,6 +12,7 @@ let md = markdownit({
     breaks: true,
     linkify: true
 })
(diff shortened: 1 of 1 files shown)


JohnONolan:
Leaving this in for now

PR #761, opened 2017-06-25
Additional check of privacy.useGravatar for gh-profile-image component

letsjustfixit:
closes TryGhost/Ghost#8612

- added a check for privacy.useGravatar flag (simply useGravatar on client side)
- checked: tests are OK

### DIFF
--- a/app/components/gh-profile-image.js
+++ b/app/components/gh-profile-image.js
@@ -60,17 +60,17 @@ export default Component.extend({
     },
 
     imageBackground: computed('validEmail', 'size', function () {
+        let useGravatar = this.get('config.useGravatar');
         let email = this.get('validEmail');
         let size = this.get('size');
         let style = '';
+        let defaultImageUrl = `url("${this.get('ghostPaths.assetRoot')}/img/user-image.png")`;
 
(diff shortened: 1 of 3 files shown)


kevinansfield:
Hey @letsjustfixit, your code worked, thanks(!) but when reviewing it I realised the `gh-profile-image` code in general had got into quite a state so I've pushed another commit that fixes/improves things in general:

- we had a CP with side-effects 🤢
  - replace CP with basic properties that can be set within the component
  - use `didReceiveAttrs` and `ember-concurrency` to debounce changes to the email property and make the functionality easier to reason about
- fix the broken fade-in animation when the avatar changes
- fix tests - `.to.be.blank` was always returning true, replaced with `to.be.empty` which caused the tests to fail properly then replaced them with the expectations for the actual values

There's still a fair bit of refactoring that can be done here but for a component that's only seen once for the life of the blog it's probably not worth doing 😬 One thing that is still left is to remove the usage of `jquery-file-upload` ([issue here](https://github.com/TryGhost/Ghost/issues/6661)) which should now be simpler with the `gh-uploader` component that was added recently.

PR #762, opened 2017-06-27
🐛  Add CSSNano fix to minify keyframes properly

AileenCGN:
closes TryGhost/Ghost#8610

There's an issue in CSSnano, which we're using in production mode to minify our CSS that caused animation frames to not work properly anymore.
See: https://github.com/ben-eb/gulp-cssnano/issues/33#issuecomment-210518957


AileenCGN:
Please test in production mode, as it doesn't need fixing in dev mode.

PR #763, opened 2017-06-28
💄  Consistent layout for about page

AileenCGN:
closes TryGhost/Ghost#8623

Uses the same classes as the other pages for more consistency and sexiness. Also updated the links to slack and contributers to be correct.

image


JohnONolan:
@AileenCGN is it possible to lose the header on this screen? so basically the logo is at the top instead of "about ghost"

AileenCGN:
I know, right? Doesn't really look good the way it is now.

AileenCGN:
Updated the about page to look like this:

image

PR #765, opened 2017-06-29
Fix enter key doesn't submit sign-in form

NotGrm:
closes TryGhost/Ghost#8588

As discuss in issue, I removed keyPress hook on text-input mixin. This hook was blocking form submission by pressing ENTER key on text input as it prevents event default when key is ENTER.

### DIFF
--- a/app/mixins/text-input.js
+++ b/app/mixins/text-input.js
@@ -46,15 +46,6 @@ export default Mixin.create({
         this._super(...arguments);
     },
 
-    keyPress(event) {
-        // prevent default ENTER behaviour if we have a keyEvent for it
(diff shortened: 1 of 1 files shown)


kevinansfield:
Hey @NotGrm, thanks for this but I'm going to close in favour of https://github.com/TryGhost/Ghost-Admin/pull/771. I added some [further explanation](https://github.com/TryGhost/Ghost/issues/8588#issuecomment-312477888) to the original issue for why this approach wouldn't work as-is.

NotGrm:
Thanks for the feedback @kevinansfield ! Sorry for not replying to you before. 

Will look to fix more issues 😄

PR #766, opened 2017-07-02
Fix for broken image upload button

patrickkim:
closes https://github.com/TryGhost/Ghost/issues/8637
- add a default object to `_openImageFileDialog` args so that destructuring with default params works when not passed a value

### DIFF
--- a/app/components/gh-markdown-editor.js
+++ b/app/components/gh-markdown-editor.js
@@ -311,7 +311,7 @@ export default Component.extend(ShortcutsMixin, {
         delete this._onEditorPaneScroll;
     },
 
-    _openImageFileDialog({captureSelection = true}) {
+    _openImageFileDialog({captureSelection = true} = {}) {
         if (captureSelection) {
             // capture the current selection before it's lost by clicking the
             // file input button


app/components/gh-markdown-editor.js
```diff
@@ -311,7 +311,7 @@ export default Component.extend(ShortcutsMixin, {
         delete this._onEditorPaneScroll;
     },
 
-    _openImageFileDialog({captureSelection = true}) {
+    _openImageFileDialog(captureSelection = true) {
```
kevinansfield: Hey @patrickkim 👋 Thanks for taking a look at this! This is is almost but not quite the right fix, the syntax here is valid - it's using es6 destructuring to simulate named parameters (in es5 it would be an options hash), you can see it used in the function call [here](https://github.com/patrickkim/Ghost-Admin/blob/50eec0f1f6aa696d34cb9c9648f4a9d2da4d7992/app/components/gh-markdown-editor.js#L404).

The problem is that method definition is missing a default value so when a call is made without any parameters (as it is when the [button is clicked](https://github.com/patrickkim/Ghost-Admin/blob/50eec0f1f6aa696d34cb9c9648f4a9d2da4d7992/app/components/gh-markdown-editor.js#L80)) it's failing because it's not possible destructure `undefined`.

The correct fix here would be to ensure that the method can be called without any parameters by supplying a default value in the method definition:

```javascript
_openImageFileDialog({captureSelection = true} = {}) {
```

The named parameters were used because it seemed like the method would require further options. An alternative fix would be to go in the direction you've started in but also update the [usage of the method](https://github.com/patrickkim/Ghost-Admin/blob/50eec0f1f6aa696d34cb9c9648f4a9d2da4d7992/app/components/gh-markdown-editor.js#L404) so that we don't end up capturing the selection when we shouldn't be, we could then revert to named params later on if the need arises.

kevinansfield: FYI, this is a great walkthrough of named parameters in es6 http://exploringjs.com/es6/ch_parameter-handling.html#sec_named-parameters

patrickkim:
Hi Kevin, thanks for reviewing. I looked into both methods suggested. I think setting a default value is the correct approach as it encourages consistent usage while remaining open to the usage of named params in the future. Made changes to reflect.

PR #768, opened 2017-07-04
Focus editor content area by default (attempt 2 😁) 

kevinansfield:
closes TryGhost/Ghost#8525
- always give focus to the editor content area by default when loading the editor
- allow content autosave to work for new posts (it was previously turned off for new posts)
- move transition-on-save behaviour from editor/new controller into the controller mixin's save routine
- cancel background autosave when "are you sure you want to leave?" modal is shown as it can cause the "leave" option to fail because it attempts to delete the post record that can be in flight (plus if we're saving anyway it doesn't make much sense to ask the user  🙈) - this is quite an edge-case as it will only happen if the user makes a content change to a draft post then tries to leave the screen within 3 seconds
- change the editor placeholder text

### DIFF
--- a/app/components/gh-post-settings-menu.js
+++ b/app/components/gh-post-settings-menu.js
@@ -40,7 +40,9 @@ export default Component.extend(SettingsMenuMixin, {
         this._super(...arguments);
 
         this.get('store').query('user', {limit: 'all'}).then((users) => {
-            this.set('authors', users.sortBy('name'));
+            if (!this.get('isDestroyed')) {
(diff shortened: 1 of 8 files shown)


AileenCGN:
Almost... Probably an edge-case, but I ran into a mean issue:

1. Click on new post
2. Focus on title and type anything
3. Go out of editor before autosave kicks in (usually forces the 'are you sure' modal)

image
image


This causes the Admin to crash.

kevinansfield:
Fixed, I hope 🤞

PR #772, opened 2017-07-05
🐛 prevent session sync issues with multiple tabs/refreshes

kevinansfield:
refs https://github.com/TryGhost/Ghost/issues/8616
- only refresh tokens on boot if we last refreshed more than 24hrs ago - this should prevent rapidly changing access/refresh tokens when opening new admin tabs or refreshing whilst other tabs are open
- fix token refresh test which was testing it's own behaviour instead of the applications 🙈

This may not be the full solution to the session issues but it closes one potential culprit and should at least reduce token churn which can only help track down the real cause.

### DIFF
--- a/app/routes/application.js
+++ b/app/routes/application.js
@@ -6,6 +6,7 @@ import Route from 'ember-route';
 import ShortcutsRoute from 'ghost-admin/mixins/shortcuts-route';
 import ctrlOrCmd from 'ghost-admin/utils/ctrl-or-cmd';
 import injectService from 'ember-service/inject';
+import moment from 'moment';
 import observer from 'ember-metal/observer';
 import run from 'ember-runloop';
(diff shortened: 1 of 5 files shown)


kirrg001:
Travis failed 🙈

kirrg001:
Ready to merge. @kevinansfield Could you fix travis please? Thanks!

kevinansfield:
Ugh, linting error. Should be fixed now.

kirrg001:
Still fails 🙈

kevinansfield:
First fail was a random one, subsequent retries incl. cache clears have shown a consistent failure in the tags screen that shouldn't be affected by this PR in any way 😕

I haven't been able to reproduce locally yet.

kevinansfield:
Managed to reproduce with Firefox 54 on Linux, interestingly it was passing in my VM before FF updated itself.

PR #773, opened 2017-07-06
🚀 supercharge tags performance

kevinansfield:
closes https://github.com/TryGhost/Ghost/issues/8540
- use `{{vertical-collection}}` in the tags dropdown filter list, opening the dropdown is now virtually instant as it's not attempting to immediately render components for every tag in the list
- remove pagination/infinite scroll from tags screen
- load all tags when accessing the tags screen
  - will pause to show spinner if no tags have previously been loaded
  - if tags exist in the ember data store, show the list immediately and load/update list in the background
- use `{{vertical-collection}}` to render enough tags to fill the scrollable area with a small buffer and use occlusion and element re-use to swap tags in whilst scrolling (suuuuper fast no matter number of tags loaded)
- scroll tags into view when they are selected (keyboard nav now makes a lot more sense)
- tested with 875 tags and 2x/5x CPU throttling with no major slowdowns 🎉 

TODO:
- [x] implement {{vertical-collection}} for the tags dropdown to speed up rendering
- [ ] ~~see if there's a way to scroll to tags that are in the occluded area~~
  - edge-case, only relevant when accessing the individual tag route directly - everything works at the moment but the active tag isn't scrolled into view
  - may be possible by using tag index and scrolling to calculated height rather than using browser's `element.scrollIntoView()`)

NB: I'm leaving the scroll-to-tag TODO for now, it's a nice-to-have for an edge case and we can re-visit if needed later once the tags screen has been redesigned

### DIFF
--- a/app/components/gh-tag.js
+++ b/app/components/gh-tag.js
@@ -1,15 +1,5 @@
 import Component from 'ember-component';
-import injectService from 'ember-service/inject';
-import {invokeAction} from 'ember-invoke-action';
 
 export default Component.extend({
-    feature: injectService(),
-
-    willDestroyElement() {
-        this._super(...arguments);
-
-        if (this.get('tag.isDeleted') && this.get('onDelete')) {
-            invokeAction(this, 'onDelete');
-        }
-    }
+    tagName: ''
 });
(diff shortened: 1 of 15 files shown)


kirrg001:
When i run `grunt dev` i see now:

```
===================================================================
	Vertical Collection
	:: Including CSS for Visual Debugger
	:: (included in non production builds only)
	:: To use, set {{#vertical-collection debug=true}}
	:: To debug your applied CSS rules, set {{#vertical-collection debugCSS=true}}
===================================================================
```

PR #776, opened 2017-07-08
🐛 empty posts shouldn't be publishable

kevinansfield:
closes TryGhost/Ghost#8501
- don't show publish menu for new posts - the publishmenu options don't make much sense for a post that hasn't been saved yet
- show post status as "new" when a post is unsaved rather than "draft"


kirrg001:
Based on https://github.com/TryGhost/Ghost-Admin/pull/768, waiting with merging!

PR #784, opened 2017-07-17
💄  Theme upload modal style improvements

AileenCGN:
no issue

needs GScan 1.1.4

With GScan sending error details now, the modal was a bit overloaded.

This PR adds a toggle for each error rule which - when clicked - shows the details and the affected files.

TODOs:
- [x] final styles (transitions? font-size? make arrows rotate when toggle is open)
- [x] tests
- [x] make code in `gh-theme-error-li.js` component better?

### DIFF
--- a/app/components/gh-theme-error-li.js
+++ b/app/components/gh-theme-error-li.js
@@ -2,5 +2,12 @@ import Component from 'ember-component';
 
 export default Component.extend({
     tagName: '',
-    error: null
+    error: null,
+    showDetails: false,
+
+    actions: {
+        toggleDetails() {
+            this.toggleProperty('showDetails');
+        }
+    }
 });
--- a/app/templates/components/gh-theme-error-li.hbs
+++ b/app/templates/components/gh-theme-error-li.hbs
@@ -1,13 +1,28 @@
-<li>
-    {{#if error.details}}
-        {{{error.details}}}
-    {{else}}
+<a href="" class="theme-validation-toggle-details" {{action "toggleDetails"}} data-test-toggle-details>
+    <h4 class="theme-validation-rule-text">
         {{{error.rule}}}
-    {{/if}}
+    </h4>
+    <div class="theme-validation-rule-icon">
+        {{#if showDetails}}
+            {{inline-svg "arrow-down"}}
+        {{else}}
+            {{inline-svg "arrow-right"}}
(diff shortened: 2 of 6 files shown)


AileenCGN:
theme_validation_activation_modal

AileenCGN:
@kevinansfield Would love to hear, how I could improve the code in the component... I think, there are ways and you'd probably know them ;)

app/templates/components/modals/theme-warnings.hbs
```diff
@@ -17,7 +17,9 @@
             </div>
         {{/if}}
         {{#each fatalErrors as |error|}}
-            {{gh-theme-error-li error=error}}
+            <div class="theme-validation-item">
```
kevinansfield: Not too sure about the HTML here, the only direct child that is allowed in a `<ul>` element is an `<li>` element 🤔 If it's working across the browsers we care about then it's probably fine for now but something we should come back to.

Ref: https://html.spec.whatwg.org/multipage/grouping-content.html#the-ul-element

app/templates/components/gh-theme-error-li.hbs
```diff
@@ -1,13 +1,30 @@
-<li>
-    {{#if error.details}}
+<h4 class="theme-valitation-rule">
```
kevinansfield: Minor typo here, should be `validation`

app/components/gh-theme-error-li.js
```diff
@@ -2,5 +2,18 @@ import Component from 'ember-component';
 
 export default Component.extend({
     tagName: '',
-    error: null
+    error: null,
+    showDetails: false,
+
+    actions: {
+        toggleDetails() {
+            let toggleDetails = this.get('showDetails');
```
kevinansfield: Ember has a built in method for toggling properties, you could do `this.toggleProperty('showDetails')`

Alternatively, we pull in the `ember-composable-helpers` addon in Ghost-Admin so it would be possible to do this in the template directly:

```hbs
<a href="" class="theme-validation-toggle-details" {{action (toggle "showDetails")}} data-test-toggle-details>
```

AileenCGN:
This is ready to go, @kevinansfield 🎉 

I applied the review suggestions and tested on Chrome, Firefox and Safari!

PR #785, opened 2017-07-18
Upgrade dependencies for Node 8.x+

sjuxax:
These package bumps include fixes that allow the client to run on Node 8.x+. These are _not_ necessarily the latest upstream packages available; they are the oldest packages that include fixes for Node 8.x.


kevinansfield:
Hey @sjuxax 👋 I appreciate that you may be using Node 8 but we won't be making changes specifically for version 8 until it moves into LTS.

In this case the packages you have upgraded have further reaching consequences than simply changing the version numbers. After 1.0 is released we'll be looking at upgrading all dependencies along with taking their side effects into account.

sjuxax:
These are minor version bumps that swap out long-deprecated calls for equivalent modern calls. I understand the reluctance to make changes to accommodate an unstable target, but I would suggest that some type of patch to make Ghost run on Node 8+ without issue is worthwhile, especially since Ghost is currently in the RC stage for its first major release. Please note that major Linux distributions are already distributing Node 8.x as the default.

It would not look good if Ghost 1.0 were unable to run on the latest stable version of Node at the time of its release. Personally, I think it'd be worth it to create a custom distribution of the upstreams with _just_ the Node compatibility patches backported.

kevinansfield:
@sjuxax all deps will be updated after 1.0 is released and well before Node 8 reaches LTS, there is zero point forking and swapping small parts of major deps for a node version we don't support

sjuxax:
OK. While I disagree that "there is zero point" in being compatible with the latest stable release of Node, I appreciate the input and respect your decision. Cheers.

kevinansfield:
Sorry, I should have clarified further with "there is zero point _at this moment_" - the deps have been in lockdown since we went into RC in order not to introduce unknown bugs. Node 8 support will be here soon enough.

It's also worth noting this repo will only ever be used by _developers_ (the end-product built files are distributed to users where node version is irrelevant as it's all browser code) and we don't encourage developers to use anything higher than our "recommended" version so that we don't introduce backwards incompatible syntax.

PR #786, opened 2017-07-18
🐛 fix guard for inviting owner in setup/three

kevinansfield:
refs https://github.com/TryGhost/Ghost/issues/8692
- fixes the guard on setup/three against inviting the owner user created on setup/two (we were sending invites for every e-mail in the list rather than only valid e-mails)
- adds a check to see if we get any invalid errors from the server and will show those separately to the `x emails failed to send` alert

screen shot 2017-07-18 at 22 19 15


kevinansfield:
@kirrg001 tests are green now ✅

PR #787, opened 2017-07-19
🎨 fix wording and capitalisation on empty stories screen

kevinansfield:
closes https://github.com/TryGhost/Ghost/issues/8689
- remove titlecase and switch `posts` to `stories`

### DIFF
--- a/app/templates/posts.hbs
+++ b/app/templates/posts.hbs
@@ -93,11 +93,11 @@
                 <li class="no-posts-box">
                     <div class="no-posts">
                         {{#if showingAll}}
-                            <h3>You Haven't Written Any Posts Yet!</h3>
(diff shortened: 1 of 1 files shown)


app/templates/posts.hbs
```diff
@@ -93,7 +93,7 @@
                 <li class="no-posts-box">
                     <div class="no-posts">
                         {{#if showingAll}}
-                            <h3>You Haven't Written Any Posts Yet!</h3>
+                            <h3>You haven't written any stories yet!</h3>
                             {{#link-to "editor.new"}}<button type="button" class="gh-btn gh-btn-green gh-btn-lg"><span>Write a new Post</span></button>{{/link-to}}
```
AileenCGN: I think the text in the button should be changed to 'Write a new story' as well?

kevinansfield: 👍 Good catch!

kevinansfield:
@AileenCGN updated, also caught the "no posts match your filter" case which I'd missed 🙈

PR #790, opened 2017-07-20
🐛 fix broken editor at mobile sizes

kevinansfield:
closes https://github.com/TryGhost/Ghost/issues/8657
- add checks for the toolbar being present before attempting to set classes on the buttons


AileenCGN:
Tested with Chrome dev tools and on iPhone using ngrok! Worked well for me!

PR #792, opened 2017-07-20
🎨 use blue instead of yellow for warning alerts

kevinansfield:
closes https://github.com/TryGhost/Ghost/issues/8680


kevinansfield:
Pushed test fixes, should be 💚 this time 😬

PR #799, opened 2017-07-21
🎨 update markdown help modal shortcuts

kevinansfield:
no issue
- match shortcuts to the new editor shortcuts
- add shortcut for ordered lists

screen shot 2017-07-21 at 19 36 13

### DIFF
--- a/app/templates/components/modals/markdown-help.hbs
+++ b/app/templates/components/modals/markdown-help.hbs
@@ -17,7 +17,7 @@
             <tr>
                 <td>**text**</td>
                 <td><strong>Bold</strong></td>
-                <td>Ctrl/⌘ + B </td>
+                <td>Ctrl/⌘ + B</td>
             </tr>
             <tr>
                 <td>*text*</td>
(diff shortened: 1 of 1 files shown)


ErisDS:
Both single code and multiline code shortcuts are the same.  cmd + alt + c definitely works for single line code.

Also cmd + h cycles through 5 levels of headings whereas on LTS this only cycled through 2 and 3. Not sure if this is an agreed upon change, definitely not a blocker just something to note.

kevinansfield:
> Also cmd + h cycles through 5 levels of headings whereas on LTS this only cycled through 2 and 3. Not sure if this is an agreed upon change, definitely not a blocker just something to note.

Yeah, I was wondering how to describe that or if it was even necessary, I guess it may be fairly self-explanatory? The toolbar button has the same behaviour and shows it's shortcut on hover. There's also `Cmd-Shift-H` which cycles backwards but again I wasn't too sure how to describe that succinctly.

kevinansfield:
FYI this was updated to remove the fenced code block.

PR #802, opened 2017-07-28
Admin author url preview

zce:


kevinansfield:
@zce thank you for taking the time to contribute but I don't know what issue this is attempting to fix. If you have found a bug please raise an issue then open a PR against it so that reviewers are able to see what's being fixed.

PR #813, opened 2017-08-01
adjust mobile styles for apps-module (#8744)

farhd:
responsive styles for .apps-configured and .apps-card adjusted, so that elements fit smaller device screens.

### DIFF
--- a/app/styles/layouts/apps.css
+++ b/app/styles/layouts/apps.css
@@ -114,7 +114,12 @@
 .apps-card-meta {
     display: flex;
     flex-direction: column;
-    padding: 0 70px 0 0;
+    padding-right: 70px;
+}
+@media (max-width: 500px) {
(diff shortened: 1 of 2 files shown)


AileenCGN:
Hey @farhd! Thanks for the PR! Usually, we use media queries in Ghost to take care of anything that is not desktop width. The way you implemented it, is the other way around. If you could change it, so it fits better in the Ghost coding style, that would be great!

farhd:
Hi @AileenCGN. thank you for your feedback. I adjusted as you asked.
Question: Do you plan to refactor CSS media-queries to use some custom-properties, because at the moment a developer has to find out what MQ is implicitly 'defined'. i.e. I had to look for other MQs and try to get a feeling for the right one. it would be nicer though to have 3-5 defined values for different screen-width ranges.

PR #819, opened 2017-08-09
[WIP] ✨  Unsplash App feature

AileenCGN:
no issue

- adds a new app to settings/apps
- adds a new transform and UnsplashObject to settings models


AileenCGN:
First screenshots:

image
image
image

app/services/unsplash.js
```diff
@@ -24,6 +23,7 @@ export default Service.extend({
     _columnHeights: null,
     _pagination: null,
 
+    applicationId: or('config.unsplashAPI', 'settings.unsplash.firstObject.applicationId'),
```
kevinansfield: @AileenCGN turned out to be one of those simple things that's easy to overlook in the end, took me a while to spot it too! 🙈 `settings.unsplash` is an array so it's necessary to use `.firstObject` to grab the first item 😖

PR #820, opened 2017-08-10
✨ Unsplash App feature: fetch config from private endpoint + validations

AileenCGN:
refs TryGhost/Ghost#8859

needs TryGhost/Ghost#8865

- treats `settings.unsplash` as Object
- only make Application ID visible when app is enabled
- `fetchPrivate()` method for config service
- adds validations to Unsplash App:
    - no special validation when Application Id is provided via config (-> no input field anyway)
    - unable to save and enable Unsplash, when invalid (whitespaces) or no key is provided
- allows user to validate key against Unsplash API


app/controllers/settings/apps/unsplash.js
```diff
@@ -9,15 +9,17 @@ export default Controller.extend({
...
     testRequestDisabled: empty('model.applicationId'),
 
     save: task(function* () {
         let unsplash = this.get('model');
         let settings = this.get('settings');
 
         try {
-            settings.get('unsplash').clear().pushObject(unsplash);
+            // TODO: Make validation work:
+            // yield unsplash.validate();
```
AileenCGN: @kevinansfield I tried to make the validation for the Unsplash app input and checkbox working, but failed. I commented this line out, which I thought should work, but it doesn't...

app/controllers/settings/apps/unsplash.js
```diff
@@ -9,15 +9,24 @@ export default Controller.extend({
...
     testRequestDisabled: empty('model.applicationId'),
 
     save: task(function* () {
         let unsplash = this.get('model');
         let settings = this.get('settings');
+        let hasValidated = unsplash.get('hasValidated');
+        console.log(this.get('model'));
+
+        if (this.get('config.unsplashAPI')) {
+            hasValidated.pushObject('applicationId');
```
kevinansfield: @AileenCGN I changed this conditional to the following:

```javascript
        // api key field is hidden if set via config so don't validate in that case
        if (!this.get('config.unsplashAPI')) {
            try {
                yield unsplash.validate({property: 'applicationId'});
            } catch (error) {
                console.log('validation failed', error);
                // validation failed, quit early
                if (error === undefined) {
                    return;
                }

                throw error;
            }
        }
```

The `console.log` in this case returned:

```
validation failed ["The validator specified, "unsplashIntegration", did not exist!"]
```

This points to the fact that the validation hasn't been registered in [validation-engine.js](https://github.com/AileenCGN/Ghost-Admin/blob/cc289167816a8943204bbecaa0d5cceda4205518/app/mixins/validation-engine.js#L37-L50)

Hopefully that points you in the right direction 😄

app/templates/settings/apps/unsplash.hbs
```diff
@@ -31,35 +31,36 @@
             <div class="gh-setting-action">
                 <div class="for-checkbox">
                     <label for="isActive" class="checkbox">
+                        {{#gh-form-group errors=model.errors hasValidated=model.hasValidated property="isActive"}}
                         {{one-way-checkbox model.isActive id="isActive" name="isActive" type="checkbox" update=(action "update") data-test-unsplash-checkbox=true}}
```
kevinansfield: To make the testSelectors a little more manageable, rather than:

```
data-test-unsplash-checkbox=true
```

I've stated using the following pattern:

```
data-test-checkbox="unsplash"
data-test-input="unsplash"
data-test-button="send-request"
```

Then in tests you can do things like:

```
fillIn(testSelector('input', 'unsplash'), 'xxxx')
click(testSelector('button', 'send-request')
```

Not essential but I've found it can make things a bit more consistent 😄

app/templates/settings/apps/unsplash.hbs
```diff
@@ -31,35 +31,36 @@
...
-                        {{/gh-form-group}}
+                        <div class="gh-setting-title">Unsplash integration</div>
+                        <div class="gh-setting-desc">Access Unsplash images from within the editor.</div>
+                        <div class="gh-setting-content-extended">
+                            {{#gh-form-group errors=model.errors hasValidated=model.hasValidated property="applicationId"}}
+                                {{gh-input model.applicationId name="unsplash" update=(action "updateId") onenter=(action "save") placeholder="0f387e82271755665bd49683e914856b1e34..." data-test-unsplash-url-input=true}}
+                                {{#unless model.errors}}
+                                    <p>Set up a new Unsplash application <a href="https://unsplash.com/documentation#registering-your-application" target="_blank">here</a>, and grab the Application ID.</p>
+                                {{else}}
+                                    {{gh-error-message errors=model.errors property="model.applicationId"}}
```
kevinansfield: `"model.applicationId"` should just be `"applicationId"` here I think?

app/controllers/setup/two.js
```diff
@@ -169,11 +169,15 @@ export default Controller.extend(ValidationEngine, {
...
+        let promises = [];
+
+        promises.pushObject(this.get('settings').fetch());
+        promises.pushObject(this.get('config').fetchPrivate());
 
         this.get('hasValidated').addObjects(['blogTitle', 'session']);
 
         return this.validate().then(() => {
-            return this.get('settings').fetch()
+            return RSVP.all(promises)
```
kevinansfield: This isn't needed here, we still run `this._afterAuthentication` which loads settings & config - the reason the settings fetch is here is to make sure we have all the settings before saving the blog title rather than fetching everything once auth has completed.

Probably moot anyway as the oauth code will be removed shortly 🙈

kevinansfield: It's probably also why the tests are failing, it's expecting a settings object in the `.then()` but `RSVP.all()` will return an array.

PR #830, opened 2017-08-17
[WIP] 🐵 Feature MailChimp app

AileenCGN:
First implementation of MailChimp app.

- adds new app 'MailChimp' to apps, incl. MailChimp icon
- adds a request function that gets called after focus out
- adds validation for mailchimp API key
- adds mailchimp to settings model incl. transform and MailChimpObject


app/controllers/settings/apps/mailchimp.js
```diff
@@ -0,0 +1,213 @@
+import Controller from 'ember-controller';
+import fetch from 'fetch';
+import injectService from 'ember-service/inject';
+import {alias, empty} from 'ember-computed';
+import {resolve} from 'rsvp';
+import {task} from 'ember-concurrency';
+
+const API_URL = 'https://us16.api.mailchimp.com/3.0';
```
kevinansfield: Just a quick note but this can't stay as a constant, it needs to be generated from the api key - the `us16` bit will change on a per-account basis so it needs to be grabbed from the end of the api key.

AileenCGN:
Closing this, as we're having a feature branch `mailchimp` to work with.

PR #831, opened 2017-08-17
fix iconStyle when icon is a full url

yanke-guo:
this fixes iconStyle when 'icon' is a full url, when custom storage adapter is used. 

Refs: 

https://github.com/TryGhost/Ghost/issues/8885


kevinansfield:
Thanks @yanke-guo 👍

PR #841, opened 2017-08-28
🐛 Fix #tags added via PSM having "public" visibility

kevinansfield:
refs TryGhost/Ghost#8943
- Tag model has an observer watching the `isNew` and `name` properties so that it can set the `visibility: "internal"` property for new tags based on their name starting with a `#`
- PSM creates tags using `store.createRecord('tag', {name: tagName})` which means that the tag is instantiated with the `isNew` and `name` properties already set (unlike the tags screen where an empty tag is first created before the name is set via UI) - because the properties are there at instantiation the observer never fires because it's only watching for changes
- adds the `.on('init')` modifier to the `setVisibilityOnNew` observer so that it's run for the watched properties during instantiation as well as on change

### DIFF
--- a/app/models/tag.js
+++ b/app/models/tag.js
@@ -53,5 +53,5 @@ export default Model.extend(ValidationEngine, {
         if (this.get('isNew') && !this.get('isSaving')) {
             this.setVisibility();
         }
-    })
+    }).on('init')
 });


acburdine:
Wouldn't it be better to call `setVisibility()` in the PSM after the tag is created? I mean it's probably not a huge performance difference but this observer would be fired _anytime_ a tag object is instantiated - even if nothing's done it's still an extra function call.

kevinansfield:
Yes, there's a very minor perf difference if loading large amounts of tags simultaneously and I did considere that option. Ultimately I decided to keep the visibility logic in one place rather than spread it out. There's definitely scope for a later

kevinansfield:
Ffs, mobiles! 😅

AileenCGN:
Works fine for me!! 👍

PR #842, opened 2017-08-29
🐛 Fix post saving indicator only indicating autosaves

kevinansfield:
refs https://github.com/TryGhost/Ghost/issues/8932
- show "Saving..." text when autosave or manual save tasks are running


kirrg001:
Build failed.

kevinansfield:
Pushed a fix. "Saving..." is normally shown for a minimum of 3 seconds so it was causing tests to time out now that it's shown more often.

PR #851, opened 2017-09-06
Checkboxes now have greater contrast when unchecked to make them more visible

JohnONolan:
No issue

Before:
image

After:
image


JohnONolan:
Forgot to account for Nightshift - plz squash

PR #853, opened 2017-09-08
Fixed IE11 incorrect publish menu date/time input sizes

berliners1:
Closes https://github.com/TryGhost/Ghost/issues/8514

These changes should stop the affected elements from breaking outside their container in IE11, and should keep the fields the same sizes they were across other browsers.


kevinansfield:
Thanks @berliners1 👍

PR #854, opened 2017-09-09
Flatten nested components

kevinansfield:
no issue
- the upcoming [Module Unification re-organisation](https://github.com/dgeb/rfcs/blob/module-unification/text/0000-module-unification.md) in Ember will no longer support nested components
- this PR pre-emptively moves our usage of nested components (almost entirely modals) into a flat file structure


AileenCGN:
LGTM 👍

PR #856, opened 2017-09-10
Remove unwanted String.prototype extensions

kevinansfield:
closes https://github.com/TryGhost/Ghost/issues/8958
- `markdown-it-named-headers` pulled in [`string.js`](http://stringjs.com) despite it not being used which added a lot of String.prototype extensions and caused Ember deprecation notices
- moves the short `markdown-it-named-headers` functionality directly into our app code without using `String.js`


AileenCGN:
LGTM 👍

PR #858, opened 2017-09-12
🐛  Fix datepicker bug for post dates in past

AileenCGN:
closes TryGhost/Ghost#8891

Having a very old post date, reverting it to a draft and then planning to publish/schedule it, resulted in the datepicker dropdown navigation to show a the wrong date. This was caused, because in the `gh-date-time-picker.js` component, we used only the `didReceiveAttrs()` hook to populate the needed `_date` property. Once, this property is passed to the `power-datepicker`'s navigation `dp.nav' it wouldn't get updated, when the value was changed.

In this PR, I added a property `_initalDate` to the `gh-date-time-picker` component, which will get populated in the `didReceiveAttrs()` hook with the same values as before my changes and used to render the date in the input field (which is read-only). I additionally added a `didUpdateAttrs()` hook, which then sets the `_date` property after any changes of the blog publish date and gets passed as the value for the `power-datepicker`. Any other changes on the publish date/time reflects this.

- adds `didUpdateAttrs()` to `gh-date-time-picker` component
- adds `_initialDate` property which is used in the read-only input fields and is populated in the `didReceiveAttrs()`

---

I don't know if this is the most correct/elegant way, but I tried many different approaches. It didn't seem to me like the issue was a wrongly set `published_at` time, but rather the value passed to the `power-datepicker` before being updated with the correct value. 

Ember `power-calender`, which is one of the components used in Ember `power-datepicker`, offers a [customisation of the navigation](http://www.ember-power-calendar.com/docs/the-nav) (the date/time that appears in the dropdown calender), but there seems to be no way of doing the same in `power-datepicker`. See [here](https://ember-datepicker.development.pagefrontapp.com/).

The end-result of this PR is this:

datepicker_bug

### DIFF
--- a/app/components/gh-date-time-picker.js
+++ b/app/components/gh-date-time-picker.js
@@ -20,6 +20,9 @@ export default Component.extend({
     _previousTime: '',
     _minDate: null,
     _maxDate: null,
+    // used to populate the initial date value in the date picker, before
+    // it gets updated and re-rendered to reflect those changes.
+    _initialDate: null,
 
     blogTimezone: reads('settings.activeTimezone'),
     hasError: or('dateError', 'timeError'),
@@ -55,13 +58,13 @@ export default Component.extend({
         let blogTimezone = this.get('blogTimezone');
 
         if (!isBlank(date)) {
-            this.set('_date', moment(date));
+            this.set('_initialDate', moment(date));
         } else {
-            this.set('_date', moment().tz(blogTimezone));
+            this.set('_initialDate', moment().tz(blogTimezone));
         }
 
         if (isBlank(time)) {
-            this.set('_time', this.get('_date').format('HH:mm'));
+            this.set('_time', this.get('_initialDate').format('HH:mm'));
         } else {
             this.set('_time', this.get('time'));
         }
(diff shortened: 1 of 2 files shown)


AileenCGN:
This test fails locally for me also on master... _looking_

AileenCGN:
This particular test seems to fail randomly (afaik since a while already)...

kirrg001:
I would like to wait for @kevinansfield's review. This bug is not critical, it's okay to wait for the review.

kevinansfield:
@AileenCGN I think this has been approached at the wrong level - the issue is only about the default "schedule at" date/time in the publish menu whereas this PR modifies the behaviour for every instance of `gh-date-time-picker`.

If you look [here](https://github.com/TryGhost/Ghost-Admin/blob/master/app/components/gh-publishmenu-draft.js#L49) you can see that we already reset the date to "now+5mins" when the "schedule it for later" option is selected.

The bug is that the popup calendar view does not jump immediately even though the date has been changed. I think this is a bug in the calendar component because as soon as you click the `>>` button it jumps to "now + 1 month" because the selection in the calendar is correct, it's just that it's displaying the wrong month.

kevinansfield:
@AileenCGN sorry, I missed some of what you wrote in the PR description. I've implemented a simpler version of this fix in https://github.com/TryGhost/Ghost-Admin/pull/860 and added a comment to the original issue about always setting the default to "now".

Closing this PR for the moment as it looks like there's a simpler alternative.

PR #859, opened 2017-09-13
Home and End keys for md editor now behave like in html textarea control

jamesvhays:
closes TryGhost/Ghost#8775

- CodeMirror (and therefore simplemde) manages wrapped lines differently than text editors designed for prose
- On editor startup, rebind Home, End, Shift+Home, and Shift+End to go to the end of the visible line.


kevinansfield:
Hey @jamesvhays 👋 Thanks for taking a look at this. I'm going to close this as I've implemented a simpler version in our SimpleMDE fork rather than directly in our editor wrapper component. See https://github.com/TryGhost/Ghost-Admin/pull/870

PR #860, opened 2017-09-18
🐛 Fixed "scheduled at" datepicker not showing correct month when opening

kevinansfield:
closes https://github.com/TryGhost/Ghost/issues/8891

- set the `center` option as well as the `selected` option for `{{power-datepicker}}` inside `{{gh-date-time-picker}}` to force the month view to jump to the correct date when the selection changes


AileenCGN:
I can't believe the fix was **that** simple... 😩
Works fine for me!!

PR #862, opened 2017-09-19
🎨 Enabled Unsplash integration by default

kevinansfield:
no issue
- Unsplash integration is enabled by default for all users
- it's no longer necessary to create your own Unsplash application and configure your application ID


kirrg001:
Did you change the code again after i've tested your branch around mid? 
If no, i don't have to test this again 🙈

app/services/unsplash.js
```diff
@@ -61,17 +61,6 @@ export default Service.extend({
         }
     },
 
-    sendTestRequest(testApplicationId) {
```
kirrg001: line 26 can be removed i think
 
> applicationId: or('config.unsplashAPI.applicationId', 'settings.unsplash.applicationId'),

PR #863, opened 2017-09-20
🎨 Link tags story count to filtered stories list

kevinansfield:
no issue
- adds a link to the existing stories count in the tags list that when clicked will transition to the stories screen showing all stories with that tag

tags-link


ErisDS:
I have wanted this forever 😍

PR #869, opened 2017-09-24
Fix bright white box in night shift mode

ifvictr:
Currently, when night shift is enabled, the info box looks like this:

dark-before
It doesn't fit in with the night shift theme -- it's too bright. This PR changes it to this:

dark-after


kevinansfield:
Thanks @ifvictr 😄

PR #871, opened 2017-09-27
🎣  Added modal for unsaved changes in settings

AileenCGN:
closes TryGhost/Ghost#8483

leave-settings-model

- Added a new modal component that gets rendered when leaving general/settings after changes have been done but not saved
- Added test

TODOs:
- [x] clarify behaviour for Twitter and Facebook, as they get saved independently from the save-button as soon as a valid input is given

### DIFF
--- a/app/controllers/settings/general.js
+++ b/app/controllers/settings/general.js
@@ -134,6 +134,45 @@ export default Controller.extend({
             }
         },
 
+        toggleLeaveSettingsModal(transition) {
+            let leaveTransition = this.get('leaveSettingsTransition');
+
+            if (!transition && this.get('showLeaveSettingsModal')) {
+                this.set('leaveSettingsTransition', null);
+                this.set('showLeaveSettingsModal', false);
+                return;
+            }
+
+            if (!leaveTransition || transition.targetName === leaveTransition.targetName) {
+                this.set('leaveSettingsTransition', transition);
+
+                // if a save is running, wait for it to finish then transition
+                if (this.get('save.isRunning')) {
+                    return this.get('save.last').then(() => {
+                        transition.retry();
+                    });
+                }
+
+                // we genuinely have unsaved data, show the modal
+                this.set('showLeaveSettingsModal', true);
+            }
+        },
+
+        leaveSettings() {
+            let transition = this.get('leaveSettingsTransition');
+            let model = this.get('model');
+
+            if (!transition) {
+                this.get('notifications').showAlert('Sorry, there was an error in the application. Please let the Ghost team know what happened.', {type: 'error'});
+                return;
+            }
+
+            // roll back changes on model props
+            model.rollbackAttributes();
+
+            return transition.retry();
+        },
+
         validateFacebookUrl() {
             let newUrl = this.get('_scratchFacebook');
             let oldUrl = this.get('model.facebook');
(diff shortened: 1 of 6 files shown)


AileenCGN:
@kevinansfield 

I need to expand this for dated permalinks, where we use a [CP](https://github.com/TryGhost/Ghost-Admin/blob/master/app/controllers/settings/general.js#L29). I will probably do it similar as in the editor-base-controller, where we watch [specific properties](https://github.com/TryGhost/Ghost-Admin/blob/master/app/mixins/editor-base-controller.js#L24)). Is this the right approach in your opinion?

One other question I have is reg. the social URLs. Both, Facebook and Twitter get saved as soon as the entered value is valid and work independently from the save button. Do we want to keep it that way?

kevinansfield:
> One other question I have is reg. the social URLs. Both, Facebook and Twitter get saved as soon as the entered value is valid and work independently from the save button. Do we want to keep it that way?

No, I think we should be consistent and only save when the save button is pressed. The way settings saves work is that they save _all_ settings so I think it would be a bit confusing if you messed with some settings then they were saved without your knowledge because you touched the twitter/facebook inputs.

AileenCGN:
@kevinansfield that's what I thought as well. Found it a bit confusing. I removed the separate saving logic for the social URLs and unsaved changes get detected / can rollback now as well.

app/controllers/settings/general.js
```diff
@@ -134,6 +134,44 @@ export default Controller.extend({
...
+            if (!transition && this.get('showLeaveSettingsModal')) {
+                this.set('showLeaveSettingsModal', false);
+                return;
+            }
+
+            if (!leaveTransition || transition.targetName === leaveTransition.targetName) {
+                this.set('leaveSettingsTransition', transition);
+
+                // if a save is running, wait for it to finish then transition
+                if (this.get('saveTasks.isRunning')) {
```
kevinansfield: This has been copied directly from the editor screen but won't work here because the `settings/general` controller does not have a `saveTasks` task group.

AileenCGN:
>This is suffering from the same issue that was fixed in #880. If you try to leave with unsaved settings and close the modal or click Stay then you can't navigate anywhere other than the link you first clicked on.

I implemented the same fix as used in #880 here to prevent this behaviour.

>We can probably do this in separate PRs but the general concept should be expanded to cover the Design and the various Apps screens too.

I didn't understand this from the original issue, but will push more commits to also implement this for Design and Apps.

app/controllers/settings/general.js
kevinansfield: @AileenCGN sorry, I wasn't very clear in my message yesterday 😞 There wasn't any need to remove this, preventing the transition until the save is complete is still useful so that errors can be displayed and there's no chance the task will be cancelled by the transition.

There are two approaches for a fix:

1. add the code back and change `get('saveTasks.isRunning'`) to `get('save.isRunning')`
2. add the code back and also add a ` saveTasks` task group that has the `save` task in it

AileenCGN: Gotcha! Was thinking about option 1, but then not sure... 😬

AileenCGN:
I will submit separate PRs to have the same behaviour for Design and Apps, so this PR can be merged itself when it's ok.

PR #877, opened 2017-09-28
🐛 Fixed search behaviour to not use RegExp

AileenCGN:
closes TryGhost/Ghost#8959

- Treated the search input as a literal string rather than `RegExp` to allow characters that need escaping in `RegExp` and disable `RegExp` characters like `|`.
- Replaced any non-word characters in `highlighted-text` fn with escaped characters, so they're working with `RegExp`.

regex_search


kevinansfield:
👍 LGTM 😄

PR #879, opened 2017-09-28
🐛 Fixed incorrect autosave of published posts when leaving editor

kevinansfield:
closes https://github.com/TryGhost/Ghost/issues/9072
- the checks when leaving the editor were detecting that the autosave tasks were running and so forcing an immediate autosave even though the autosave tasks wouldn't do anything
- exit early from autosave tasks if autosave isn't allowed

### DIFF
--- a/app/mixins/editor-base-controller.js
+++ b/app/mixins/editor-base-controller.js
@@ -71,27 +71,29 @@ export default Mixin.create({
 
     // save 3 seconds after the last edit
     _autosave: task(function* () {
+        if (!this.get('_canAutosave')) {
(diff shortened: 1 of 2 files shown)


kirrg001:
Travis failed. Random?

kevinansfield:
Yep, random failure. If you see a failure on `Acceptance: Settings - Tags when logged in loads tag via slug when accessed directly` (search page for `not ok` to see failures) it's safe to restart, that test has a timing issue 😞

kirrg001:
Good to know for next time 👍

PR #881, opened 2017-09-28
🐛 Fixed error when navigating to tag from admin search box

kevinansfield:
closes https://github.com/TryGhost/Ghost/issues/9074
- remove unused pagination route attrs that were interfering with routing service - these were a pre-1.0 hangover from when the tags screen had infinite scroll
- fix `adapter returned an array for the primary data of a 'queryRecord' response` deprecation when loading a tag route directly
- change `Posts` to `Stories` in the search dropdown for consistency

### DIFF
--- a/app/components/gh-search-input.js
+++ b/app/components/gh-search-input.js
@@ -30,7 +30,7 @@ export default Component.extend({
     contentExpiresAt: false,
     currentSearch: '',
 
-    posts: computedGroup('Posts'),
+    posts: computedGroup('Stories'),
     pages: computedGroup('Pages'),
     users: computedGroup('Users'),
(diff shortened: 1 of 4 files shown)


AileenCGN:
LGTM 👍

kevinansfield:
@AileenCGN should be all fixed now.

I reverted the change from `.queryRecord` to `.query` because using `.query` meant that the API wouldn't filter by slug properly and wouldn't raise a 404 if the slug didn't exist which is what the tests picked up. To get around the `queryRecord` deprecations I modified the application serializer's `normalizeResponse` method to conform to Ember Data's expected format.

AileenCGN:
works fine for me!

PR #882, opened 2017-09-29
Fixes error when clearing search input

nithindavid:
Closes TryGhost/Ghost#9075

- Adding validation for search term before calling select.search invokeAction


kevinansfield:
Hey @nithindavid! Thanks for taking a look at this. Unfortunately the approach here masks the underlying problem and has a side-effect of not clearing search results correctly - we want the `search` action to be run even when the search term is blank.

I've opened #883 which addresses the underlying runloop problem and also improves the UX by not losing focus on the input.

PR #883, opened 2017-10-02
🐛 Fixed search losing focus when removing search term

kevinansfield:
closes https://github.com/TryGhost/Ghost/issues/9075
- ensure we re-focus the input element after closing the dropdown so that it's not necessary to click on the search input again after using backspace
- removes unnecessary use of `run.scheduleOnce`
  - this was causing the error in TryGhost/Ghost#9075 because search was happening out of sync with the open/close of the dropdown in turn causing `ember-basic-dropdown` to try measuring non-existent elements
  - alternative fix would be to wrap each open/close, focus, and search call in separate `run.scheduleOnce('afterRender', ...)` calls
- remove use of `invokeAction`, it's not been necessary in Ember for a while now and we should be calling methods directly where possible for easier debugging

### DIFF
--- a/app/components/gh-search-input-trigger.js
+++ b/app/components/gh-search-input-trigger.js
@@ -1,7 +1,5 @@
 import Component from '@ember/component';
-import {invokeAction} from 'ember-invoke-action';
 import {isBlank} from '@ember/utils';
-import {run} from '@ember/runloop';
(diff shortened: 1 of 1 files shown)


AileenCGN:
LGTM 👍

PR #884, opened 2017-10-02
Disable fuzzy auto-linking of URLs in markdown preview

kevinansfield:
refs https://github.com/TryGhost/Ghost/issues/8987
- set `linkify-it` `fuzzyLink` option to false so that it only auto-links URLs starting with `http(s)://` or other valid schemes

### DIFF
--- a/app/utils/format-markdown.js
+++ b/app/utils/format-markdown.js
@@ -40,7 +40,6 @@ let markdownitNamedHeaders = function markdownitNamedHeaders(md) {
     };
 };
 
-// eslint-disable-next-line new-cap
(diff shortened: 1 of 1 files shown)


kirrg001:
When i enter `example.com`, the editor doesn't highlight anything. Correct.

When i enter `www.example.com`, the editor highlights this blue. But the preview shows it's not recognised as link. A bit confusing maybe?

kirrg001:
https://github.com/TryGhost/Ghost-Admin/pull/884#issuecomment-333787761

^ discussed. Nothing we can do about that right now, because of codemirror.

PR #886, opened 2017-10-03
Set background for body in nightmode

vikaspotluri123:
- Refs: TryGhost/Ghost#9093

This pull request _just_ sets the background color of the entire page to the same color as the `.gh-main` area. I don't want to say it closs 9093, because it's just a band-aid fix.

My train of thought for this is there might be other areas that have similar issues, so setting the background color for the body will make these issues less annoying for end users as it should cover up any white areas.

### DIFF
--- a/app/styles/app-dark.css
+++ b/app/styles/app-dark.css
@@ -68,6 +68,7 @@
 
 body {
     color: #e5eff5;
+    background: #263238;
 }
 
 pre {


kevinansfield:
Thanks for reporting/looking into this @vikaspotluri123. I don't believe this fix is needed now as the elastic scroll bug should be fixed by https://github.com/TryGhost/Ghost-Admin/commit/c76db04cf89f9a86748f01ae9a9bb044f53993fb

If you notice any more issues after the next release please open an issue 😄

PR #890, opened 2017-10-09
Limit simultaneous upload requests

kevinansfield:
closes https://github.com/TryGhost/Ghost/issues/9120
- use `ember-concurrency` to enqueue uploads in `{{gh-uploader}}`
- set simultaneous upload limit to 2

To test this out:
1. open web inspector's Network tab and change the throttle to Fast 3g
2. use the image toolbar button to select 4 or more large-ish images
3. watch the network inspector - there should only be 2 upload requests in progress at one time and the progress bar displayed in the admin should reflect the overall upload status


AileenCGN:
I tested it and it works exactly how it should! 👍

PR #893, opened 2017-10-11
[WIP] 🐛 Fixed Slack invalid sending of test notification

AileenCGN:
closes TryGhost/Ghost#9130

- When an invalid Slack Webhook URL is submitted (one, that our validator can detect) and set as invalid, throw an error in the saving process to prevent the tasks in `sendTestNotification` from proceeding.


AileenCGN:
Hmm... Not sure if this is really the right attempt. Maybe @kevinansfield has a better idea how to solve this?

AileenCGN:
This has became irrelevant with https://github.com/TryGhost/Ghost-Admin/commit/e269a218b683dd8dfc471315a74efc5f23d81518

PR #894, opened 2017-10-11
🎨 Move upgrade check to About screen

kevinansfield:
refs https://github.com/TryGhost/Ghost/issues/5071

Upgrade messages are now shown on the About screen rather than as alerts. Notifications that are marked as `top` or `custom` are still shown as alerts to allow for certain upgrade messages to be given more visibility.

- remove old `upgrade-notification` service
- update the `upgrade-status` service:
  - add a `message` property that contains an upgrade notification if any exists
  - add a `handleUpgradeNotification` method that accepts a Notification model instance and extracts the `notification.message` property into a html safe string for use in templates
- when loading server notifications during app boot, pass notifications that aren't marked as `top` or `custom` to the new `handleUpgradeNotification` method
- update the `about.hbs` template to pull the upgrade message from the `upgradeStatus` service

screen shot 2018-01-04 at 18 01 42


kevinansfield:
@kirrg001 ready for review now. I didn't put any limit on how often the check will be performed - it seemed like a lot of additional complexity for what's essentially a pre-optimisation that can easily be worked around by refreshing the about page.

kirrg001:
How does this react for Pro blogs? 

We didn't show any release notifications in the past - we didn't even request the service (disabled by the privacy flag).

kirrg001:
https://github.com/TryGhost/Ghost-Admin/pull/894#issuecomment-336176559

Discussed in slack. We will expose this via a private configuration endpoint.

kirrg001:
This is on hold till i have time to finish https://github.com/TryGhost/Ghost/pull/9123.

kirrg001:
> How does this react for Pro blogs?
> Discussed in slack. We will expose this via a private configuration endpoint.

Can't we use client extensions?

kevinansfield:
> Can't we use client extensions?

We can but it's not a long-term solution, _as far as we know_ Ghost(pro) is the only setup using client extensions at the moment but that's only because we haven't advertised their existence. We could add a specific client extension that is only used on pro for this purpose for better forwards compatibility.

Seeing as we want to expose the update check url via configuration I don't think there's much difference in approach between one and the other?

kirrg001:
> I don't think there's much difference in approach between one and the other?

I think the only advantages i can think of is that with client extensions, Ghost does not have to expose if a blog is pro via a private endpoint. Plus it's a question of which concept to use.

> Seeing as we want to expose the update check url via configuration

As we have to expose the update check url and Ghost needs to know if a blog is Pro anyway, we can  simply expose both information via the (private) configuration endpoint.

kevinansfield:
@kirrg001 rebased and added guards for the `allowUpgradeCheck` config value

kirrg001:
Just leaving this here, because i just noticed. I think there is one thing i forgot about to share. Ghost will only keep the latest release notification in the database. As soon as there is another release available, it overrides the latest. By that we don't have to pull notifications in the admin? 🤔  I have to look up the slack conversation again.

kevinansfield:
> By that we don't have to pull notifications in the admin? 🤔

I think I may be missing something, why does that mean the admin doesn't need to pull notifications?

My understanding from what I can remember (it's been a while since we started this so I could very well be mistaken!):
- general upgrade notification has moved from being an alert to being on the About screen, for this it makes the most sense to check the upgrade service directly otherwise there needs to be some way to filter out upgrade notifications that are shown in alerts
- "forced" upgrade notifications and custom messages are still shown as alerts via the old alerts/notification system where all notifications are pulled during admin bootup

kirrg001:
> general upgrade notification has moved from being an alert to being on the About screen, for this it makes the most sense to check the upgrade service directly otherwise there needs to be some way to filter out upgrade notifications that are shown in alerts

If Ghost only remembers one release notification and x custom notifications, there is no need to pull the notifications from the service here directly.

Imagine. We release 1.19, Ghost caches this notification and returns the release notification to the client. You show the notification in the about page, because it's not a custom one. If your Ghost version is 1.19, then don't show. Now 1.20 get's released, Ghost will override 1.19, because we only keep one release notification at a time for master. Custom notifications are independent from that, they are shown as top bar notifications and never on the about page. You can differentiate between release and custom notifications with `custom: true|false`. The concept of `seen notifications` does not make sense for release notifications. We always keep the latest release notification in "db cache".

The only advantage i can still see is that if the admin client pulls the latest release version from the service directly if you navigate to the about page, it is realtime. The server has a 24hours request delay build in.

kevinansfield:
@kirrg001 updated to fetch upgrade message from the server notifications. When merging please squash the commits and use the PR body as the commit message 😄

kirrg001:
@kevinansfield works great 👍

PR #897, opened 2017-10-17
✨Increase password length for changes and signups

AileenCGN:
refs TryGhost/Ghost#9150
needs TryGhost/Ghost#9152

- Increase the password length in validators for signups, pw resets, and password changes for users to 10 instead of 8
- Update tests

Todos:
- [x] update/add more tests to make sure, existing signins etc. will still work


ErisDS:
This LGTM 👍  @kevinansfield plz check if you're happy with the code 😁

PR #899, opened 2017-10-19
✨ Improved password validations rules

AileenCGN:
refs TryGhost/Ghost#9150

- added a new validator for password validations that will take care of the rules client side
- Passwort rules added:
   - Disallow obviously bad passwords: 1234567890, qwertyuiop, asdfghjkl; and asdfghjklm
   - Disallow passwords that contain the words "password" or "ghost"
   - Disallow passwords that match the user's email address
   - Disallow passwords that match the blog domain or blog title
   - Disallow passwords that include 50% or more of the same characters: 'aaaaaaaaaa', '1111111111' and 'ababababab' for example.
- When changing the own password, the old password is not affected by the new validations
- Validation are running on
   - setup
   - signup
   - password change in Team - User (only new passwords are validated)
   - passwort reset

### DIFF
--- a/app/controllers/reset.js
+++ b/app/controllers/reset.js
@@ -16,6 +16,7 @@ export default Controller.extend(ValidationEngine, {
     notifications: injectService(),
     session: injectService(),
     ajax: injectService(),
+    config: injectService(),
 
     email: computed('token', function () {
         // The token base64 encodes the email (and some other stuff),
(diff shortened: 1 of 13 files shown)


ErisDS:
I gave this a quick test with the server side version, I'm happy for it to go in when @kevinansfield is happy too 😬

PR #901, opened 2017-11-01
🐛 Fixed "Someone else is editing" errors showing when no-one else is editing

kevinansfield:
closes https://github.com/TryGhost/Ghost/issues/8969

Collision detection errors were appearing incorrectly because the save routines in the post settings menu bypass the sequential saves in the main editor controller and so are prone to having stale data if a previous save is still in progress.

- added a new `savePost` task that is part of the `saveTasks` group so that all post saves are sequential
- pass the `savePost` task to the `{{gh-post-settings-menu}}` component
- use `savePost` task in `gh-post-settings-menu` instead of calling `save()` directly on the model instance

**Note:** See https://github.com/TryGhost/Ghost/issues/8969 for local reproduction steps.

### DIFF
--- a/app/components/gh-post-settings-menu.js
+++ b/app/components/gh-post-settings-menu.js
@@ -178,7 +178,7 @@ export default Component.extend(SettingsMenuMixin, {
                 return;
             }
 
-            this.get('model').save().catch((error) => {
+            this.get('savePost').perform().catch((error) => {
                 this.showError(error);
                 this.get('model').rollbackAttributes();
             });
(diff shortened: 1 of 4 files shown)


kirrg001:
Cool! Will review tomorrow morning 👍

kirrg001:
Wasn't able to find a problem with this change. Feel free to self-merge for the release later 👍

kevinansfield:
Ugh, I really need to fix that test which consistently randomly fails on Travis 😞

PR #904, opened 2017-11-06
✨  Increased tag description length to 500

AileenCGN:
no issue

- Increased the possible length of the tag description field from 200 to 500


ErisDS:
@AileenCGN not sure if you saw but it looks like this is legitimately failing tests?

kevinansfield:
@AileenCGN there was a unit test for the `tag-settings` validator that had been missed. Pushed a fix 😄

PR #908, opened 2017-11-11
Improve users table mobile style

martgnz:
Another go at https://github.com/TryGhost/Ghost/issues/8744.

Without changing the markup it's going to be difficult to wrap the badges when we have more than one. It's a bit of a compromise, as it looks fine with a slightly larger screen (iPhone 6 plus) and I want to avoid changing the design.

I've tried to get more space where possible reducing a bit the internal padding.

### DIFF
--- a/app/styles/layouts/apps.css
+++ b/app/styles/layouts/apps.css
@@ -132,7 +132,7 @@
 @media (max-width: 500px) {
     .apps-card-meta {
         flex-basis: 70%;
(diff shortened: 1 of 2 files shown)


kevinansfield:
> Without changing the markup it's going to be difficult to wrap the badges when we have more than one.

Feel free to change the markup if it makes things easier or improves the layout 😄

PR #921, opened 2017-12-06
🐛 Fixed Ctrl/Cmd+S save shortcut not working when tags input has focus

kevinansfield:
closes https://github.com/TryGhost/Ghost/issues/9298
- override the default Ember Power Select behaviour of stopping keydown event propagation for any alpha-numeric key events which was preventing our keyboard shortcuts from being triggerable


AileenCGN:
Works fine for me 👍

PR #925, opened 2017-12-11
Trims the URL before saving or sending test notification

Abijeet:
Towards https://github.com/TryGhost/Ghost/issues/9263
- trimming the URL value in the validator

Having a space in the URL would cause Ghost to complain - 
> The URL must be in a format like https://hooks.slack.com/services/<your personal key>

Now trimming the URL before validating.

Signed-off-by: Abijeet <abijeetpatro@gmail.com>
- [x] Commits are squashed 

### DIFF
--- a/app/validators/slack-integration.js
+++ b/app/validators/slack-integration.js
@@ -5,7 +5,7 @@ export default BaseValidator.create({
     properties: ['url'],
 
     url(model) {
-        let url = model.getWithDefault('url', '').trim();
+        let url = model.get('url');
         let hasValidated = model.get('hasValidated');
 
         // eslint-disable-next-line camelcase
--- a/app/validators/slack-integration.js
+++ b/app/validators/slack-integration.js
@@ -5,7 +5,7 @@ export default BaseValidator.create({
     properties: ['url'],
 
     url(model) {
-        let url = model.get('url');
+        let url = model.getWithDefault('url', '').trim();
         let hasValidated = model.get('hasValidated');
 
         // eslint-disable-next-line camelcase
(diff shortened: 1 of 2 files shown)


app/validators/slack-integration.js
```diff
@@ -5,7 +5,7 @@ export default BaseValidator.create({
     properties: ['url'],
 
     url(model) {
-        let url = model.get('url');
+        let url = model.get('url').trim();
```
kevinansfield: It's possible that the url value is `null` or `undefined`, there should be a guard to protect against that when assuming string methods are available, eg:

```javascript
let url = (model.get('url') || '').trim();
```

kevinansfield: or alternatively:

```javascript
let url = model.getWithDefault('url', '').trim();
```

Abijeet:
```
let url = model.getWithDefault('url', '').trim();
```
Good point, thanks, I'll add that.

Abijeet:
@kevinansfield - Thank you. I've made the changes.

kevinansfield:
Hey @Abijeet 👋 Sorry for the slow response, we've been on Christmas break and I'm just getting back into reviewing everything.

I didn't spot this before but validators should not be manipulating data, only validating what's already set on a model.

I think this would be better off being done in the slack app controller's [`updateUrl` action](https://github.com/TryGhost/Ghost-Admin/blob/master/app/controllers/settings/apps/slack.js#L68):

```
        updateURL(value) {
            this.set('model.url', value.trim());
            this.get('model.errors').clear();
        },
```

That will have the effect of not allowing trailing spaces to be entered in the field so the model never gets invalid data set on it.

Abijeet:
@kevinansfield - Thank you. I'll make the changes over the weekend.

Abijeet:
@kevinansfield - Done.

PR #940, opened 2018-01-04
Fixed Japanese IME popup position

tamurashingo:
closes TryGhost/Ghost#9289
- add css to fix hidden textarea position

## found bug
Japanese IME popup position.
34098096-9f24a208-e41e-11e7-859e-9ca2b51e082e

hidden textarea size and position.
34098036-70c695d8-e41e-11e7-8ac3-97683cf998b7

## fix version
Japanese IME popup position.

hidden textarea size and position.

- [x] Commits are squashed 

### DIFF
--- a/app/styles/layouts/editor.css
+++ b/app/styles/layouts/editor.css
@@ -482,6 +482,13 @@
     background: transparent;
 }
 
+.gh-editor .CodeMirror-wrap > div > textarea {
+    top: 0;
+    height: 26px;
+    min-height: 26px;
+    margin-bottom: -26px;
+}
+
 .gh-editor .CodeMirror pre {
     padding: 0;
     color: color(var(--darkgrey) l(+5%));


kevinansfield:
Hey @tamurashingo 👋 Thanks for taking a look at this!

Quick question, how does the IME input behave when the editor text size varies due to header text or mobile screen sizes?

screen shot 2018-01-04 at 11 07 37

screen shot 2018-01-04 at 11 10 28

It would be good to see if there's a more generic styling approach, CodeMirror is also used on the Code Injection screen and in the code injection panel of the post settings menu, both of which will be affected by our generic `textarea` style.

tamurashingo:
Hi, @kevinansfield Thanks for comment!
I checked the IME behavior in header and mobile.

## header

header

This is not perfect.
In header, it will be good that `textarea` bottom position is the same as the bottom of the header.
(or font size in `textarea` is the same as header)

Original CodeMirror has same problem.
I think this header problem should be owned by CodeMirror.

CodeMirror's behavior

## mobile
mobile screen size

It seems good.


In this PR size is written by absolute value, but it seems that written by relative value is better.
like:
```css
.gh-editor .CodeMirror-wrap > div > textarea {
    top: 0;
    height: 1.6rem;
    min-height: 1.6rem;
    margin-bottom: -1.6rem;
}
```

this approach, media query can be used.

```css
@media (max-width: 960px) {
    .gh-editor .CodeMirror-wrap > div > textarea {
        height: 1.4rem;
        ......
    }
}
```

kevinansfield:
> CodeMirror and SimpleMDE have same problem.
> I think this header problem should be owned by CodeMirror or SimpleMDE.

Agreed, the header issue should probably be raised on CodeMirror's repo.

Happy to merge as-is with the fixed height, thanks! 😄

PR #946, opened 2018-01-09
Update .github/PULL_REQUEST_TEMPLATE.md

PeterDaveHello:
There is no CONTRIBUTING file so there won't be a link to it, the prompt
should be removed in case of misleading.

- [x] Commits are squashed 


kevinansfield:
We have a `.github/CONTRIBUTING.md` file now so this change is no longer needed but thanks for pointing out the file omission! 😄

PeterDaveHello:
Cool

PR #948, opened 2018-01-16
Contributor Role - Client side

acburdine:
refs https://github.com/TryGhost/Ghost/issues/9314

TODOs:
- [x] figure out publish menu for contributor
- [x] update tests

### DIFF
--- a/app/templates/posts.hbs
+++ b/app/templates/posts.hbs
@@ -44,6 +44,7 @@
             {{/power-select}}
             {{/unless}}
 
+            {{#unless session.user.isContributor}}
             {{#power-select
                 selected=selectedTag
                 options=availableTags
@@ -61,6 +62,7 @@
             }}
                 {{tag.name}}
             {{/power-select}}
+            {{/unless}}
         </div>
 
         <div class="gh-contentfilter-right">
--- a/app/templates/posts.hbs
+++ b/app/templates/posts.hbs
@@ -7,6 +7,7 @@
     </header>
     <div class="gh-contentfilter">
         <div class="gh-contentfilter-left">
+            {{#unless session.user.isContributor}}
             {{#power-select
                 selected=selectedType
                 options=availableTypes
@@ -22,8 +23,9 @@
             }}
                 {{type.name}}
             {{/power-select}}
+            {{/unless}}
 
-            {{#unless session.user.isAuthor}}
+            {{#unless session.user.isAuthorOrContributor}}
             {{#power-select
                 selected=selectedAuthor
                 options=availableAuthors
--- a/app/controllers/editor.js
+++ b/app/controllers/editor.js
@@ -80,6 +80,7 @@ export default Controller.extend({
     notifications: service(),
     router: service(),
     slugGenerator: service(),
+    session: service(),
     ui: service(),
 
     /* public properties -----------------------------------------------------*/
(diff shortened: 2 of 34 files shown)


app/templates/posts.hbs
```diff
@@ -7,6 +7,7 @@
     </header>
     <div class="gh-contentfilter">
         <div class="gh-contentfilter-left">
+            {{#unless session.user.isContributor}}
```
kevinansfield: We're hiding the post type dropdown here but from what I can see Contributor's are still able to create pages. Should we either a) not allow Contributors to create pages or b) allow them to filter between posts and pages?

Admittedly this is a bit of an edge case as it's unlikely a contributor will have enough draft posts and pages for the filter to be useful, I think the distinction of a contributor being allowed to create pages is the more important design decision.

app/templates/posts.hbs
```diff
@@ -42,6 +44,7 @@
             {{/power-select}}
             {{/unless}}
 
+            {{#unless session.user.isContributor}}
```
kevinansfield: This is another edge case but if a contributor is collaborating on many drafts with other users who can apply the tags do we want to allow them to filter their post list by tag? 🤔

ErisDS:
The two comments on here are similar, I think - in both cases a dropdown is being hidden?

@acburdine can you explain the reason why these changes were made? Maybe I'm missing something but they seem unnecessary.

acburdine:
So the first dropdown I marked hidden was the tags filter - my thinking behind that one is that since contributors can't (at the moment) edit tags and they only see draft posts in the content list, that most likely they wouldn't be filtering drafts by tag (unless, as @kevinansfield mentioned, they're collaborating on a post with other users who _do_ have the permission to add tags). Totally fine to un-hide this one.

The second dropdown is a bit trickier, as it allows filtering by published, scheduled, and draft posts, as well as featured posts & pages. Am good to un-hide this one as well, it will be a bit more complicated though as the list of options in the dropdown (which is static at the moment) will need to be updated to _only_ have the "page" and "featured post" selections.

ErisDS:
OK - IMO Contributors are allowed to create pages, don't see any reason not to.

I had a little think about this, and it feels like removing these for now _is_ the best course of action. Adding things back later when we have more refined options feels nicer than taking things away.

👍 to leaving this as-is from me.

acburdine:
Tests are (finally) passing - anything else need doing on this @kevinansfield ?

kevinansfield:
> Tests are (finally) passing - anything else need doing on this @kevinansfield ?

Nothing that I can see 😄

kirrg001:
Just leaving to notes here.

Contributors can't publish, but on the overview i see "Published by Contributor" (This was the username i've chosen for my invited contributor). It was written by the contributor, but not published?

I wonder why contributors can't see their own posts when they were published? I find that a bit weird 🙊 I have not found a reason for this decision. Happy to hear it.

kevinansfield:
> I wonder why contributors can't see their own posts when they were published? I find that a bit weird 🙊 I have not found a reason for this decision. Happy to hear it.

Ghost-Admin is only set up for *editing* posts but contributors don't have permission to edit their posts once they've been published although at that point they can view them on the actual blog if they want to.

If we were to list the posts in the admin then it suggests they are editable (which they're not) so we'd need to either A: not allow the posts to be clickable and show that they're read-only, or B: add a "read-only" state for the editor and all associated components.

kirrg001:
Basic functionality works good 👍 

I have not yet looked at the tests (would like to wait because of https://github.com/TryGhost/Ghost-Admin/pull/948#issuecomment-359782483) and i have not yet tried to find security problems (e.g. is there a loophole somewhere). But will do tomorrow.

PR #949, opened 2018-01-16
Refactor new->edit transition to avoid editor re-renders

kevinansfield:
closes TryGhost/Ghost#8287
closes TryGhost/Ghost#8750
- moved all editor model functionality into the new `editor` route+controller
- moved editor template from `editor/edit.hbs` to `editor.hbs`
- refactored `editor` controller and the `editor.new`/`editor.edit` routes to work with a persistent editor rather than a teardown/re-render when transitioning from new->edit
- fixed issue in `{{gh-markdown-editor}}` for `autofocus` behaviour when the component isn't re-rendered
- fixed issue in `{{gh-simplemde}}` that was causing multiple updates to the `value` property when a new value is passed in to the component
- added `{{gh-scheduled-post-countdown}}` component to lower the noise in the editor controller
- removed editor route/controller mixins
- removed the `editor.edit` and `editor.new` controllers

TODOs
- [x] re-implement loading behaviour for the edit route


kirrg001:
When i add a new post and select the title, then type and open the PSM, the PSM closes very fast and auto-saves. Expected? 🤔

kevinansfield:
Yes, that's the same as the current behaviour. Maybe something we can open a separate issue for if it's a pain.

kevinansfield:
Not sure why Travis says it failed, all the tests passed and the coverage report was triggered! Going to merge as-is because it was ✅ and nothing changed apart from commits being squashed.

PR #950, opened 2018-01-18
✨ Improved warning for user delete modal

AileenCGN:
no issue

Make it more clear, that when deleting the user, all content will be deleted as well.


kevinansfield:
Hmm, wording change is fine but we should already have been highlighting how many posts were about to be deleted. If that part isn’t working then it suggests there’s a different problem 🤔

AileenCGN:
Yeah, you're right. Only realised now really, that the warning message was already saying everything pretty clear. I think we can ignore this PR then and just keep going.

PR #951, opened 2018-01-22
⭐️  Improved delete user confirm dialog

AileenCGN:
closes TryGhost/Ghost#9405

Gives the user a clearer message before deleting a user.

User that is also author of posts:
image

User that hasn't written any posts:
image

---------

Todos:
- [x] change wording in modal
- [x] make it look good (@peterzimon)

### DIFF
--- a/app/styles/components/modals.css
+++ b/app/styles/components/modals.css
@@ -105,7 +105,7 @@
 
 .modal-header {
     position: relative;
-    margin-bottom: 18px;
+    margin-bottom: 22px;
 }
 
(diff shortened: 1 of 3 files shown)


AileenCGN:
@peterzimon I changed the wording from "All x posts created by this user" to "x posts created by this user", is it would look very weird if it's only one post ("All 1 post created by this user").

AileenCGN:
The commits can be squashed. I think the first commit message should be enough as a description.

PR #922, opened 2017-12-06
Update `ember-cli-mocha` to v0.15.0

Turbo87:
This PR updates `ember-cli-mocha` to test the changes in https://github.com/emberjs/ember-mocha/pull/173 and https://github.com/ember-cli/ember-cli-mocha/pull/229

PR #954, opened 2018-02-03
tests/gh-task-button: Return Promises instead of using `done()`

Turbo87:
Newer versions of `mocha` will complain that `done()` was called with a `non-error` argument if used this way. This PR is needed to unblock #922.

/cc @kevinansfield


kevinansfield:
👍 thanks 😄

PR #955, opened 2018-02-05
Remove misleading leading slash on tag slug

chris-morgan:
I’m a little fuzzy on why it’s there at all, frankly. If it’s supposed
to model the tag’s URL, it’s failing because it lacks the leading
`/tag`. If it’s supposed to be part of supporting tag hierarchies (which
don’t look to be supported at this stage, though `tag.parent` exists?),
then a leading slash has a *little* merit, but is still not necessary—
`foo/bar` works about as well as `/foo/bar`. In all, I reckon it’s best
to remove the leading slash, at the very least until tag hierarchies are
supported.

### DIFF
--- a/app/templates/components/gh-tag.hbs
+++ b/app/templates/components/gh-tag.hbs
@@ -1,7 +1,7 @@
 <div class="settings-tag" id="gh-tag-{{tag.id}}">
     {{#link-to 'settings.tags.tag' tag class="tag-edit-button"}}
         <span class="tag-title">{{tag.name}}</span>
-        <span class="label label-default">/{{tag.slug}}</span>
+        <span class="label label-default">{{tag.slug}}</span>
 
         {{#if tag.isInternal}}
             <span class="label label-blue">internal</span>


kevinansfield:
It's just a visual indication that the label is related to a URL so that it's not confused with other labels that can be attached to tags such as `internal`. Closing as this is by design.

chris-morgan:
I appeal the decision; it’s *misleading*. If it’s supposed to indicate it’s a tag as distinct from some other kind of label, stylistic conventions should be used (I haven’t seen the other labels you describe, so I’m not certain about them), or perhaps even a `#` prefix, which is a common convention these days.

(I will shut up on this issue beyond this appeal; don’t feel obliged to respond.)

kevinansfield:
Appeal noted 😄 This screen is also due for a complete re-design very soon (you may notice it's out-of-sync with all the other screen designs since 1.0) so I'll raise the concern during that process.

> or perhaps even a `#` prefix

This would be more confusing I think because hash tags have a very specific purpose in Ghost which is to mark them as [internal tags](https://help.ghost.org/article/27-internal-tags).

PR #956, opened 2018-02-06
Escape html for error messages

kirrg001:
no issue

- if we would like to re-add html in our error messages, we have to differentiate between static messages which can contain html and messages which are dynamic and can contain dynamic input, which should not allow html
- e.g. it is always possible that Ghost receives a message from a third party module, which concats static text with input text
- if we want to show html content from an error, we need to come up with a concept
- for now: no support for html in error messages

I just went through the code base, where `<p class="main-error">` is used. Maybe there are more.

@kevinansfield feel free to push if something is missing.


kirrg001:
Server PR is coming in a bit.

kirrg001:
@kevinansfield 🖕🏻?

PR #964, opened 2018-03-05
Fixed "Are you sure" modal shown after toggling private mode- #9472

hey24sheep:
Fixed "Are you sure" modal shown after toggling private mode

Closes #9472
- sets default value of password 
- when private mode is toggled back to initial state

### DIFF
--- a/app/controllers/settings/general.js
+++ b/app/controllers/settings/general.js
@@ -34,6 +34,8 @@ export default Controller.extend({
     _scratchFacebook: null,
     _scratchTwitter: null,
 
+    _initialPasswordVal: null,
+
     isDatedPermalinks: computed('settings.permalinks', {
         set(key, value) {
             this.set('settings.permalinks', value ? '/:year/:month/:day/:slug/' : '/:slug/');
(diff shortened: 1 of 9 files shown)


app/controllers/settings/general.js
```diff
@@ -109,6 +111,8 @@ export default Controller.extend({
             // set a new random password when isPrivate is enabled
             if (isPrivate && this.get('settings.hasDirtyAttributes')) {
                 this.get('settings').set('password', randomPassword());
+            } else {
+                this.get('settings').set('password', this._initialPasswordVal);
```
kevinansfield: Rather than storing the initial value on the controller and requiring a change on the route (which shouldn't really be involved in this) we can instead use Ember Data's change tracking to reset the value and keep all the logic in one place:

*services/settings.js*
```javascript
    changedAttributes() {
        return this.get('content').changedAttributes();
    },
```

*controllers/settings/general.js*
```javascript
        toggleIsPrivate(isPrivate) {
            let settings = this.get('settings');

            settings.set('isPrivate', isPrivate);
            settings.get('errors').remove('password');

            let changedAttrs = settings.changedAttributes();

            // set a new random password when isPrivate is enabled
            if (isPrivate && changedAttrs.isPrivate) {
                settings.set('password', randomPassword());

            // reset the password when isPrivate is disabled
            } else if (changedAttrs.password) {
                settings.set('password', changedAttrs.password[0]);
            }
        },
```

hey24sheep:
Hi @kevinansfield,
I have pushed the requested changes, have a look. Let me know if there is anything else. Thanks.

PR #965, opened 2018-03-06
Fixed -If night-shift is enabled, checkbox has to be clicked three times

hey24sheep:
Fixed -If night-shift is enabled, checkbox has to be clicked three times to disable

Closes #9471
- added check for undefined value
- fixed the condition in feature.js _setAdminTheme


hey24sheep:
@kevinansfield @kirrg001 Kindly have a look and let me know if there is anything for me to do. Thank You.

PR #968, opened 2018-03-13
Multiple authors

kevinansfield:
requires https://github.com/TryGhost/Ghost/pull/9426
- update `embedded-relation-adapter` to fetch deep-nested embedded relationships
- fixed default token component display in {{gh-token-input}}
    - if no `tokenComponent` is passed to `{{gh-token-input}}` then it should default to the ember-drag-drop `draggable-object` component but instead it didn't output anything
    - put `draggable-object` in quotes because `{{component}}` needs a component name rather than an object
    - rename `option` attribute to `content` to match the default `{{draggable-object}}` interface
- add embedded `authors` attr to the Post model
    - ensure authors is populated when starting new post
    - add validation for empty authors list
- swap author dropdown for a token input in PSM
- show all post authors in posts list
- update tests for `post.authors`
  - always provide through an authors array
  - fix mirage serialisation for paginated responses (embedded records were not being serialised)
- unify tags and author inputs design
  - remove highlight of primary tags
  - highlight internal tags
  - remove unnecessary/redundant title attributes on tags
  - use SVG icon for "remove option" button in token inputs

TODO:
- [x] show list of authors on posts list
- [x] ~update permissions to handle multiple authors~ (same as existing perms)
- [x] ensure authors list is populated when creating new post
- [x] handle situation where all authors are removed from the list
- [x] add option to remove "Add x..." option in `{{gh-token-input}}`
- [x] update mirage mocks to use authors list
- [x] tests

### DIFF
--- a/app/adapters/embedded-relation-adapter.js
+++ b/app/adapters/embedded-relation-adapter.js
@@ -115,15 +115,25 @@ export default BaseAdapter.extend({
     getEmbeddedRelations(store, modelName) {
         let model = store.modelFor(modelName);
         let ret = [];
+        let embedded = [];
 
         // Iterate through the model's relationships and build a list
         // of those that need to be pulled in via "include" from the API
-        model.eachRelationship(function (name, meta) {
-            if (meta.kind === 'hasMany'
+        model.eachRelationship((name, meta) => {
+            if (
+                meta.kind === 'hasMany'
                 && Object.prototype.hasOwnProperty.call(meta.options, 'embedded')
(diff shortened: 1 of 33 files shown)


peterzimon:
Unified UI design and improvement on Tags and Multiple Authors input. Key points:
- remove indication for primary tag
- similarly, no indication for primary author
- change style of internal tags to indicate their difference

ui design for author input

kevinansfield:
Updated with new design

PR #969, opened 2018-03-15
Bump ember-cli-mirage to 0.4.2

acburdine:
closes TryGhost/Ghost#9433
- remove unnecessary bower deps
- update mirage usage
- fix tests

Took a bit longer than expected, but got this working as far as I can tell, minus some flaky tests. This will likely need a fairly significant update too once the multi-author stuff is merged in.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -2,11 +2,9 @@
   "name": "ghost-admin",
   "dependencies": {
     "devicejs": "0.2.7",
-    "Faker": "3.1.0",
     "google-caja": "6005.0.0",
     "keymaster": "1.6.3",
     "normalize.css": "3.0.3",
-    "pretender": "1.1.0",
     "rangyinputs": "1.2.0",
     "validator-js": "3.39.0"
   }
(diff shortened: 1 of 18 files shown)


acburdine:
@kevinansfield it looks like tests are failing because of something in Koenig

tests/acceptance/settings/tags-test.js
```diff
@@ -84,7 +84,8 @@ describe('Acceptance: Settings - Tags', function () {
             return authenticateSession(application);
         });
 
-        it('it renders, can be navigated, can edit, create & delete tags', async function () {
+        // TODO: Unskip and fix
+        it.skip('it renders, can be navigated, can edit, create & delete tags', async function () {
```
kevinansfield: What was the problem with these tests?

acburdine: They were failing oddly. When I ran the settings tags acceptance tests specifically, they passed, but when running all acceptance tests or even just the settings acceptance tests, they would fail. I assumed that it might have a similar failure as the other skipped test in this folder so I skipped it. I’ll unskip it though and see if it works.

acburdine:
@kevinansfield looks like the test errors were just me 😅. Added the tests back & updated some mirage serializers to handle the embedded record stuff, so this should be 👍now.

PR #847, opened 2017-09-02
[WIP] Move bower components to npm

kevinansfield:
no issue
- `ember-cli` now has a built-in method of importing files from `node_modules`
- between `ember-cli`'s native behaviour and [`ember-cli-node-assets`](https://github.com/dfreeman/ember-cli-node-assets) we should be able to eliminate bower completely

PR #972, opened 2018-03-18
Bower => NPM

acburdine:
Supersedes #847, requires #969 

This PR will remove all bower dependencies, swapping out the ones we can for npm dependencies and removing ones that aren't used at all.

## Tasks
- [x] Remove validator & clean up validator usage
- [x] Remove remainder of bower components
- [x] clean up node-assets imports
- [x] After #969 is merged, rebase & remove bower dep.


acburdine:
now that #969 is merged i'll clean this up and get it working completely.

kevinansfield:
👍 thanks for looking into this, it's been on the back burner for far too long! 😄

package.json
```diff
@@ -118,6 +118,7 @@
     "simplemde": "https://github.com/kevinansfield/simplemde-markdown-editor.git#ghost",
     "testem": "2.0.0",
     "top-gh-contribs": "2.0.4",
+    "validator": "7.2.0",
```
acburdine: validator's latest version is technically 9.x, however there was a breaking change in 8.x that changed how the isURL function worked with localhost urls. Probably could update it to 9.x, but that should be a separate pr IMO, as the usages of isURL will need to be evaulated. (For what it's worth, Ghost-CLI is stuck on validator 7.2 for the same reason)

kevinansfield: 👍

kevinansfield:
Pushed a rebase to fix yarn.lock conflicts

PR #973, opened 2018-03-18
Replace jquery-deparam with deparam dependency

acburdine:
no issue

- swap out jquery plugin for non-jquery version


acburdine:
Was getting trolled a bit by the npm import stuff, but this should be good to go now.

PR #974, opened 2018-03-19
Prevent Firefox Fieldset Scrolling

CriticalRespawn:
Super small issue but I noticed when I was using the Ghost Admin in Firefox, when I opened the "Invite Team" modal, a scrollbar would briefly appear and the form would be scrollable by 1px. This PR fixes it so it no longer scrolls.

screencast-2018-03-18-23-03-11

### DIFF
--- a/app/styles/components/modals.css
+++ b/app/styles/components/modals.css
@@ -119,7 +119,6 @@
 
 .modal-body {
     position: relative;
-    overflow-y: auto;
 }
 
 .modal-body p {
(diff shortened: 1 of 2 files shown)


kevinansfield:
Hey @CriticalRespawn 👋 Thanks for taking the time to report and look for a fix! I think in this situation the better solution would be to remove [this line](https://github.com/TryGhost/Ghost-Admin/blob/master/app/styles/components/modals.css#L122), it seems to be the cause of the Firefox bug and it's from some very old modal styling which is no longer necessary.

CriticalRespawn:
Thanks for the reply @kevinansfield! 

I've made your suggested fix and updated the PR 😄

PR #979, opened 2018-03-19
Bump ember-light-table dependency

acburdine:
closes TryGhost/Ghost#9434

Just bumping the dependency didn't seem to cause any issues 🤔- I tested it manually as well as ran tests locally. @kevinansfield were there specific things you noticed that were broken?


kevinansfield:
Hmm, I can't remember now, I'm sure when I tried the upgrade previously everything blew up in spectacular fashion so I just left it due to time constraints. Maybe it was just a dodgy module install or a conflict with something else 🤔 

Anyway, this looks fine! Thanks for looking into it 😄

PR #983, opened 2018-03-23
Fix invitation modal email validation

Charizard:
The problem with issue https://github.com/TryGhost/Ghost/issues/7131 is that, on clicking submit button the focus out is triggered on the email box and save almost at the same time which, I am not sure how, but somehow prevents the `sendInvitation` action from being called.

So, I have moved the email validation from `focus-out` to on debounced input. This enables the email to be validated before the user can click on the "Invite" button.

Let me know if this is good from a requirements perspective.

Another option I would suggest is disabling the invite button till a "valid email" is entered.


kevinansfield:
Original issue was resolved via https://github.com/TryGhost/Ghost-Admin/commit/6a8adfec8c8f5b69717a872a149275b3c897193a

PR #988, opened 2018-03-23
updated the position of the close button on modals

CriticalRespawn:
- moved `a.close` into the `header.modal-header` element to correctly align the close button with the modal title

screenshot-2018-3-20 team - testing 1


CriticalRespawn:
Before this PR gets merged, if somebody can just confirm that fix is OK, I'll go ahead and update the other modal files so we can get them all updated in one PR.

Thanks!

kevinansfield:
Hi @CriticalRespawn 👋 We think it's preferable that the close button stays as it is, it's designed to have the disconnection from the title and give more visual priority to the modal content and actions. Thanks for the suggestion though!

PR #990, opened 2018-03-24
Subscribers: Hide Export CSV button if there's no data to export

CriticalRespawn:
- made sense to only show the button to export subscriber data if there’s actually subscriber data to export. If the total number of subscribers is above 0, then the button to export them will show as normal.


kevinansfield:
Hey @CriticalRespawn 👋 I understand the intention here but the "Export CSV" button is purposefully present when there are no subscribers because it allows the download of an empty CSV with the required headers already present as en example for how to import.

CriticalRespawn:
Hey @kevinansfield 

Makes sense! I was looking at https://help.ghost.org/article/10-subscribers-beta before I created this PR and it gives an example of how the CSV should be structured. I didn't realise you could also use the format that the export is downloaded in so my bad! 😄

PR #994, opened 2018-03-27
Admin area responsive issues

CriticalRespawn:
This PR fixes (or attempts to add a fix to) the issues raised in https://github.com/TryGhost/Ghost/issues/9239. I've put each issue below and included a screenshot, details on my fix and the reason I went with the approach I did.

### Password validation for "too short" message
screenshot-2018-4-3 setup - ghost
Rather than add any extra styling here, I've simply changed the message so it's shorter. My thinking was, the user is already told that their password must be at least 10 characters, if it isn't then it's invalid. So rather than repeating ourselves in the error message, just tell them they need to enter a valid password.

---

### Stories screen filters
screencast-2018-03-27-19-45-38

Through the magic of media queries, this PR does not alter anything unless you're on a mobile device. I've changed the filters so that they are vertically scrollable. Tbh, it was the only non-horrific solution I could come up with and unless I dreamt it, I've seen a similar solution being used for something else Ghost related. Maybe on a theme?

---

#### Mobile
screenshot-2018-3-27 editor - testing blog
#### Desktop
screenshot-2018-3-27 editor - testing blog 1

I've removed the actual notification and instead we now show just a line of text underneath the already existing post status. I've also altered the wording slightly so there's less chance of it wrapping. I opted for the light blue so it still stands out.

@kevinansfield I'm tagging you in this as you originally put the issue in! If you need any changes making here, let me know!

### DIFF
--- a/app/styles/layouts/flow.css
+++ b/app/styles/layouts/flow.css
@@ -435,10 +435,7 @@
 }
 
 .gh-flow-content .response {
-    position: absolute;
-    right: 0;
-    bottom: -25px;
-    margin: 0;
+    margin: 0 0 -10px;
     color: #a6b0b3;
     text-align: right;
     font-size: 1.2rem;


CriticalRespawn:
@kevinansfield Hello!

When I run `ember test` locally, I get a few tests failing that have nothing to do with any of the code changes in this PR. Travis keeps failing here on different tests and I'm not sure if it's something I've forgot to change somewhere?

Could you shed some light on these failures please?

Thanks!

kevinansfield:
> Rather than add any extra styling here, I've simply changed the message so it's shorter. My thinking was, the user is already told that their password must be at least 10 characters, if it isn't then it's invalid. So rather than repeating ourselves in the error message, just tell them they need to enter a valid password.

The problem with this is that as soon as anything is typed (or autofilled) in the password field then the placeholder text containing the "minimum 10 chars" message disappears. The error message should be as descriptive as possible for invalid passwords so users know what they need to fix.

> Through the magic of media queries, this PR does not alter anything unless you're on a mobile device. I've changed the filters so that they are vertically scrollable

This has been tried previously but rejected but I'll discuss it and the schedule status changes with our designer tomorrow. I think we'll probably need a more in-depth approach to how to fix these, the schedule status especially is already on our roadmap to review as part of a redesign of the publish menu.

CriticalRespawn:
Hi @kevinansfield 😄 

>The problem with this is that as soon as anything is typed (or autofilled) in the password field then the placeholder text containing the "minimum 10 chars" message disappears. The error message should be as descriptive as possible for invalid passwords so users know what they need to fix.

Understood. I've reverted the error message and tests back to the previous wording. I've altered the styling for the `.gh-flow-content .response` class so it now looks better on mobile devices. I've updated the screenshot in the original comment so you can see the changes.

>This has been tried previously and rejected but I'll discuss it and the schedule status changes with our designer tomorrow. I think we'll probably need a more in-depth approach to how to fix these, the schedule status especially is already on our roadmap to review as part of a redesign of the publish menu.

If these things are on a roadmap to be looked into in the future, perhaps use this PR as a temporary solution until a more definitive solution is ready?

kevinansfield:
The scheduled post message changes have been superseded via https://github.com/TryGhost/Ghost-Admin/commit/150f68e82647c64f5ee1b99ec093aa6aaf8a83ce

As discussed above the scrollable filters implementation was rejected - it's not an obvious interaction and hides customisation.

Closing this for now. @CriticalRespawn in the future please submit separate PRs rather than bunching disparate changes into one PR - separate PRs are easier to review and it means we can potentially merge some changes without others, thanks 😄

PR #995, opened 2018-03-28
updated screenshot on setup screen

CriticalRespawn:
Whilst getting my local ghost development environment set up, I noticed that the screenshot used on Step 1 of the setup process was out of date, quite a bit by the looks of it as it looked nothing like the current admin!

I've created a new screenshot of the same page to be used instead.

#### Before
screenshot-2018-3-28 setup - ghost 1

#### After
screenshot-2018-3-28 setup - ghost 4

This PR also doubles the amount of margin on the top of the `.gh-flow-content .gh-btn` element.


kevinansfield:
Hey @CriticalRespawn 👋 Thanks for bringing the outdated image to our attention! Unfortunately we can't accept this PR at this stage, we need to revisit whether the team screen is really what we want to use here and the shadows in the replacement image do not fit in with our aesthetic.

I'm going open an issue for this and assign our designer, hope you understand 😄

PR #996, opened 2018-03-29
asset rewriting for default user image on setup

CriticalRespawn:
When setting up my local Ghost development environment, after I ran `grunt dev`, I was getting `"GET /ghost/assets/img/user-image.png" 404` in terminal every time I viewed step 2 of the setup process.

Upon further inspection, the image it was trying to find was `/ghost/assets/img/user-image.png` which didn't exactly match any other image during the setup process as the others had some sort of ID in them such as `assets/img/install-welcome-3ae80f63d742f7376413a637c3d1b21f.png`. 

By removing `style={{placeholderStyle}}` and giving it a background URL within the CSS, the image is now successfully found as it's looking for `/ghost/assets/img/user-image-feeb00c017f02c8915d6bc3ffdecf706.png` - note that it now contains an ID, whereas it didn't before.

Functionality appears to be fine, I've tested the different instances I could think of here and screenshot the results below.

Before:
screenshot-2018-3-29 setup - ghost

After:
screenshot-2018-3-29 setup - ghost 1

After (Gravatar):
screenshot-2018-3-29 setup - ghost 2

After (Manual Upload):
screenshot-2018-3-29 setup - ghost 3

I noticed that `placeholderStyle` is defined in the `app/components/gh-profile-image.js` file... I can't see any other usages of it so if this fix is okay, would it be wise to delete references to it in that file too?

### DIFF
--- a/app/components/gh-profile-image.js
+++ b/app/components/gh-profile-image.js
@@ -43,7 +43,8 @@ export default Component.extend({
     init() {
         this._super(...arguments);
 
-        this._defaultImageUrl = `${this.get('ghostPaths.assetRoot')}img/user-image.png`;
+        let defaultImage = '/img/user-image.png';
+        this._defaultImageUrl = `${this.get('ghostPaths.assetRoot')}${defaultImage}`;
         this._setPlaceholderImage(this._defaultImageUrl);
     },
 
--- a/app/styles/layouts/flow.css
+++ b/app/styles/layouts/flow.css
@@ -300,6 +300,7 @@
     display: block;
     width: 90px;
     height: 90px;
+    background: url(assets/img/user-image.png);
     background-color: #f8fbfd;
     background-position: center center;
     background-size: cover;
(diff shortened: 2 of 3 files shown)


kevinansfield:
Hey @CriticalRespawn 👋 The correct fix here is the same as the fixes in https://github.com/TryGhost/Ghost-Admin/commit/cbcf09e4fc96eceff94a451c9548041f05b6d96f - the template strings are tripping up the asset rewriter so we need to split out the local asset name. We'll be getting rid of custom CSS classes soon in favour of our [Tachyons](http://tachyons.io) based framework so it's better to keep URLs in the JS.

CriticalRespawn:
Thanks @kevinansfield I'll take a look and try and get an updated pushed to this branch in the next few days 👍

CriticalRespawn:
Hi @kevinansfield 😄 

I've changed this PR so it now contains the correct fix!

PR #999, opened 2018-04-04
added aria-label to model close for screen readers

CriticalRespawn:
fixes https://github.com/TryGhost/Ghost/issues/8989
- added aria-label to the close button on models so that screen readers
can pick them up which provides a more accessible experience


kevinansfield:
If we're going in this direction then it would also be worth removing the `<span class="hidden">Close</span>` markup as that is currently hidden from both the visual output and the screen reader output.

kevinansfield:
@CriticalRespawn ping RE https://github.com/TryGhost/Ghost-Admin/pull/999#issuecomment-387391128

CriticalRespawn:
@kevinansfield Really sorry for the delay in getting back to this. I've read the comment and will action later on in the week :)

PR #1001, opened 2018-04-04
[SPIKE] Koenig - Second render pass to apply Spirit vertical riddem classes

kevinansfield:
no issue
- walk the DOM after the editor renders to apply relevant `{{kg-style}}` classes for Spirit styling and consistent vertical rhythm


kevinansfield:
Closing this because we've gone in a different direction by using CSS selectors rather than classes

PR #1004, opened 2018-04-10
Replaced config routeKeyword with hardcoded one

AileenCGN:
refs TryGhost/Ghost#9561

- use the hardcoded routeKeyword for `preview`
- removed `routeKeywords` from mirage configurations API fixtures

### DIFF
--- a/app/models/post.js
+++ b/app/models/post.js
@@ -180,10 +180,11 @@ export default Model.extend(Comparable, ValidationEngine, {
         return this.get('ghostPaths.url').join(blogUrl, postUrl);
     }),
 
-    previewUrl: computed('uuid', 'ghostPaths.url', 'config.{blogUrl,routeKeywords.preview}', function () {
+    previewUrl: computed('uuid', 'ghostPaths.url', 'config.blogUrl', function () {
         let blogUrl = this.get('config.blogUrl');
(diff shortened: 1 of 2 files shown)


kirrg001:
@kevinansfield Could you please review this PR tomorrow? 👻
I think we could/should follow this order:

- merge this PR
- update server client ref
- see that everything is fine
- merge https://github.com/TryGhost/Ghost/pull/9561

app/models/post.js
```diff
@@ -180,10 +180,11 @@ export default Model.extend(Comparable, ValidationEngine, {
         return this.get('ghostPaths.url').join(blogUrl, postUrl);
     }),
 
-    previewUrl: computed('uuid', 'ghostPaths.url', 'config.{blogUrl,routeKeywords.preview}', function () {
+    previewUrl: computed('uuid', 'ghostPaths.url', function () {
```
kevinansfield: Minor thing but we still use `config.blogUrl` inside this CP so we should probably keep it in the dependent keys list. I realise `config` is unlikely to change dynamically at the moment but it's better to be consistent with specifying dependent keys for future compatibility and avoiding difficult to spot issues in tests.

AileenCGN: 👍 Oops! Fixed and pushed!

PR #1005, opened 2018-04-12
Fix gravatar resolution

CriticalRespawn:
fixes [#9563](https://github.com/TryGhost/Ghost/issues/9563)
- the usage of escapeExpression was causing the gravatar url to not
render properly in the browser. During setup we don’t use
escapeExpression for Gravatar URLs so we shouldn’t here either

Before:
screenshot-2018-4-12 team - user - testing
After:
screenshot-2018-4-12 team - user - testing 1

### DIFF
--- a/app/controllers/team/user.js
+++ b/app/controllers/team/user.js
@@ -72,9 +72,7 @@ export default Controller.extend({
 
     userImageBackground: computed('user.profileImage', 'userDefault', function () {
         let url = this.get('user.profileImage') || this.get('userDefault');
-        let safeUrl = Handlebars.Utils.escapeExpression(url);
-
-        return htmlSafe(`background-image: url(${safeUrl})`);
+        return htmlSafe(`background-image: url(${url})`);
     }),
     // end duplicated
 


kevinansfield:
The problem with this fix is that in re-introduces a XSS attack vector.

The `url` value still needs to be escaped but we need to find a url-friendly way of doing that. We could use `encodeURI` instead but we run the risk of double-escaping if we're passed an already escaped URL 🤔

CriticalRespawn:
@kevinansfield Understood. Is there anything that would check whether the URL is escaped before escaping it again? I can't think of anything off the top of my head.

hey24sheep:
@kevinansfield @CriticalRespawn Why not use regex?

Just check for '%' in a string if it is encoded or not. We could also decode in a loop to check (if encoded multiple times(I don't know if it is possible but just an edge case)) and check if it changes then it is already escaped.

function isEncoded(url) {
 return /\%/i.test(url);
}

kevinansfield:
Closing this because the proper solution is likely to move away from using a background image via inline styles.

kevinansfield:
Hey @hey24sheep 👋 Sorry, I realised I didn't reply to your suggestion. The idea with the URL encoding is to prevent against XSS attacks, if you give an attacker an easy bypass of including a `%` in the constructed URL then you've negated the reason for the encoding in the first place.

hey24sheep:
@kevinansfield Hi, I see. But, as my other suggestion was to decode in a loop to check if it was already encoded or not, Something like the given snippet. What about that?

kevinansfield:
You can see the solutions in https://github.com/TryGhost/Ghost-Admin/commit/c71437048a73b575a789e1b3ee5a257d9e9be4b0 and https://github.com/TryGhost/Ghost-Admin/commit/902e361e2275dd5bb3e9f8bbad08c1902dac3a6c along with some other general cleanup.

Main points were to drop the HTML encoding and run a URI decode before the URI encode to deal with double-encoding.

hey24sheep:
@kevinansfield Thanks

PR #1014, opened 2018-05-07
Koenig - Making full width images work with closed sidebar

peterzimon:
no issue

### DIFF
--- a/lib/koenig-editor/addon/templates/components/koenig-card-image.hbs
+++ b/lib/koenig-editor/addon/templates/components/koenig-card-image.hbs
@@ -20,7 +20,7 @@
     }}
         <div class="relative">
             {{#if (or previewSrc payload.src)}}
-                <img src={{or previewSrc payload.src}} class="{{kg-style kgImgStyle}} {{if ui.hasSideNav "kg-image-full--sidebar"}}" alt={{payload.alt}}>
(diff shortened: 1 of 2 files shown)


lib/koenig-editor/addon/templates/components/koenig-card-image.hbs
```diff
@@ -20,7 +20,7 @@
     }}
         <div class="relative">
             {{#if (or previewSrc payload.src)}}
-                <img src={{or previewSrc payload.src}} class={{kg-style kgImgStyle}} alt={{payload.alt}}>
+                <img src={{or previewSrc payload.src}} class="{{kg-style kgImgStyle}} {{if ui.hasSideNav "kg-image-full--sidebar"}}" alt={{payload.alt}}>
```
kevinansfield: If the sidebar is present then the `.kg-image-full--sidebar` class will always be present, is that what we want? If it is maybe we can use a more generic class name such as `.kg-image--sidebar` so that we can combine selectors when we need to adjust styles for sidebar/non-sidebar (`.kg-image-full.kg-image--sidebar`)?

Alternatively, we can add parameters to the `{{kg-style}}` helper. Eg. `{{kg-style kgImgStyle sidebar=ui.hasSideNav}}`

PR #1016, opened 2018-05-14
ember-i18n initial integration

ololoken:
Translation is not yet completed, but I need first review


ololoken:
sry, going to rework it

PR #1018, opened 2018-05-20
[BUGFIX perf] don't fingerprint assets in development

runspired:
Doing so causes a double pass through asset-rewrite and forces both to take much longer. This cuts 3s off of rebuilds, bringing them to ~500-700ms on my mac.  Using asset-rewrite as is done in dev mode is also suspicious, likely a better setup could be devised that doesn't require rewrite to interop with the local ghost server. I'd estimate 250ms of savings on rebuilds by doing so.

The end result of the RFC for pre-built addons will likely reduce initial build times by 50%, and rebuilds by 250ms based on my observations using https://github.com/rwjblue/heimdalljs-visualizer

Reducing dependence on packages using ember-browserify would also help, rollup is often a better solution here, but that work overlaps with the RFC work.

### DIFF
--- a/ember-cli-build.js
+++ b/ember-cli-build.js
@@ -92,7 +92,7 @@ module.exports = function (defaults) {
             }
         },
         fingerprint: {
-            enabled: true,
+            enabled: isProduction,
             extensions: ['js', 'css', 'png', 'jpg', 'jpeg', 'gif', 'map']
         },
         minifyJS: {


kevinansfield:
👍 thanks @runspired! Having asset rewrite on in development was intentional because we experienced a few cases where functionality broke in a release due to asset rewrite problems that we hadn't seen in dev with it disabled, especially with our lazy loading implementation.

I'll merge this to disable again and see how we get on now because the perf increase is considerable.

> Using asset-rewrite as is done in dev mode is also suspicious, likely a better setup could be devised that doesn't require rewrite to interop with the local ghost server.

I think I'm missing a bit of context here, is something specific to our app or asset-rewrite in general? Why would rewrite need to interop with the local ghost server?

runspired:
@kevinansfield specific to ghost-admin. Asset-rewrite (separate from fingerprinting, which is why this PR addressed) is usually only done in production, but ghost-admin has set it up to happen in either case. I assumed it was because the local ghost api/asset server is running on a separate domain or port.  If that's not the case, you can decrease your rebuild times during development by another 250ms fairly easily :) 

The rest of your rebuild perf issues though are mostly just addons and ember-browserify usage that could be cleaned up and will mostly be improved by the Prebuilt Addons RFC work.

kevinansfield:
Interesting, we haven't set up `asset-rewrite` ourselves anywhere as far as I'm aware. Perhaps one of our addons is doing it 🤔 unless I'm misunderstanding.

Definitely excited for the prebuilt addons future! 🎉

PR #1021, opened 2018-06-01
fix editor's loading state

evexoio:
closes https://github.com/TryGhost/Ghost/issues/9582

fix editor's loading bug by deleting the loading template

- [X] Commit message has a short title & issue references
- [X] Commits are squashed 
- [X] The build will pass (run `ember test` from the repo root - will be `core/client` if working from the submodule in Ghost).

### DIFF
--- a/app/routes/editor/edit.js
+++ b/app/routes/editor/edit.js
@@ -1,13 +1,18 @@
 import AuthenticatedRoute from 'ghost-admin/routes/authenticated';
 
 export default AuthenticatedRoute.extend({
-    beforeModel() {
+    beforeModel(transition) {
         this._super(...arguments);
-        let editor = this.controllerFor('editor');
-        editor.set('post', null);
-        editor.reset();
+
+        // if the transition is not new->edit, reset the post on the controller
+        // so that the editor view is cleared before showing the loading state
+        if (transition.urlMethod !== 'replace') {
+            let editor = this.controllerFor('editor');
+            editor.set('post', null);
+            editor.reset();
+        }
     },
-    
+
     model(params) {
         let query = {
             id: params.post_id,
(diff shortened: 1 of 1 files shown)


kevinansfield:
Hey @evexoio 👋 Thanks for looking into this but I'm not sure this is the correct fix, the loading template is still required for transitions from a non-editor route to the editor route.

I think the bug is a bit more subtle in that it's only apparent when directly transitioning from one post to another with an async load. You'd need to use the search box to find and select a post whilst already in the editor to see the problem.

kevinansfield:
@evexoio the correct fix here would be to add this to the `editor/edit` route:

```javascript
...
export default AuthenticatedRoute.extend({
    beforeModel() {
        this._super(...arguments);
        let editor = this.controllerFor('editor');
        editor.set('post', null);
        editor.reset();
    },
    ...
```

The reason the `editor/edit-loading.hbs` template was causing problems in the edit->edit transition was that it was being rendered into the `{{outlet}}` in `editor.hbs`. If you have a look in that template you'll see we have a `{{#if post}}` conditional at the top as well as the `{{outlet}}` at the bottom. In the edit->edit transition the editor route's `willTransition` hook isn't called so the controller doesn't get reset meaning we end up with both the old post and the loading template being rendered at the same time.

evexoio:
@kevinansfield that makes sense, I'll fix the PR and push again.

Thanks for reviewing

evexoio:
@kevinansfield PR is fixed, thanks for the instructions

kevinansfield:
@evexoio I had to add a little extra logic to avoid the reset on the new->edit transition where we don't want to blow out the editor model. Without the guard if an autosave triggered a save on a new post it was causing a re-render of the editor which throws off the cursor positioning

PR #1026, opened 2018-07-14
Update index.html

andy1247008998:
favicon.icon should be the icon uploaded from Settings.

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -19,7 +19,7 @@
     <meta name="apple-mobile-web-app-status-bar-style" content="black" />
     <meta name="apple-mobile-web-app-title" content="Ghost" />
 
-    <link rel="shortcut icon" href="assets/img/favicon.ico" />
+    <link rel="shortcut icon" href="{{settings.icon}}" />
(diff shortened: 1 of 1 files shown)


acburdine:
@andy1247008998 So the change you've posted here won't work - as the index.html from the admin client isn't rendered via Handlebars. The handlebars helpers currently present in that file are part of the Ember-CLI build process as opposed to handlebars helpers from Ghost, and thus it's not possible to get the favicon with this approach.

That's not to say there isn't another approach that _would_ work, but for now I'm going to close this PR.

andy1247008998:
@acburdine OK. Thank you. Could you please also fix the index.html in Ghost-core? It's locked to /favicon.ico, which is incorrect when I use external storage.

kevinansfield:
@andy1247008998 /favicon.io is correct even when using a custom storage adapter. You can see the [favicon.ico middleware](https://github.com/TryGhost/Ghost/blob/master/core/server/web/middleware/serve-favicon.js) for the implementation.

PR #1027, opened 2018-07-16
Add "username" in Slack settings

i8ramin:

- [ ] Commit message has a short title & issue references
- [ ] Commits are squashed 


app/transforms/slack-settings.js
```diff
@@ -27,9 +27,9 @@ export default Transform.extend({
         if (isEmberArray(deserialized)) {
             settingsArray = deserialized.map((item) => {
                 let url = (item.get('url') || '').trim();
-                if (url) {
-                    return {url};
-                }
+                let username = (item.get('username') || '').trim();
+
+                return {url, username};
```
i8ramin: This change actually saves empty values for `url` and `username` in the db if the user hits save, but I think that is ok, since an empty value for `url` still makes the feature _inactive_. I can add some conditional logic around this so that the value being stored is still `[]` vs `['"url":"","username":""']`, but didn't think it was necessary.

PR #1043, opened 2018-09-06
[WIP] Simple session prototype

allouis:
This is the client side part of the simple session prototype - it adds a new authenticator, switches the signin controller to use that, and fixes up some general ajax/adapter stuff to work with it.

### DIFF
--- a/app/adapters/base.js
+++ b/app/adapters/base.js
@@ -7,6 +7,9 @@ import {inject as service} from '@ember/service';
 export default RESTAdapter.extend(DataAdapterMixin, AjaxServiceSupport, {
     host: window.location.origin,
     namespace: ghostPaths().apiRoot.slice(1),
+    xhrFields: {
+        withCredentials: true
+    },
 
     session: service(),
 
--- a/app/authenticators/cookie.js
+++ b/app/authenticators/cookie.js
@@ -1,0 +1,42 @@
+import Authenticator from 'ember-simple-auth/authenticators/base';
+import RSVP from 'rsvp';
+import {computed} from '@ember/object';
+import {inject as service} from '@ember/service';
+
+export default Authenticator.extend({
+    ajax: service(),
+    ghostPaths: service(),
+
+    sessionEndpoint: computed('ghostPaths.apiRoot', function () {
+        return `${this.get('ghostPaths.apiRoot')}/session`;
+    }),
+
+    restore: function () {
+        return RSVP.resolve();
+    },
+
+    // disable general token revocation because the requests will always 401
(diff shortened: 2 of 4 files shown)


app/authenticators/cookie.js
```diff
@@ -0,0 +1,42 @@
...
+            contentType: 'application/json;charset=utf-8'
+        };
+
+        return this.get('ajax').post(sessionEndpoint, options);
+    },
+
+    invalidate() {
+        // no need to do anything with the response - the browser will store the cookie in case of a successful request
+        let sessionEndpoint = this.get('sessionEndpoint');
+        return this.get('ajax').del({
```
allouis: I didn't actually test this yet :see_no_evil:

app/authenticators/cookie.js
```diff
@@ -0,0 +1,42 @@
...
+        return `${this.get('ghostPaths.apiRoot')}/session`;
+    }),
+
+    restore: function () {
+        return RSVP.resolve();
+    },
+
+    // disable general token revocation because the requests will always 401
+    // (revocation is triggered by invalid access token so it's already invalid)
+    // we have a separate logout procedure that sends revocation requests
```
allouis: Copypasta :flushed:

app/adapters/base.js
```diff
@@ -7,6 +7,9 @@ import {inject as service} from '@ember/service';
 export default RESTAdapter.extend(DataAdapterMixin, AjaxServiceSupport, {
     host: window.location.origin,
     namespace: ghostPaths().apiRoot.slice(1),
+    xhrFields: {
+        withCredentials: true
```
allouis: I'm pretty sure we need this

kevinansfield: Yeah. 🤔 We could set it only on the login request as that's what's setting the cookie but I guess that limits later separation of API and admin because when the origins are different no cookies will be sent without the `withCredentials` flag.

One thing I would say is that the flag probably belongs in the `ajax` service instead. Ember Data uses the ajax service under the hood via the [AjaxServiceSupport mixin](https://github.com/TryGhost/Ghost-Admin/blob/master/app/adapters/base.js#L1) but we also use the ajax service directly for API calls (NB. the ajax service should only be used for calls to our API, for external calls we should be using `fetch`).

allouis: > I guess that limits later separation of API and admin because when the origins are different no cookies will be sent without the withCredentials flag.

Yeah I think we should deffo support that from the get go - less headaches later.

> One thing I would say is that the flag probably belongs in the ajax service instead.

Makes sense :+1:

allouis:
closing because WIP

PR #1046, opened 2018-09-21
Session auth

allouis:
refs https://github.com/tryghost/ghost/issues/9865
requires https://github.com/TryGhost/Ghost/pull/9912
This updates the client to use the v2/admin api, and the session based authentication. It probably needs more tests, i haven't quite learnt how all the ember testing stuff works though. I've tested locally creating/editing posts, tags, settings and setting up a blog from scratch.


allouis:
We think the tests are failing because of this change: https://github.com/TryGhost/Ghost-Admin/commit/4ba9c6e5824564689cb1812974a1da2fd35e57ae

tests/acceptance/authentication-test.js
```diff
@@ -21,92 +19,6 @@ describe('Acceptance: Authentication', function () {
         destroyApp(application);
     });
 
-    describe('setup redirect', function () {
```
kevinansfield: I think this test is still required?

allouis: I think you're right :+1:

tests/acceptance/signin-test.js
```diff
@@ -1,4 +1,3 @@
-import deparam from 'npm:deparam';
```
kevinansfield: The two places this was used have been removed so we can fully remove this dependency.

kevinansfield:
Code looks fine, there's not too much to test here because the API itself hasn't changed so the existing mirage setup is still fine.

Two other places I'm aware of that will need changes are the [labs](https://github.com/TryGhost/Ghost-Admin/blob/master/app/controllers/settings/labs.js#L131-L142) and [theme](https://github.com/TryGhost/Ghost-Admin/blob/master/app/controllers/settings/design.js#L173-L184) file download routines because they use query-param access tokens in an iframe to trigger browser downloads. I'm assuming we won't have to do anything special with them now because the request will have the cookie attached?

allouis:
Sweet :relaxed: 

I just tested them and they still work :+1: So i'll remove the param from them!

kevinansfield:
Awesome 😄 I'll do some manual testing later then 🙂 I'll hold off on merging until the session auth is merged in Ghost.

tests/unit/authenticators/cookie-test.js
```diff
@@ -0,0 +1,64 @@
+import CookieAuthenticator from 'ghost-admin/authenticators/cookie';
```
allouis: Added authenticator tests here @kevinansfield  :relaxed:

tests/unit/authenticators/cookie-test.js
```diff
@@ -0,0 +1,64 @@
...
+import sinon from 'sinon';
+import {beforeEach, describe, it} from 'mocha';
+import {expect} from 'chai';
+
+describe.only('BaseAuthenticator', () => {
```
kevinansfield: :trollface:

allouis: LOL :sob:

PR #1048, opened 2018-09-24
Added image validation error message to gallery card

rishabhgrg:
closes https://github.com/TryGhost/Ghost/issues/9840
- added error on trying to upload an invalid image format to gallery card
  - updated error message logic to use both overall error method(`onFailed`) to set error message as well as individual upload failures(`onUploadFailure`)

### DIFF
--- a/lib/koenig-editor/addon/components/koenig-card-gallery.js
+++ b/lib/koenig-editor/addon/components/koenig-card-gallery.js
@@ -167,7 +167,7 @@ export default Component.extend({
             this.images.removeObject(image);
 
             this._buildAndSaveImagesPayload();
-            let fileName = (uploadResult.fileName.length > 20) ? `${this.substr(0, 20)}...` : uploadResult.fileName;
+            let fileName = (uploadResult.fileName.length > 20) ? `${uploadResult.fileName.substr(0, 20)}...` : uploadResult.fileName;
             this.set('errorMessage', `${fileName} failed to upload.`);
         },
         handleErrors(errors) {
--- a/lib/koenig-editor/addon/components/koenig-card-gallery.js
+++ b/lib/koenig-editor/addon/components/koenig-card-gallery.js
@@ -168,7 +168,7 @@ export default Component.extend({
 
             this._buildAndSaveImagesPayload();
             let fileName = (uploadResult.fileName.length > 20) ? `${uploadResult.fileName.substr(0, 20)}...` : uploadResult.fileName;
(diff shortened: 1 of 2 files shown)


lib/koenig-editor/addon/components/koenig-card-gallery.js
```diff
@@ -167,8 +167,10 @@ export default Component.extend({
             this.images.removeObject(image);
 
             this._buildAndSaveImagesPayload();
-
-            this.set('errorMessage', 'Some images failed to upload');
+        },
+        handleErrors(errors) {
+            let errorMssg = ((errors[0] && errors[0].message)) || 'Some images failed to upload';
+            this.set('errorMessage', errorMssg);
```
rishabhgrg: `onUploadFailure` was called after individual image upload error, but even in that case error message was collected to be sent to `onFailed`. Since it didn't make sense to have separate functions to set error message as latter is already a superset and called, changed it just set error message here. Todo is to maybe handle and not expose ember error messages directly. Wdyt?

kevinansfield: The downside to doing it this way is that if the error is coming from the server side then the error message won't be displayed until all of the uploads finish, might be a little confusing to see images disappear with no error whilst waiting for a long upload.

I think we can probably use both approaches? That way we get the instant client-side validation error for 
https://github.com/TryGhost/Ghost/issues/9840 whilst also keep the instant error message on upload failure.

rishabhgrg: @kevinansfield Makes sense, have added back instant error message on upload failure, and tied  overall error as well to be shown.

rishabhgrg: In the case you mentioned uploads actually will not start, as currently uploader first validates all files and in case any file fails validation, aborts with validation error message without starting uploads. 😃Ideally validation errors should never mix with others because of this, but definitely some design work is needed to see if this is desired behavior or not and in general how multiple uploads should behave.

kevinansfield: Client-side validation only works in certain cases. The original reason for adding the extra validation/error handling was because Linux wasn't providing enough info for client-side validation to work, in that case the flow is as described above.

rishabhgrg: 🙊 Sorry yeah in that case we'll end up with error messages changing. Should I open a separate issue for design work on how we want to handle these cases?

PR #1054, opened 2018-10-15
Feature image alt text

ntmk:
Added a text box for featured image alt text in the posts settings.

no issue

- SEO team I work with requested this feature
- brand new to ghost - learning curve
- must add field to model in ghost:
    - /core/server/data/schema/schema.js
- any feature_image reference in theme must add:
    - alt='{{featured_alt_text}}' to set the text

### DIFF
--- a/app/components/gh-post-settings-menu.js
+++ b/app/components/gh-post-settings-menu.js
@@ -179,6 +179,16 @@ export default Component.extend(SettingsMenuMixin, {
                 });
         },
 
+        setFeaturedAltText(featuredAltText) {
+            let post = this.get('post');
+            let altText = post.get('featuredAltText');
+            if (featuredAltText === altText) {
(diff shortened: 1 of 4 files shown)


kevinansfield:
@ntmk I appreciate you taking the time out to open a PR but for anything that involves design changes I'd definitely recommend reaching out first - if it's a bug then open an issue or if you're proposing a feature change then via [the forum](https://forum.ghost.org).

We do plan to offer a way to set alt/title data for images but it will be in an integrated fashion across all images in the app.

tagawa:
@ntmk I was looking for the same thing. There's a thread in the Ghost forum here if you'd like to add support for its addition: https://forum.ghost.org/t/how-do-i-add-alt-texts-to-images-when-using-the-koenig-editor/3889

PR #1057, opened 2018-10-19
✨ Added ability to delete integrations

allouis:
Just a WIP to get some feedback on my :fire: ember stuff

Adds delete integration modal, and delete button on the integration page to trigger it.

Missing tests

### DIFF
--- a/app/components/modal-delete-integration.js
+++ b/app/components/modal-delete-integration.js
@@ -1,0 +1,25 @@
+import ModalComponent from 'ghost-admin/components/modal-base';
+import {alias} from '@ember/object/computed';
+import {inject as service} from '@ember/service';
+import {task} from 'ember-concurrency';
+
+export default ModalComponent.extend({
+    router: service(),
+    notifications: service(),
+    integration: alias('model'),
+    actions: {
+        confirm() {
+            this.deleteIntegration.perform();
+        }
+    },
+    deleteIntegration: task(function* () {
+        try {
+            yield this.confirm();
+            this.router.transitionTo('settings.integrations');
+        } catch (error) {
+            this.notifications.showAPIError(error, {key: 'integration.delete.failed'});
+        } finally {
+            this.send('closeModal');
+        }
+    }).drop()
+});
--- a/app/controllers/settings/integration.js
+++ b/app/controllers/settings/integration.js
@@ -61,7 +61,20 @@ export default Controller.extend({
             this.integration.rollbackAttributes();
 
             return transition.retry();
+        },
+
+        deleteIntegration() {
+            this.integration.destroyRecord();
+        },
+
+        confirmIntegrationDeletion() {
+            this.set('showDeleteIntegrationModal', true);
+        },
+
+        cancelIntegrationDeletion() {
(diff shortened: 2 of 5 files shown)


app/controllers/settings/integration.js
```diff
@@ -44,6 +44,11 @@ export default Controller.extend({
                     });
                 }
 
+                // if the model is deleted, allow the transition to complete
+                if (this.integration.get('isDeleted')) {
+                    return setTimeout(() => transition.retry(), 0);
```
allouis: Pretty sure this should be like `Ember.run` or something instead?

app/templates/settings/integration.hbs
```diff
@@ -155,6 +155,9 @@
             </tfoot>
         </table>
     </div>
+    <button class="gh-btn gh-btn-red gh-btn-icon" {{action "confirmIntegrationDeletion" integration}}>
```
allouis: This needs padding

app/controllers/settings/integration.js
```diff
@@ -61,7 +66,20 @@ export default Controller.extend({
...
 
             return transition.retry();
+        },
+
+        deleteIntegration() {
+            this.integration.destroyRecord();
+        },
+
+        confirmIntegrationDeletion(integration) {
+            this.set('integrationToDelete', integration);
```
allouis: Similar pattern to webhooks, feels unnecessary as there's only ever one integration on a page. But it's useful to show the modal, maybe have a `showIntegrationDeleteModal` instead and set to true/false...?

app/templates/settings/integration.hbs
allouis: Or a `<span>` rather

app/controllers/settings/integration.js
kevinansfield: Yeah, a true/false toggle would be clearer here I think

app/controllers/settings/integration.js
kevinansfield: I don't think you need anything here, I would opt for putting a `!isDeleted` conditional in the [`willTransition` hook](https://github.com/TryGhost/Ghost-Admin/blob/190a710c9c9827bba1ba3d6711cbec5e793fcc37/app/routes/settings/integration.js#L49) that decides whether to show the unsaved changes modal

allouis: Ahhhhh okay :+1: i didn't even know about that- thank you thank you!

app/components/modal-delete-integration.js
```diff
@@ -0,0 +1,25 @@
+import ModalComponent from 'ghost-admin/components/modal-base';
+import {alias} from '@ember/object/computed';
+import {inject as service} from '@ember/service';
+import {task} from 'ember-concurrency';
+
+export default ModalComponent.extend({
+    router: service(),
+    notifications: service(),
+    integration: alias('model'),
+    actions: {
```
kevinansfield: Stylistic thing but we generally keep everything spaced apart in Ghost-Admin. Single-line properties can be grouped such as the service imports, and one-line CPs but everything else has line breaks.

PR #1058, opened 2018-10-22
Added ability to upload integration icons

kevinansfield:
no issue
- add `{{aspect-ratio-box}}` to make it easier to adjust box sizes at a fixed ratio based on container height which isn't possible with CSS directly
- used `{{gh-uploader}}` to add upload facility to the icon shown on the integration screen

### DIFF
--- a/app/components/aspect-ratio-box.js
+++ b/app/components/aspect-ratio-box.js
@@ -1,0 +1,51 @@
+import Component from '@ember/component';
+import {assert} from '@ember/debug';
+import {debounce, run} from '@ember/runloop';
+
+export default Component.extend({
+    ratio: '1/1',
+    base: 'height',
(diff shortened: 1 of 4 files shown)


kevinansfield:
@peterzimon this currently has no validation for image type/size/aspect ratio so horizontal images can end up looking a bit odd because they have gaps top/bottom on the integration edit screen and the background repeats on the integrations list.

Should we add a new validation for square icons? We could re-use the `/uploads/profile-image/` endpoint for "square image" validation but there are no size checks on that in case it matters. (/cc @kirrg001 maybe we need a more generic way of specifying image validations server-side?)

kevinansfield:
Merging as-is for now as this is still behind the developer flag. Integrations polishing and final tweaks are already happening on master.

PR #1070, opened 2018-11-12
✨Highlight new integrations directory

JohnONolan:
image


peterzimon:
I'd recommend using Spirit classes whenever it's possible even if you have to mix it with custom classes - this helps us in the process of moving completely to Spirit. For example `.item-id` contains a lot of definitions that already exist in Spirit:

`<a class="id-item"...` 

...could be 

`<a class="id-item flex justify-center items-center w-100 h25 tc darkgrey ba b--lightgrey bg-white br3"`

...and then `id-item` would become

```
    padding: 10px;
    text-decoration: none;
    transition: all .8s ease;
```
( ...from which we could standardize further e.g. `transition` and `text-decoration` later - not in this PR).

Same with `.id-item-logo` where the following attributes could be exchanged with Spirit classes:
```
display: flex;              => flex
justify-content: center;    => justify-center
align-items: center;        => items-center
margin-top: 4px;            => mt1
```

JohnONolan:
@peterzimon I started with that - but quickly removed it because it feels like a really terrible idea to mix both spirit and non-spirit styles on a single element - it needs to be one or the other to be reasonably maintainable. I think that needs to be part of a larger discussion for future

peterzimon:
I think both approaches have pros and cons. Let's leave it like this for now and discuss how we handle this in the future separately.

kevinansfield:
👍 I’ll merge as-is for now then

PR #1074, opened 2018-11-20
Updating Spirit to 0.0.48

peterzimon:
refs. https://github.com/TryGhost/Team/issues/148
- fix vertical rhythm for divs and headings


kevinansfield:
@peterzimon could revert your changes, upgrade your version of `yarn` and then try again? It looks like you're on an old version which has stripped all of the integrity hashes from `yarn.lock`

PR #1077, opened 2018-12-02
🎨 clean up hover state borders in night shift

vikaspotluri123:
- Not sure if there's an issue ref, I couldn't find one
- Improvements to input fields and button borders and colors in night shift on hover / focus
  - Some (at least one) of the issues arise because colors get redefined for the dark theme in spirit, which messes up the final result (due to color-mod being used)
- There definitely needs to be some input re: color choices, they're currently a mish-mash of similar colors from similar rules

### DIFF
--- a/app/styles/app-dark.css
+++ b/app/styles/app-dark.css
@@ -85,6 +85,19 @@ input,
     background: color-mod(var(--lightgrey));
 }
 
+input:focus,
+.gh-input:focus,
+.gh-select:focus,
+.gh-select select:focus {
+    border-color: color-mod(var(--darkgrey) l(-27%) blackness(15%));
+}
+
+.for-checkbox label:hover input:not(:checked) + .input-toggle-component,
+.for-radio label:hover input:not(:checked) + .input-toggle-component {
+    /* light=>dark theme => lightgrey=>middarkgrey */
+    border-color: color-mod(var(--middarkgrey) l(-15%) s(-10%));
+}
+
 .gh-nav,
 .settings-menu-container {
     background: #212A2E;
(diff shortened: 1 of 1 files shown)


app/styles/app-dark.css
```diff
@@ -85,6 +85,19 @@ input,
     background: color-mod(var(--lightgrey));
 }
 
+input:focus,
+.gh-input:focus,
+.gh-select:focus,
+.gh-select select:focus {
+    border-color: color-mod(var(--darkgrey) l(-27%) blackness(15%));
```
peterzimon: I'd suggest to use a little more subtle change (like we do on the regular/non-nightshift theme):
`border-color: color-mod(var(--lightgrey) l(+10%));`

app/styles/app-dark.css
```diff
@@ -85,6 +85,19 @@ input,
...
+}
+
+.for-checkbox label:hover input:not(:checked) + .input-toggle-component,
+.for-radio label:hover input:not(:checked) + .input-toggle-component {
+    /* light=>dark theme => lightgrey=>middarkgrey */
+    border-color: color-mod(var(--middarkgrey) l(-15%) s(-10%));
```
peterzimon: Same here as for the previous one: 
`border-color: color-mod(var(--lightgrey) l(+10%));`

app/styles/app-dark.css
```diff
@@ -151,12 +164,22 @@ input,
     color: color-mod(var(--darkgrey) l(-27%) blackness(+15%));
     text-shadow: none;
 }
+.gh-btn:hover:not(.gh-btn-link) {
+    border-color: color-mod(var(--darkgrey) l(-27%));
```
peterzimon: Also here, I think a little more subtle hover style is better:
```
border-color: color-mod(var(--darkgrey) l(-10%));
color: color-mod(var(--darkgrey) l(-10%));
```

vikaspotluri123:
Thanks @peterzimon! Updated 😊

vikaspotluri123:
Found a couple other places that could use some love:

- Colored action buttons in general

image

- Unsplash integration icon in tags page

image

PR #1078, opened 2018-12-03
Add new theme directory links to Ghost Admin

JohnONolan:
In the same spirit as the recent addition of links to the integrations directory in Ghost Admin, this adds some links and CTAs to a few top themes, the marketplace/directory, and the theme documentation.

All of this is an effort to help Ghost users with discoverability of _what's possible with Ghost_ when browsing the admin area. 

image


peterzimon:
Looks cool!

Just a couple of questions of minor details:
- do we always show the same five themes or are they shuffled in any way?
- do you have the same hover options on the thumbnails as on marketplace? if no, then what happens on click (download link or preview)? all make sense in diff. ways...
- maybe we could style "FREE" in green as on Marketplace?

m1guelpf:
A direct install for free themes would be super cool, probably out of scope tho...

JohnONolan:
> do we always show the same five themes or are they shuffled in any way?

Fixed theme selection and order for balanced quality and colour contrast

> do you have the same hover options on the thumbnails as on marketplace? if no, then what happens on click (download link or preview)? all make sense in diff. ways...

You go to the theme listing, which usually contains link to preview :)

> maybe we could style "FREE" in green as on Marketplace?

I thought about this but maybe too many colours?

JohnONolan:
> A direct install for free themes would be super cool, probably out of scope tho...

Def would be cool in future - think it would probably be one of those things which ends up being at on of work to solve all the different edge cases though

truekasun:
This is a cool feature but when a blog is up and running, we hardly change the live theme to a completely different theme and it's taking an unnecessary space from the design page.

Can we add a property to the config file to hide this section?

Your thoughts on this are appreciated. This is just something came up into my mind when I visit the design page. :)

JohnONolan:
If the site is up and running and you don’t need to change the theme, then
you also won’t need to visit the design settings or ever see it.

truekasun:
Well actually, as a developer I update my theme regularly with news features, optimizations and bugfixes.

So other than the articles page, most of my time is spent there in the design page uploading and re-uploading my theme updates.

app/templates/settings/design.hbs
```diff
@@ -28,8 +29,90 @@
...
+                        <div class="td-cta-content">
+                            <h4>Theme Marketplace</h4>
+                            <p>Explore a huge range of free and premium themes for Ghost with a range of design and layout options</p>
+                        </div>
+                        <div class="td-cta-arrow">
+                            {{svg-jar "arrow-right"}}
+                        </div>
+                    </div>
+                </a>
+                <a class="td-cta-box td-cta-docs" href="https://marketplace.ghost.org" target="_blank" rel="noopener">
```
vikaspotluri123: I think this should link to https://docs.ghost.org/api/handlebars-themes/

PR #1087, opened 2018-12-17
Removed `invitedBy` from signup routes

kirrg001:
refs https://github.com/TryGhost/Ghost/issues/10286

- not used
- looking for more information? visit https://github.com/TryGhost/Ghost/pull/10291


kirrg001:
updated 🤣

PR #1090, opened 2019-01-06
✨ updated importer to use new async endpoint

vikaspotluri123:
refs tryghost/ghost#9388
depends on tryghost/ghost#10340

This is my first time actually remotely working in an ember app, so I might have broken some standards! I'll definitely need a hand adding tests as needed 😄

### DIFF
--- a/app/controllers/settings/labs.js
+++ b/app/controllers/settings/labs.js
@@ -86,7 +86,7 @@ export default Controller.extend({
             let formData = new FormData();
             let notifications = this.notifications;
             let currentUserId = this.get('session.user.id');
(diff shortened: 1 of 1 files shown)


kevinansfield:
@vikaspotluri123 as a part of a more aggressive push to reduce the number of open issues that are not actively worked on the TryGhost/Ghost#9388 has been closed down. Closing this PR to keep "open PRs" tab clean. We definitely will come back to this once main issue is reopened/higher in priorities as the approach seems correct to me but needs more planning and broader team discussion as mentioned above 😉

PR #1099, opened 2019-01-28
Removed redundant 'status=all'  in user queries

gargol:
refs #5947

There doesn't seem to be a need to send this status in use requests as it's being handled on the server side. @kevinansfield would like to check if there could be some more simplifications done to this adapter after removing the status parameter? From what I've tested, seems like find/findAll methods can be safely removed as well?


kevinansfield:
@gargol if the v2 admin API always includes all status' then the `find`, `findRecord`, and `findAll` methods can all be deleted 😄

gargol:
awesome @kevinansfield! It's both v0.1 and v2 that to that now because default/enforced filters are parts of a shared `User` model (referenced in commit message). My understanding whole adapter can be reduced to single:
```
export default ApplicationAdapter.extend(SlugUrl);
```
right?

gargol:
Looks like it works. And the syntax for the adapter is not exactly the same as for tag :+1: Anything more I could check before merging this into master? :smiley:

kevinansfield:
The `queryRecord` method is still needed for the special handling of the `/users/me/` route.

PR #1103, opened 2019-02-22
Drop uploads in admin api

gargol:
refs https://github.com/TryGhost/Ghost/issues/10438

@kevinansfield changed `/uploads` endpoints to `/images` and updated response handling for those endpoints to use new structure (`{url: 'image_url'}`). Would love to receive feedback on possible other places that might be affected by the response structure change


kevinansfield:
@gargol there are two places we upload images within controllers, otherwise this looks 👍 

- https://github.com/TryGhost/Ghost-Admin/blob/master/app/controllers/setup/two.js#L92-L97
- https://github.com/TryGhost/Ghost-Admin/blob/master/app/controllers/signup.js#L107-L112

PR #1109, opened 2019-03-06
🐛 Post -> Page in Page PSM and delete modal

vikaspotluri123:
Closes tryghost/ghost#10577

- Update templates to use `post.displayName` rather than hard-coded `Post`

Fixed three separate things:
 - Code injection
 - Delete ... action button
 - Delete ... modal


kevinansfield:
👍 great, thank you!

PR #1110, opened 2019-03-07
Prototype admin<>site view

JohnONolan:
Add inline site view to Ghost Admin


JohnONolan:
Was this just waiting because it was in draft mode? or does it need something else?

kevinansfield:
> Was this just waiting because it was in draft mode?

Yeah, I hadn't seen/heard anything about this being more than a draft prototype. If we're happy with the design and we can iterate on functionality later then 👍

peterzimon:
I would like to fix, refine a couple of things before we release. I'll work on it early this week.

PR #1118, opened 2019-03-13
Update Renovate Configuration

kevinansfield:
no issue
- switch scheduling to early Monday mornings


kevinansfield:
Re-opening as Renovate doesn't appear to be picking up the changes from a force-push

PR #1131, opened 2019-03-25
🐛 Fixed the scroll problem using safari on low resolutions smartphones

guisouzabr:
closes [#9494 issue](https://github.com/TryGhost/Ghost/issues/9494) 
-Modified the content_list css class

It's my first PR here. You can find the use-case in posts list (admin panel), where the problem was reported in the issue.

### DIFF
--- a/app/styles/layouts/content.css
+++ b/app/styles/layouts/content.css
@@ -105,7 +105,6 @@
 
 .content-list {
     position: relative;
-    overflow-y: auto;
 }
 
 .content-list ol {


kevinansfield:
@guisouzabr have you tested this with enough posts to trigger infinite scroll? I believe the scroll behaviour needs to be fixed on `.gh-main` for infinite scroll to work.

guisouzabr:
> @guisouzabr have you tested this with enough posts to trigger infinite scroll? I believe the scroll behaviour needs to be fixed on `.gh-main` for infinite scroll to work.

hi @kevinansfield, how many posts i need to create to trigger the infinite scroll? I tried with 31 posts and everything worked fine.
Thanks!

PR #1139, opened 2019-04-02
Update Renovate Configuration

ErisDS:
I believe this config preset is a slightly better version of the longform that was there, but feel free to ignore this if I'm wrong :)

- simplify automerge rules using `:automergePatch` config preset
- See: https://renovatebot.com/docs/presets-default/#automergepatch

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -4,7 +4,8 @@
     "group:jsTestMonMajor",
     "group:linters",
     ":maintainLockFilesWeekly",
-    "schedule:earlyMondays"
+    "schedule:earlyMondays",
+    ":autoMergePatch"
   ],
   "ignoreDeps": [
     "ember-drag-drop",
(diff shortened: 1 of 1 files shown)


kevinansfield:
It looks almost the same but it will also automerge the lockfile maintenance PR 🤔 I was trying to leave that one as the last PR that is merged after any minors etc are manually merged so that it gets updated automatically.

My reason for merging that last was because the lockfile maintenance PR cleans up any duplicated packages which can be a bit of a pain in client-side projects depending on how the bundler deals with them. The normal `yarn add -D ...` package update command doesn't revisit other dependencies which is where the duplicates in yarn.lock occur.

ErisDS:
Your repo your choice :D

I think renovate is smart about the order of things: https://github.com/TryGhost/Ghost-SDK/pull/65

I am aiming to be optimistic in most repos and use `automergeMinor` until it's proven to be too optimistic 😈

kevinansfield:
Closing this as I'd prefer to keep the lock file maintenance PR separate for now. I'm also trying to figure out what Renovate is actually doing with it as it seems to have got stuck after it closed it's previous maintenance PR 😕

PR #1140, opened 2019-04-03
🐛 Fixed routes uploader validation allowing .yml files

vikaspotluri123:
no issue

- The server never supported uploading yml files


kevinansfield:
👍 Thanks @vikaspotluri123!

PR #1141, opened 2019-04-05
Made date input manual editable

Leon0824:
closes #9256

Remove the `readonly` attribute.


kevinansfield:
Hey @Leon0824 👋 Thanks for taking a look at this issue, however removing the `readonly` attribute makes it look like the field is editable but it's not, there's no two-way data binding between the application state and the field value so manually typing a value in won't affect anything.

The approach to this needs to be setting up an action on the input so that the application state is properly updated and to add error handling for incorrect format/syntax and invalid dates. As such I'm going to close this PR for now, please open a new one with a full implementation if you want to tackle it 🙂

PR #1142, opened 2019-04-05
Gscan styles

peterzimon:
refs. https://github.com/TryGhost/Team/issues/206

Improvements for theme upload and activation dialogs:
- updated copy for buttons and static descriptions
- updated style for better readability of gscan errors
- fixed modal scrolling when gscan output is too long to fit on screen

### DIFF
--- a/app/styles/layouts/settings.css
+++ b/app/styles/layouts/settings.css
@@ -598,7 +598,7 @@
 
 /*Errors */
 .theme-validation-container {
-    max-height: calc(100vh - 12vw - 140px);
+    max-height: calc(100vh - 12vw - 110px);
     overflow-y: scroll;
     margin: -40px -40px 0;
     padding: 40px 40px 0;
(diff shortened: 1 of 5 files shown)


app/styles/layouts/settings.css
```diff
@@ -597,59 +597,109 @@
...
+.theme-validation-item {
+    background: var(--whitegrey-l2);
+}
+
 .theme-validation-toggle-details {
     display: flex;
     justify-content: space-between;
     flex-grow: 1;
-    align-items: center;
+    align-items: start;
```
kevinansfield: `autoprefixer` is complaining about this because it has "mixed support" and recommends `flex-start` instead. Looking at [the flexbox spec](https://www.w3.org/TR/css-flexbox-1/#align-items-property), `start` isn't listed as one of the possible values.

```suggestion
    align-items: flex-start;
```

PR #1143, opened 2019-04-05
Fixed post/page labels in feature image and save components

vikaspotluri123:
Closes TryGhost/Ghost#10658

- Update publish menu states to use `{{post.displayName}}` instead of hard coded `post`
- Update settings menu image uploader to generate CTA using `post.displayName`


kevinansfield:
👍 thanks @vikaspotluri123!

PR #1150, opened 2019-04-11
Updated member management section

rishabhgrg:
- Updated members model to add subscriptions
- Added subscription transforms
- Updated member screen to show real data from API

### DIFF
--- a/app/models/member.js
+++ b/app/models/member.js
@@ -4,5 +4,6 @@ import attr from 'ember-data/attr';
 export default DS.Model.extend({
     name: attr('string'),
     email: attr('string'),
-    createdAt: attr('moment-utc')
+    createdAt: attr('moment-utc'),
+    subscriptions: attr('member-subscription')
 });
--- a/app/routes/member.js
+++ b/app/routes/member.js
@@ -1,7 +1,26 @@
 import AuthenticatedRoute from 'ghost-admin/routes/authenticated';
 
 export default AuthenticatedRoute.extend({
-    // TODO: add model method to load member if not passed in
+    model(params) {
+        this._isMemberUpdated = true;
+        return this.store.findRecord('member', params.member_id, {
+            reload: true
+        });
+    },
+
+    setupController(controller, model) {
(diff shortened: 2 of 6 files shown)


app/models/member.js
```diff
@@ -4,5 +4,6 @@ import attr from 'ember-data/attr';
 export default DS.Model.extend({
     name: attr('string'),
     email: attr('string'),
-    createdAt: attr('moment-utc')
+    createdAt: attr('moment-utc'),
+    subscriptions: attr('member-subscription')
```
rishabhgrg: @kevinansfield Would be cool if you can take a 👀 if this is the best way to add `subscriptions` info on member model. This uses the transform + separate model definition way which seemed to be the right way from other references in Admin 😄

kevinansfield: 👍 looks fine to me

app/routes/member.js
```diff
@@ -1,7 +1,26 @@
 import AuthenticatedRoute from 'ghost-admin/routes/authenticated';
 
 export default AuthenticatedRoute.extend({
-    // TODO: add model method to load member if not passed in
+    model(params) {
+        this.set('isMemberUpdated', true);
```
kevinansfield: This is more of an FYI. Generally we try to limit use of `.set()` to when a property is going to be used in a template or another computed property where tracking is needed. In this case it's completely internal to the route instance so `this.isMemberUpdated = true` would be preferable I think, or even `this._isMemberUpdated = true` to indicate that this is a "private" property.

The more we can standardise things now the easier it will be to switch to newer Ember/esnext idioms such as [tracked properties](https://blog.emberjs.com/2019/02/26/coming-soon-in-ember-octane-part-3.html) and real [private properties on classes](https://github.com/tc39/proposal-class-fields#private-fields).

rishabhgrg: 💯 agree, that sounds pretty sane to me. Its also good to know the right usage or best practices wherever possible in PRs, so please keep them coming 😄 I'll update here as well asap

PR #1162, opened 2019-04-17
Updated .travis.yml swap renovate build to push

ErisDS:
Fundamental problem: cos renovate pushes branches to the main repo, when it opens a PR there are 2 checks, one for push and one for the pr.

Currently to fix that we exclude renovate branches, and only build renovate PRs
That means we can't use automerge branch, as we wouldn't run any checks!
This swaps to running only renovate pushes, which means the branch always gets checked

PR builds check the merge commit, but as renovate always rebases from latest master, this is effectively the same commit as the push.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,9 +7,8 @@ env:
   global:
   - MOZ_HEADLESS=1
   - JOBS=1
-branches:
-  except:
-  - "/^renovate\\/.+$/"
+# Don't run builds for renovate PRs, only renovate pushes
+if: NOT branch =~ /^renovate\\/.+$/ AND type != pull_request
 addons:
   firefox: latest
   chrome: stable


ErisDS:
@kevinansfield making this a PR just so that it goes past your eyeballs as it's quite a weird change!

ErisDS:
@kevinansfield to clarify what this is for:

Currently my inbox gets destroyed every Monday (I'm sure yours does to) and that's not how I want life with Renovate to be. There are certain dependencies - for example mocha, that we just do not ned any sort of notification about when there's a patch or even minor release, providing tests pass, let alone getting 1 email per repo! 

Renovate has a mode where it will open a branch, let the checks run, and then if they pass merge instantly without making a PR. Meaning no-noise.

This is obviously not suitable for all deps, but linting and testing tools are great candidates for it I think. It may also be suitable for "well known" stable packages like express, lodash, etc.

ErisDS:
Updated with corrected config

kevinansfield:
> P.S. been meaning to ask - how come the scripts in .travis.yml are using npm rather than yarn?

No particular reason that I know of, just one of those things that's been overlooked I think.

PR #1194, opened 2019-05-09
Auto set video thumbnail to feature image (if feature image not set)

tranduongms1:
Auto set video thumbnail to feature image (if feature image not set)

When user insert of video it thumbnail will auto set to feature image of post (if it not set)

### DIFF
--- a/app/controllers/editor.js
+++ b/app/controllers/editor.js
@@ -149,6 +149,18 @@ export default Controller.extend({
         updateScratch(mobiledoc) {
             this.set('post.scratch', mobiledoc);
 
+            let postFeatureImage = this.get('post.feature_image');
+            if (!postFeatureImage) {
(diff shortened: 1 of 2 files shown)


kevinansfield:
Hey @tranduongms1 👋 We appreciate you jumping in here and solving pain points that you are encountering whilst using Ghost!

Unfortunately after discussing this internally it's not a feature we'd like to implement in core at this stage. The implementation in this PR is very specific to your use-case and doesn't scale well to general Ghost users, eg. it only works with video embeds (if the provider exposes the right data) rather than with any embed/card, it's client-side only and not retrospective so the behaviour is hit-and-miss and not particularly discoverable, and there's no way to _not_ have a featured image if you have a video embed.

As such I'm going to close this PR for now but thank you for giving us something to think about!

PR #1211, opened 2019-06-01
🎨Added 'add page' button

paulperkins:
no issue
Added 'add page' button to match 'add post' button
I've been looking to learn Ghost, Ember, and node.js, and this little addition seemed like a good way to start!


kevinansfield:
Hey @paulperkins 👋 Thanks for taking the time out to make s PR, however in this case the existing design is very much intentional so we won’t be making the change.

For any design changes it’s always worth asking on the forum first, or if you find a bug then opening an issue, especially if the change may involve a bit of work 🙂

PR #1212, opened 2019-06-02
🐛Changed date-time picker date format to 'YYYY-MM-DD'.

paulperkins:
refs #10767
In the absence of being able to determine full locale data (i.e. being able to differntiate between EN-US and EN-AU, EN-GB, etc.), using YYYY-MM-DD seems to be the most unambiguous date format. For those not in American English speaking locales, the American date format (MM/DD/YYYY) can be very confusing.


paulperkins:
Many apologies - completely neglected to run tests prior to submission. Closing PR.

PR #1213, opened 2019-06-02
🎨 Changed post/page date picker format to 'YYYY-MM-DD'

paulperkins:
refs TryGhost/Ghost#10767

The use of the date format 'MM/DD/YYYY' is confusing to those not in American English locations. In the absence of being able to dynamically determine full locale data, using 'YYYY-MM-DD' seems to the least ambiguous date format which would be intuitive to most users.

### DIFF
--- a/app/templates/components/gh-date-time-picker.hbs
+++ b/app/templates/components/gh-date-time-picker.hbs
@@ -8,7 +8,7 @@
     }}
         {{#dp.trigger tabindex="-1" data-test-date-time-picker-datepicker=true}}
             <div class="gh-date-time-picker-date {{if dateError "error"}}">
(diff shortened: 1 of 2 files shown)


seanhamlin:
+1 for [ISO 8601](https://en.wikipedia.org/wiki/ISO_8601). As a user not from the US, I would very much appreciate this change.

seanhamlin:
Great stuff, thanks @kevinansfield and @paulperkins !

paulperkins:
Thanks, @kevinansfield! And thanks to @seanhamlin for the supportive words :)

JohnONolan:
Thanks both of you for this change! ISO 8601 definitely feels correct and what we should be using for an international audience :)

PR #1235, opened 2019-06-22
Renamed "Preview site" to "View site"

allouis:
no-issue

Preview implies something which is not-yet-live, but this feature only shows the live site.

### DIFF
--- a/app/templates/components/gh-nav-menu.hbs
+++ b/app/templates/components/gh-nav-menu.hbs
@@ -20,7 +20,7 @@
             <li class="relative">
                 <span {{action "transitionToOrRefreshSite" on="click"}}>
                     {{#link-to "site" data-test-nav="site" current-when=isOnSite preventDefault=false}}
-                        {{svg-jar "house"}} Preview site
+                        {{svg-jar "house"}} View site
                     {{/link-to}}
                 </span>
                 <a href="{{config.blogUrl}}/" class="gh-secondary-action" title="Open site in new tab" target="_blank">


ndbroadbent:
Hi @allouis! I just found this PR while I was searching for a way to configure a custom "preview URL" for the "View site" iframe. ([I've also posted on the Ghost forums about my configuration](https://forum.ghost.org/t/overwrite-x-frame-option-on-ghost-site/12794/3?u=ndbroadbent).)

I've set up Ghost on Digital Ocean, and I've put it behind a proxy on my root domain (via my web application server.) Everything on my root domain is also fully cached with Cloudflare. 

 In `config.production.json`, I configured the `"url"` property to be the root domain, so that Ghost would generate all of the correct URLs. But I also disabled access to the `/blog/ghost` path on my root domain, so I am only able to access the Ghost admin pages on a subdomain (which isn't behind Cloudflare.) This means I log in to `https://<subdomain>.<root domain>/blog/ghost` to update posts.

The default `X-Frame-Options: SAMEORIGIN` header from Nginx was causing some problems with the "View site" feature, because Ghost Admin is trying to load the site inside the iframe, but it's on a different domain.

I got it working by removing the `X-Frame-Options` header, but now the problem is that the root domain is fully cached behind Cloudflare, so if I want to see what any updates look like on the blog, I have to manually purge the Cloudflare cache. (And I'm also making updates to the live site.)

So it would be really awesome if I could configure a `"preview_url"` setting in `config.production.json`, which would just be the un-cached subdomain. Then Ghost Admin would show the site from the same subdomain in the iframe, and it wouldn't be cached. 

Another benefit is that Cloudflare's caching would also give me a rudimentary "site preview" feature. But it would mean that I could preview live changes to my site before I "commit" them by purging the Cloudflare cache. (Although this wouldn't work perfectly, since Cloudflare can request an uncached page at any time.)

Anyway, sorry for the long comment! I just saw that you are a Ghost developer, so I was wondering if this is a use-case that could be supported? 

I know that Ghost(Pro) offers a managed subdirectory install with support for proxying, so I was wondering how the site preview is achieved for these customers?

kevinansfield:
@ndbroadbent please keep feature requests on the [Ideas category](https://forum.ghost.org/c/Ideas/5) of the forum, thanks! 🙂

PR #1261, opened 2019-07-09
🎨 Enable Manual Typing of Date in Post Settings Menu

marclafountain:
refs [#9256](https://github.com/TryGhost/Ghost/issues/9256)

- Removed readonly parameter so date field is editable
- Added onchange action to set date when date field is changed

I'm brand new to Javascript, Node, and Git. So it's very possible that I have missed something here. If so, feedback is welcome.

### DIFF
--- a/app/templates/components/gh-date-time-picker.hbs
+++ b/app/templates/components/gh-date-time-picker.hbs
@@ -8,7 +8,7 @@
     }}
         {{#dp.trigger tabindex="-1" data-test-date-time-picker-datepicker=true}}
             <div class="gh-date-time-picker-date {{if dateError "error"}}">
-                <input type="text" readonly value={{moment-format _date "YYYY-MM-DD"}} disabled={{disabled}} data-test-date-time-picker-date-input>
+                <input type="text" value={{moment-format _date "YYYY-MM-DD"}} disabled={{disabled}} onchange={{action "setDate" value="target.value"}} data-test-date-time-picker-date-input>
                 {{svg-jar "calendar"}}
             </div>
         {{/dp.trigger}}
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "ghost-admin",
-  "version": "2.25.4",
+  "version": "2.25.5",
   "description": "Ember.js admin client for Ghost",
   "author": "Ghost Foundation",
   "homepage": "http://ghost.org",
--- a/package.json
+++ b/package.json
@@ -84,7 +84,7 @@
     "ember-one-way-select": "4.0.0",
     "ember-power-calendar-moment": "0.1.7",
     "ember-power-datepicker": "0.6.0",
-    "ember-power-select": "2.3.4",
+    "ember-power-select": "2.3.5",
     "ember-resolver": "5.1.3",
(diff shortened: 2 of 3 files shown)


kevinansfield:
@marclafountain unfortunately this is not as simple to change as it first looks which is why it hasn't been tackled so far and the input is marked as readonly. The code in this PR is currently setting an invalid or unexpected date on every keypress until a (hopefully) valid date is typed in a format that JS is able to understand.

A full solution to allowing typed date entry would require restricting the input to a known format and correctly parsing from that into a valid date object but only once the date entry has been finished (we'd need some design input to decide when that is - when the expected format is complete, on blur, on enter, etc). We'd also need to implement appropriate error handling/messaging as well as a reset ability for when invalid input needs to be removed.

Thank you for taking the time out to look at this 🙌 but I'm going to close the PR for now because it's too far away from a working solution.

marclafountain:
@kevinmansfield I used onchange rather than oninput in the hopes that date validation would only happen once after the full date has been entered (when someone hits Return or Tab in the date field, clicks / moves outside the date field, leaves the PSM, etc.)

In my testing that seemed to work as planned and date validation appeared to only happen one time per attempt due to cases like those above. Valid dates were saved and invalid dates got an error message once. Apologies, I really thought this was working correctly. Very bummed!

marclafountain:
@kevinansfield I'd like to try and dig deeper on this. Could you share a bit more information? 

Are you saying the code is trying to set the date with every keypress even though I am using "onchange"? That doesn't make sense to me because I can put in a valid date and it doesn't get saved on the post until I trigger an "onchange" event (hit return or tab, click outside the date field, leave the PSM, etc.). If I put in a valid date and don't trigger an "onchange" event (by say reloading the browser tab), no date change is saved.

You also mention adding date validation and error handling, but it appears to me the ember-power-datepicker already has those things built in (at least to some degree). The code rejects attempts to set invalid dates when I trigger an "onchange" event by showing a brief error message and putting in the current date.

I made a screen cap showing what I'm talking about [here](https://www.dropbox.com/s/wqqe29g1iufu0rr/date_test.mov?dl=0). I hit Return to trigger the "onchange" events in the video.

Thanks for any further info you can share!

kevinansfield:
Re-opening for further review

kevinansfield:
I've reviewed this again, although the oninput/onchange was a mixup in my first reply, unfortunately this change is still too shallow to work properly.

1. The lack of error handling means that the app can get into a broken state if an invalid date is entered into the "schedule post" modal
2. The "auto reset to current date" behaviour in the post settings menu when an invalid date is entered is the result of an un-intended side effect somewhere in the data handling layer rather than being properly codified
    1. this also has the unfortunate effect of resetting and saving the date, even for published posts. If you already had a published date/time and you was accidentalally or unintentionally enter an invalid date you've now lost your original published date, your public page+sitemap is updated with the wrong date, and you have to remember and manually re-enter the correct date/time

As such, I'm going to close this PR again. The original issue is still in our backlog with high priority so we'll get to the full design/spec/implementation of the feature.

marclafountain:
Understood. Thanks for taking a second look and sharing more info. Much appreciated!

PR #1283, opened 2019-07-30
Tags screen refresh

rishabhgrg:
Updates tags list and tag detail screens

- Adds new tag list component for list page following same design pattern as posts/page - `gh-tags-list-item`
- Adds tab button/view for internal/public tags
- Adds new tag detail screen for editing/creating tags
- Creating new tags and editing existing now needs to be explicitly saved
- Added slug generation based on name for new tag using utils


rishabhgrg:
Build fails currently as tests have not been updated to work with new tag routing/structure

app/routes/tags/index.js
```diff
@@ -1,16 +1,7 @@
...
     beforeModel() {
-        let firstTag = this.modelFor('tags').get('firstObject');
-
         this._super(...arguments);
-
-        if (firstTag && !this.get('mediaQueries.maxWidth600')) {
-            this.transitionTo('tags.tag', firstTag);
-        }
     }
 });
```
rishabhgrg: It seems like this file can just be removed at this state as default route will kickin, but kept it for now till the final changes are complete 😬

tests/acceptance/settings/tags-test.js
```diff
@@ -38,7 +38,7 @@ let keyup = function (code, el) {
     (el || document).dispatchEvent(event);
 };
 
-describe('Acceptance: Tags', function () {
+describe.skip('Acceptance: Tags', function () {
```
rishabhgrg: The acceptance here covers the old behavior and are in WIP for new behavior, so skipped them here and for the `tag-settings` component and will be replaced with updated tests

app/routes/tags/index.js
kevinansfield: The empty `beforeModel()` can be removed but we still want this to be an `AuthenticatedRoute` instance rather than the default auto-generated `Route`

kevinansfield: Ignore the above, missed that this was the index and not the `tags` route 🙈 The file can be removed 👍

app/templates/tags.hbs
```diff
@@ -1,28 +1,44 @@
...
+                <button class="gh-btn {{if (eq type "internal") "gh-btn-group-selected"}}" {{action "changeType" "internal"}}><span>Internal tags</span></button>
+            </div>
             {{#link-to "tags.new" class="gh-btn gh-btn-green"}}<span>New tag</span>{{/link-to}}
         </section>
     </header>
 
-    {{#gh-tags-management-container tags=tags selectedTag=selectedTag enteredMobile="enteredMobile" leftMobile=(action "leftMobile") as |container|}}
-    <div class="tag-list">
-        <section class="tag-list-content settings-tags {{if tagListFocused 'keyboard-focused'}}">
-            {{#vertical-collection sortedTags
```
kevinansfield: We need to keep the usage of `{{#vertical-collection}}`, there are too many users with enough tags that the screen will lag really badly without it.

app/components/gh-tags-list-item.js
```diff
@@ -0,0 +1,80 @@
...
+    click() {
+        this.onClick(this.tag);
+    },
+
+    doubleClick() {
+        this.onDoubleClick(this.tag);
+    },
+
+    scrollIntoView() {
+        let element = this.$();
```
kevinansfield: We shouldn't be using `this.$()` or `$()` anywhere any more - jQuery is deprecated in Ember and we're trying to move our code away from it so that we can drop the dependency. You can use `this.element` to get the native DOM element of the component.

app/controllers/tags.js
```diff
@@ -1,20 +1,24 @@
...
-import {run} from '@ember/runloop';
 
 export default Controller.extend({
 
     tagController: controller('tags.tag'),
 
+    queryParams: ['type'],
+    type: 'public',
+
+    init() {
```
kevinansfield: The `init()` method override isn't needed if there is no logic in there.

app/components/gh-tags-list-item.js
rishabhgrg: Since we don't need `scrollIntoView` anymore for tags list screen, this is being removed.

PR #1286, opened 2019-08-08
🎨 Added auto-login to private site when viewing site preview in admin

kevinansfield:
closes https://github.com/TryGhost/Ghost/issues/10995

- when first loading the site preview, if private mode is enabled submit the login form in the background to retrieve the session cookie before loading the iframe
- refactors post-authentication preloading to ensure it occurs before post-authentication route hooks are called
- adds `showSuccess` attribute to `<GhTaskButton>` so that when set to `false` it can stay in the running state after "success" to avoid state change flashes whilst waiting for a transition

### DIFF
--- a/app/authenticators/cookie.js
+++ b/app/authenticators/cookie.js
@@ -5,7 +5,11 @@ import {inject as service} from '@ember/service';
 
 export default Authenticator.extend({
     ajax: service(),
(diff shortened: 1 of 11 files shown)


JohnONolan:
👌 👌 👌

PR #1287, opened 2019-08-08
✨ Added global meta data and Twitter/Facebook card settings

kevinansfield:
closes https://github.com/TryGhost/Ghost/issues/10921

- added new fields to settings model
- added "Site SEO" section to general settings

@peterzimon ready for you to work your design magic when you're back 😄 

@JohnONolan a wording review would be good, so far it's only been copied across from a rough sketch so it could do with some tweaking - do we want to mention the fallback behaviour anywhere?

Screenshot 2019-08-08 at 16 24 58

### DIFF
--- a/app/models/setting.js
+++ b/app/models/setting.js
@@ -29,5 +29,13 @@ export default Model.extend(ValidationEngine, {
             return {isActive: true};
         }
     }),
-    membersSubscriptionSettings: attr('string')
+    membersSubscriptionSettings: attr('string'),
+    metaTitle: attr('string'),
+    metaDescription: attr('string'),
+    twitterTitle: attr('string'),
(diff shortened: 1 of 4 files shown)


JohnONolan:
Is there a reason for using new labels and descriptions rather than the existing naming from the same feature in the PSM / tags / elsewhere? (In particular: `SEO` should be `Meta`)

Should definitely be "Site" rather than "Home" as the prefix, because this isn't just the homepage, it's the homepage + the global default(s)

---

Could also do with a small review of casing consistency:

- PSM is `Twitter Title` but this is `Twitter title`
- General settings all use sentence case but `Site Language` uses title case

IIRC we settled on sentence case unless acronym or proper noun

PR #1330, opened 2019-09-18
🔥 Removed unnecessary '?status=all' query parameter

gargol:
no issue

- The default post/user filter gets expanded to the same values default values as `status=all` used to: https://github.com/TryGhost/Ghost/blob/e57e19e/core/server/api/canary/utils/serializers/input/posts.js#L97

@kevinansfield didn't find any more references to this filter seems like quite a small change :smiley:


kevinansfield:
Cool, the only other place I saw with a quick look was the mirage users endpoint here https://github.com/TryGhost/Ghost-Admin/blob/master/mirage/config/users.js#L21-L35. Seeing as we're not using the status param any more it might be good to remove that whole code block to avoid any confusion 🙂

PR #1337, opened 2019-09-24
🔥 Removed all subscriber feature related code

gargol:
refs https://github.com/TryGhost/Ghost/pull/11153

@rishabhgrg would be good a second pair to check if I haven't removed something accidentally (tested Admin UI and things look good)
@kevinansfield just FYI if you also wanna have a quick look :)


mirage/fixtures/configs.js
```diff
@@ -3,7 +3,6 @@ export default [{
     database: 'mysql',
     enableDeveloperExperiments: true,
     environment: 'development',
-    labs: {subscribers: false},
```
kevinansfield: Can we keep this line but change it to `labs: {}` instead? Labs itself isn't disappearing, just the subscribers feature.

mirage/fixtures/settings.js
```diff
@@ -70,16 +70,6 @@ export default [
         updated_by: 1,
         value: ''
     },
-    {
-        id: 12,
-        key: 'labs',
-        value: '{"subscribers":true}',
```
kevinansfield: Similar to the above, put this setting back but keep the value as `'{}'`

PR #1346, opened 2019-10-01
Content visibility configuration in labs

gargol:
~is :no_entry_sign: by https://github.com/TryGhost/Ghost/pull/11183~

Changes:
- [x] Add UI for default content visibility setting
- [x] Moves default posts visibility setting to server-side

### DIFF
--- a/app/components/gh-psm-visibility-input.js
+++ b/app/components/gh-psm-visibility-input.js
@@ -1,4 +1,6 @@
 import Component from '@ember/component';
+import {computed} from '@ember/object';
+import {inject as service} from '@ember/service';
 
 const VISIBILITIES = [
     {label: 'Everyone', name: 'public'},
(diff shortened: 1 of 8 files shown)


app/components/gh-psm-visibility-input.js
```diff
@@ -8,9 +10,15 @@ const VISIBILITIES = [
 
 export default Component.extend({
 
+    settings: service(),
+
     // public attrs
     post: null,
 
+    selectedVisibility: computed('post', function () {
+        return this.post.get('visibility') || this.settings.get('defaultContentVisibility');
```
gargol: @kevinansfield wdyt about such approach to setting the default when PSM is opened first thing when creating a new post? Is there a cleaner way?

gargol:
Broke my own tests by the last commit will fix first thing in the morning :roll_eyes:

PR #1367, opened 2019-10-16
Improved application loading state

JohnONolan:
Ghost-Admin loading state currently feels really jerky, so made a 5 minute update which I think makes it feel a lot smoother on first load:

Before:
image

After:
image

Before:
load-before

After:
load-after

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -37,6 +37,30 @@
 </head>
 <body>
 
+<style>
+.ember-load-indicator {
+    position: fixed;
+    top: 0;
+    right: 0;
+    bottom: 0;
+    left: 0;
+    background: #f2f4f7;
+}
+@media (min-width: 800px) {
+    .ember-load-indicator .gh-loading-content { padding-left: 267px; }
+    .ember-load-indicator:before {
+        content: "";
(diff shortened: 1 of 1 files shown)


kevinansfield:
@JohnONolan do we care about the loading but when not logged in/set up state? That will be more janky after this change because there will be a sidebar that then disappears

JohnONolan:
I forgot about this PR 😬 

@kevinansfield can you think of any reasonable way, with media queries or otherwise, to serve slightly adjusted styles for this on login/signup?

kevinansfield:
The only route I can think of is to change the `/ghost/` endpoint that renders the index.html so that it's aware of the authentication cookie and can modify the served html for logged-in/out states.

JohnONolan:
@ErisDS Any reasonable way for Ghost inject a `<body>` class everywhere based on route? Pretty sure I could CSS it if I could have access to things like `.gh-admin-auth` (login/signup/passreset) / `.gh-admin-setup` (install/setup ghost)

ErisDS:
Closing this as we are prepping to move admin into the monorepo.

PR #1376, opened 2019-10-21
Refined copy for members settings

allouis:
fa2fe870c1e2e34d1436808f11ebd5b0c5173274

### DIFF
--- a/app/templates/settings/labs.hbs
+++ b/app/templates/settings/labs.hbs
@@ -15,21 +15,19 @@
                 <div class="gh-members-setting-content">
                     <div class="flex">
                         <div class="flex flex-column flex-grow-1">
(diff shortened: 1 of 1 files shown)


app/templates/components/gh-members-lab-setting.hbs
```diff
@@ -31,10 +31,9 @@
                             input=(action "setSubscriptionSettings" "secret_token")
                             class="mt1 password"
                         }}
-                        <a href="https://dashboard.stripe.com/account/apikeys" target="_blank" class="mt1 fw4 f8">
-                            Find your Stripe API keys here &raquo;
-                        </a>
-                    </div>
```
allouis: Looks like this is missing incorrectly

PR #1408, opened 2019-11-18
Allowed to toggle send email when batch has an error

gargol:
Still TODO:
- [ ] wording check for a failed email message in PSM


gargol:
@peterzimon just fyi, styling is bit off for this change (everything is dotted `...`). So might need to take a look at some CSS magic along with wording when you are onto it :) 

Screenshot from 2019-11-18 21-34-03

kevinansfield:
Closing as this has been replaced with a direct retry method rather than unpublish/republish.

PR #1409, opened 2019-11-20
✨ Added new brand color publication setting

rishabhgrg:
Depends on - https://github.com/TryGhost/Ghost/pull/11408

- Reads new brand primary color setting and adds UI to modify/update it behind the dev flag


rishabhgrg:
@kevinansfield Nothing urgent but will be great to have your 👀 on this for general pattern/approach of input :)

PR #1410, opened 2019-11-20
✨ Added Secondary Navigation

ErisDS:
refs: https://github.com/TryGhost/Ghost/pull/11409

image

- Added a new UI for a second set of navigation links
- This should support most concepts of nav, e.g. left and right, or header and footer
- This PR mostly updates the design and nav components to cope with a second set of nav

@kevinansfield 👋 shoving this up now that it seems to be 100% working to get feedback on approach - ember is not my usual bag - mostly not sure about true/falsy values in this context or whether it should be passing the model. Totes aware I need to write tests 😬will get right on that.

### DIFF
--- a/app/controllers/settings/design.js
+++ b/app/controllers/settings/design.js
@@ -19,12 +19,14 @@ export default Controller.extend({
 
     dirtyAttributes: false,
     newNavItem: null,
+    newSecondaryNavItem: null,
     themes: null,
     themeToDelete: null,
 
     init() {
         this._super(...arguments);
         this.set('newNavItem', NavigationItem.create({isNew: true}));
+        this.set('newSecondaryNavItem', NavigationItem.create({isNew: true}));
     },
 
     showDeleteThemeModal: notEmpty('themeToDelete'),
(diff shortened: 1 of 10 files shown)


ErisDS:
@kevinansfield this is totes done now - I'm interested, purely for academic reasons, in if there's a better way to do it (even if it's not important to change) i.e. some way to use models rather than randomly passing true sometimes.

app/controllers/settings/design.js
```diff
@@ -40,25 +42,31 @@ export default Controller.extend({
...
 
-            return newNavItem.validate().then(() => {
-                this.addNewNavItem();
+            return itemToAdd.validate().then(() => {
+                this.addNewNavItem(secondary);
             });
         },
 
-        deleteNavItem(item) {
+        deleteNavItem(secondary, item) {
```
kevinansfield: `secondary` isn't strictly necessary here because the information should already be present on `item.isSecondary`

ErisDS: I tried that and the item didn't seem to be the model - maybe I missed something

ErisDS: Ah - I remember now.

`isSecondary` is only set on the "new" nav items. `isSecondary` is not actually used at all, I thought I'd removed it.

This was the approach that I thought was more correct, but I couldn't figure out how to make it so that the models automatically had isSecondary applied (during the transform?), when they were pulled from the API. 

So what happened was, this approach worked fine for the new items, but when deleting an existing item that wasn't created during the current session, it didn't work because isSecondary was always false.

ErisDS:
I removed the unused `isSecondary` - but now the tests are failing. This shouldn't have had an impact though :(

ErisDS:
Ok - I think I've got this nailed. I've kept commits separate temporarily to show progression.

The first extra commit is me fixing issues the "old" way, and the last commit shows the refactor to rely on the model always. I've made quite a lot of changes, although a lot of them are code removal/simplification. 

Hopefully, I've gotten to grips here with the ember-way, but very open to feedback, especially around the addNavItem code that now passes the item around.

PR #1411, opened 2019-11-21
✨ Added member "add" screen

gargol:
- Adds the ability to manually add a new member through `members/new` screen
- Adds basic validation to member form fields

On TODO menu:
- [x] make email optionally editable
- [x] put in basic validation for member model!
- [x] email + other field validation
- [x] tests


app/components/gh-member-avatar.js
```diff
@@ -21,7 +21,7 @@ export default Component.extend({
     }),
 
     backgroundStyle: computed('member.{name,email}', function () {
-        let name = this.member.name || this.member.email;
+        let name = this.member.name || this.member.email || 'NM';
```
gargol: @peterzimon came up with this 'NM' piece as a placeholder for New Member initials, without this the screen looks broken. It changes to normal initials calculation once email or name is entered. Lmk if you'd like to put something different here ;)

gargol:
@peterzimon one of scenarios that I'd want to clarify is what happens when an admin enters some dummy email and clicks save? The only way to recover from this situation is to delete the member and start editing again. We are basically giving one chance to input an email for the member.

gargol:
@kevinansfield would need your expertise on the flaky test here :) The one that is failing is behaving in a strange way, when run in isolation on CI or locally it fails. But when I open up browser through `ember test -s -f 'Acceptance: Members' --launch=chrome` it always passes. Experimented with different configurations of `await wait()` and `await timeout(xxx)` with no luck :/ Maybe you can spot the problem easily?

PR #1417, opened 2019-11-26
🎨  Added gravatars for member avatars

gargol:
- When an email has a valid gravatar handle it displays an image instead of initials for the member

TODO:
- [x] styles

@rishabhgrg would love a second pair :eyes: on the Ember code. Thaaanks!

### DIFF
--- a/app/templates/components/gh-member-avatar.hbs
+++ b/app/templates/components/gh-member-avatar.hbs
@@ -1,5 +1,5 @@
-<figure class="gh-member-gravatar" ...attributes>
-    <div class="gh-member-initials flex items-center justify-center br-100" style={{this.backgroundStyle}}>
+<figure class="gh-member-gravatar {{this.containerClass}}">
+    <div class="gh-member-initials flex items-center justify-center br-100 {{this.containerClass}}" style={{this.backgroundStyle}}>
         <span class="gh-member-avatar-label {{this.initialsClass}}">{{this.initials}}</span>
     </div>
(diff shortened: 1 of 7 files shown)


gargol:
@peterzimon  Styling is pretty much done here but would need your opinion on the animation and general looks (both were taken almost 1:1 from user avatar styles with minor adjustments)

app/templates/components/gh-member-avatar.hbs
```diff
@@ -1,3 +1,8 @@
-<div class="flex items-center justify-center br-100" style={{this.backgroundStyle}} ...attributes>
-    <span class="gh-member-avatar-label {{this.initialsClass}}">{{this.initials}}</span>
-</div>
\ No newline at end of file
+<figure class="gh-member-gravatar" ...attributes>
+    <div class="gh-member-initials flex items-center justify-center br-100" style={{this.backgroundStyle}} ...attributes>
```
kevinansfield: It's generally advised to only use splattributes `(...attributes)` once inside a component template. It's intended for passing attributes specified in the component invocation onto an element (usually the outer element) inside a component. Eg.

`<GhMemberAvatar class="my-class" @member={{this.member}} />`

Would push `class="my-class"` onto every element in the component template that has `...attributes` specified. For class attributes that's maybe not so much of a problem, but if you add element modifiers it gets a bit more tricky, eg:

`<GhMemberAvatar class="my-class" @member={{this.member}} {{did-insert this.animate}} />`

Would call the `this.animate` method three times whereas it would only be expected to be called once (poor example for this specific component but hopefully you get the gist 😄).

PR #1428, opened 2019-12-16
Added host limit checks for sending email

rishabhgrg:
depends on - https://github.com/TryGhost/Ghost/pull/11439

Adds host limit upgrade popup in case publishing a post with email newsletter exceeds total number of members allowed by the plan. Handles new 403-HostLimitError by server

### DIFF
--- a/app/controllers/editor.js
+++ b/app/controllers/editor.js
@@ -387,8 +387,8 @@ export default Controller.extend({
 
             return post;
         } catch (error) {
-            // trigger upgrade modal if forbidden error
-            if (error && isForbiddenError(error) && error.payload && error.payload.errors && error.payload.errors[0].message) {
+            // trigger upgrade modal if forbidden(403) error
+            if (error && error.payload && error.payload.errors && error.payload.errors[0].message && isForbiddenError(error)) {
                 this.set('post.status', prevStatus);
                 this.set('hostLimitError', error.payload.errors[0]);
                 this.set('showUpgradeModal', true);
--- a/app/controllers/editor.js
+++ b/app/controllers/editor.js
@@ -388,7 +388,7 @@ export default Controller.extend({
             return post;
         } catch (error) {
             // trigger upgrade modal if forbidden(403) error
-            if (error && error.payload && error.payload.errors && error.payload.errors[0].message && isForbiddenError(error)) {
+            if (this._isHostLimitError(error)) {
                 this.set('post.status', prevStatus);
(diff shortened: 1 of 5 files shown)


app/controllers/editor.js
```diff
@@ -374,6 +383,14 @@ export default Controller.extend({
 
             return post;
         } catch (error) {
+            // trigger upgrade modal if forbidden error
+            if (error && isForbiddenError(error) && error.payload && error.payload.errors && error.payload.errors[0].message) {
```
rishabhgrg: @kevinansfield I am not 💯 if this is the right place to handle the `403` and trigger the upgrade popup, as the other place I was thinking it might be better is at the top level in application where ajax error handling is done [here](https://github.com/TryGhost/Ghost-Admin/blob/master/app/routes/application.js#L128) instead. 

The idea about putting in application was that it becomes app level generic handling to throw the popup whenever `403-HostLimitError` comes, but was not sure how that will fit with controller/template. Will be good interested in hearing your thoughts on how to approach this :)

gargol:
@rishabhgrg the popup with "unsaved changes" seems to be super aggressive when the limit is hit and you try navigating out of the post. Probably something to be looked into before releasing?

app/controllers/editor.js
```diff
@@ -374,6 +383,14 @@ export default Controller.extend({
 
             return post;
         } catch (error) {
+            // trigger upgrade modal if forbidden(403) error
+            if (this._isHostLimitError(error)) {
+                this.set('post.status', prevStatus);
```
rishabhgrg: @kevinansfield This is probably not the best way to reset post to previous state in case save fails due to host limit, as after upgrade popup the "Unsaved changes" popup is quite aggressive 😬 Do we have a standard way to rollback the post changes before attempting save like model rollback or something?

app/controllers/editor.js
kevinansfield: No need to do anything generic at all at the moment, let's not pre-optimise until we have other use cases :) Explicit handling in the editor controller's save task is 👌

app/controllers/editor.js
kevinansfield: The problem with the modal was the post being put into an error state by Ember Data due to the error occurring during a `.save()` call.

Changed the "reset" to `this.post.rollbackAttributes()`
- prevents unsaved changes modal from showing due to the post being in an error state
- all to-be-saved attributes are already in scratch values so the modal will still be shown correctly if there are real unsaved changes outside of the post status

PR #1430, opened 2019-12-18
Added membership plan currency selector  

gargol:
Changes:
- refactors variable naming to be more generic in regards to currencies
- adds a dropdown which controls setting plan currency on both monthly & yearly plans

TODO:
- [x] figure out the bug with dropdown select box, as it's not displaying selected value properly.

### DIFF
--- a/app/components/gh-members-lab-setting.js
+++ b/app/components/gh-members-lab-setting.js
@@ -150,7 +150,22 @@ export default Component.extend({
             }
 
             if (key === 'currency') {
-                stripeProcessor.config.plans.forEach(plan => (plan.currency = event.value));
+                stripeProcessor.config.plans.forEach((plan) => {
+                    if (plan.name !== 'Complimentary') {
+                        plan.currency = event.value;
+                    }
+                });
+
+                // NOTE: need to keep Complimentary plans with all available currencies so they don't conflict
+                //       when applied to members with existing subscriptions in different currencies (ref. https://stripe.com/docs/billing/customer#currency)
+                let currentCurrencyComplimentary = stripeProcessor.config.plans.filter(plan => (plan.currency === event.value && plan.name === 'Complimentary'));
+
+                if (!currentCurrencyComplimentary.length) {
+                    let complimentary = stripeProcessor.config.plans.find(plan => (plan.name === 'Complimentary'));
+                    let newComplimentary = Object.assign({}, complimentary, {currency: event.value});
+                    stripeProcessor.config.plans.push(newComplimentary);
+                }
+
                 stripeProcessor.config.currency = event.value;
             }
 
--- a/app/components/gh-members-lab-setting.js
+++ b/app/components/gh-members-lab-setting.js
@@ -7,10 +7,28 @@ import {set} from '@ember/object';
 const US = {flag: '🇺🇸', name: 'US', baseUrl: 'https://api.mailgun.net/v3'};
 const EU = {flag: '🇪🇺', name: 'EU', baseUrl: 'https://api.eu.mailgun.net/v3'};
 
+const currencies = [
+    {
+        label: 'USD', value: 'usd'
+    },
+    {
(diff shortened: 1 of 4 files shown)


app/components/gh-members-lab-setting.js
```diff
@@ -26,9 +26,14 @@ export default Component.extend({
     config: service(),
     mediaQueries: service(),
 
-    currencies: currencies,
+    currencies: null,
```
kevinansfield: Even though this is a constant its best practice to never use arrays/objects as default properties in an EmberObject because the values can end up being shared across instances and leaking state.

We have a lint rule to protect against it, not sure why it didn't pick up this case 🤔 https://github.com/ember-cli/eslint-plugin-ember/blob/master/docs/rules/avoid-leaking-state-in-ember-objects.md

app/components/gh-members-lab-setting.js
```diff
@@ -26,9 +26,14 @@ export default Component.extend({
...
 
-    selectedCurrency: null,
+    // passed in actions
+    setMembersSubscriptionSettings() {},
+
+    selectedCurrency: computed('subscriptionSettings.stripeConfig.plans.monthly.currency', function () {
+        return CURRENCIES.findBy('value', this.get('subscriptionSettings.stripeConfig.plans.monthly.currency'));
```
kevinansfield: Using a CP here means that we can select the right object from our currencies array automatically any time our config changes without needing to manually set it anywhere.

app/components/gh-members-lab-setting.js
```diff
@@ -88,13 +93,14 @@ export default Component.extend({
     init() {
         this._super(...arguments);
         this.set('mailgunRegions', [US, EU]);
-        this.set('selectedCurrency', this.get('subscriptionSettings.stripeConfig.plans.monthly.currency'));
+        this.set('currencies', CURRENCIES);
```
kevinansfield: See https://github.com/TryGhost/Ghost-Admin/pull/1430/commits/74323c9431aa198e11391eddff195c70541894b6#r359361939

app/components/gh-members-lab-setting.js
```diff
@@ -140,7 +148,6 @@ export default Component.extend({
             }
 
             if (key === 'currency') {
-                this.set('selectedCurrency', event.value);
```
kevinansfield: This wasn't working because `event.value` is a single currency value string such as `'eur'` but the `{{one-way-select}}` is expecting a currency object that matches one of the objects in the options array that was passed to it.

app/templates/components/gh-members-lab-setting.hbs
```diff
@@ -64,18 +64,38 @@
...
+                        <label class="fw6 f8"for="currency">Plan currency</label>
+                        <p>{{this.selectedCurrency.value}}</p>
+                        <span class="gh-select">
+                            {{one-way-select this.selectedCurrency
+                                id="currency"
+                                name="currency"
+                                options=(readonly this.currencies)
+                                optionValuePath="value"
+                                optionLabelPath="label"
+                                update=(action "setSubscriptionSettings" "currency")
```
kevinansfield: This is where it was falling down.

- `options` is an array of currency objects, which it's necessary to specify the path for the value and label
- `this.selectedCurrency` needs to match one of the objects in the `options` 
- `update` passes the full currency object through to the action but in the action we were [only grabbing the value](https://github.com/TryGhost/Ghost-Admin/pull/1430/commits/74323c9431aa198e11391eddff195c70541894b6#diff-216299560d5da7f490965c492f9b0186L143) and setting that to `this.selectedCurrency` which then no longer matches one of the objects in `options`

gargol: Makes sense. Thanks for explaining :)

gargol:
Landed in master with 1a61b500d92b2297a210ee9e12bfcd6a4908a2f9

PR #1431, opened 2019-12-19
✨ Enabled manual typing of post publish dates

kevinansfield:
closes https://github.com/TryGhost/Ghost/issues/9256

- stores the typed date internally to the component until
  - Enter is pressed whilst focused
  - Ctrl/Cmd+S is pressed whilst focused
  - the input loses focus
- shows an error message if the typed date is not in the correct format or is invalid
- stops Ctrl/Cmd+S propagating if the typed date is not in the correct format or is invalid
- as long as the date is valid it calls the `setDate` action when the input loses focus, Ctrl/Cmd+S or Enter is pressed

TODO / Problems:
- [x] date and error should reset if the post settings menu is closed
- [x] invalid date or format should prevent the post from being saved in the publish menu
- [ ] date and error should be reset if the post is saved outside of the date/time input

### DIFF
--- a/app/components/gh-date-time-picker.js
+++ b/app/components/gh-date-time-picker.js
@@ -145,7 +145,8 @@ export default Component.extend({
         this._timeInputElem = elem;
     }),
 
-    onDateInput: action(function (event) {
+    onDateInput: action(function (datepicker, event) {
+        datepicker.actions.close();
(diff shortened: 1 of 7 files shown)


JohnONolan:
❤️ yessss

PR #1461, opened 2020-01-17
Migrated controllers to native classes

kevinansfield:
no issue

With the Octane edition, Ember is moving to using native JS class syntax in place of it's custom EmberObject extension pattern.

- added linting for classic ember object usage inside native classes
- added ember-concurrency-decorators addon
  - provides a more concise and easier-to-read decorator syntax for ember-concurrency tasks
  - helps eslint determine when there are duplicate function names (it was previously failing to detect an action and a task that had the same name)
  - preparation for migrating to native classes
- migrated controllers to native classes
  - ran `ember-native-class-codemod` (https://github.com/ember-codemods/ember-native-class-codemod) on `app/controllers` and manually upgraded a couple of controllers that the codemod skipped
    - manually updated usage of `task` and `taskGroup` to use `ember-concurrency-decorators`
  - migrated controller-backed templates to use action methods directly rather than via the `{{action}}` helper
    - switched to using `{{fn}}` where appropriate
    - switched to using `{{on}}` where appropriate
    - skipped `{{action}}` usage where `target=` or `value=` is used for the time being, `target` requires the actions indicated to be using the action helper, `value` needs some investigation for the correct migration path

TODO:
- [ ] manually migrate `controllers/setup/three.js`
- [ ] manually migrate `controllers/signup.js`
- [ ] fix tests
- [ ] do a manual walk-through test

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -19,7 +19,11 @@ module.exports = {
         // disable linting of `this.get` until there's a reliable autofix
         'ghost/ember/use-ember-get-and-set': 'off',
 
-        // disable linting of mixins until we migrate away
-        'ghost/ember/no-new-mixins': 'off'
+        // disable linting of mixins until we've migrated away
+        'ghost/ember/no-new-mixins': 'off',
+
+        // turn on some Octane rules
+        'ghost/ember/classic-decorator-hooks': 'error',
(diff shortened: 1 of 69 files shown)


kevinansfield:
Closing because this is now very out of date.

PR #1477, opened 2020-01-23
✨Added labels for Members

rishabhgrg:
depends on https://github.com/TryGhost/Ghost/pull/11538

We are adding new ability for site-owners to tag members using labels and perform actions like filtering using labels on Admin. We want site owners/theme builders to be able to tag members through the signup process/Zapier/Admin, and corresponding label(s) should be automatically saved to them.

This PR -

- Adds new label models and validation
- Adds UI for labels dropdown on members list and detail pages
- Updates Import/Export for Members to include labels


app/controllers/members.js
```diff
@@ -2,23 +2,88 @@ import Controller from '@ember/controller';
...
+
+    labelModalData: computed('modalLabel', 'availableLabels', function () {
+        let label = this.get('modalLabel');
+        let labels = this.get('availableLabels');
+
+        return {
+            label,
+            labels
+        };
+    }),
```
rishabhgrg: @kevinansfield So I am not sure if this was the best pattern to pick here, but this basically just tries to have the modal data we use for label add/edit modal, and this seemed like the best way to have it and pass it around. I was also confused if going custom views route with a service for labels is a good approach, and felt like it wasn't. Lmk what you think about it 😬

app/templates/components/gh-members-contentfilter.hbs
```diff
@@ -0,0 +1,35 @@
...
+                <a {{on "click" (fn @onLabelAdd)}}>
+                    <span>
+                        {{svg-jar "add"}}
+                        Add Label
+                    </span>
+                </a>
+            </li>
+        </ul>
+    </GhDropdown>
+</span>
```
rishabhgrg: This is what eventually went in to make the custom-ish dropdown for labels with includes edit icon and floating add button.

rishabhgrg:
Not sure whats tripping the tests here but most of the failures seem unrelated to changes in the PR 🤔

PR #1486, opened 2020-01-28
🎨 Fixed page delete redirect

andreborud:
Realised when a page is deleted the user is redirected to posts instead of staying in pages. This pull request keeps the user in pages when deleting a page and in posts when deleting a post.


andreborud:
Should I have submitted this pull request in another repo?

PR #1489, opened 2020-02-05
Added billing route w/ billing iframe

allouis:
no-issue

### DIFF
--- a/app/components/gh-billing-iframe.js
+++ b/app/components/gh-billing-iframe.js
@@ -1,0 +1,24 @@
+import Component from '@ember/component';
+import {inject as service} from '@ember/service';
+
+export default Component.extend({
+    config: service(),
+    ghostPaths: service(),
+    ajax: service(),
+
+    didRender() {
+        let iframe = this.$('iframe')[0];
+        window.addEventListener('message', (event) => {
+            if (event && event.data && event.data.request === 'token') {
+                const ghostIdentityUrl = this.get('ghostPaths.url').api('ghost-identity');
+
+                this.ajax.post(ghostIdentityUrl).then(({token}) => {
+                    iframe.contentWindow.postMessage({
+                        request: 'token',
+                        response: token
+                    }, '*');
+                });
+            }
+        });
+    }
+});
--- a/app/components/gh-billing-iframe.js
+++ b/app/components/gh-billing-iframe.js
@@ -1,10 +1,13 @@
 import Component from '@ember/component';
+import {computed} from '@ember/object';
 import {inject as service} from '@ember/service';
 
 export default Component.extend({
     config: service(),
     ghostPaths: service(),
     ajax: service(),
+
+    billingEndpoint: computed.reads('config.billingUrl'),
 
     didRender() {
         let iframe = this.$('iframe')[0];
--- a/app/components/gh-billing-iframe.js
+++ b/app/components/gh-billing-iframe.js
@@ -10,7 +10,7 @@ export default Component.extend({
     billingEndpoint: computed.reads('config.billingUrl'),
 
     didRender() {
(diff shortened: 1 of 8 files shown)


app/components/gh-billing-iframe.js
```diff
@@ -0,0 +1,24 @@
...
+        let iframe = this.$('iframe')[0];
+        window.addEventListener('message', (event) => {
+            if (event && event.data && event.data.request === 'token') {
+                const ghostIdentityUrl = this.get('ghostPaths.url').api('ghost-identity');
+
+                this.ajax.post(ghostIdentityUrl).then(({token}) => {
+                    iframe.contentWindow.postMessage({
+                        request: 'token',
+                        repsonse: token
+                    }, '*');
```
allouis: This should not be `'*'` but the actual origin of the expected iframe

app/controllers/billing.js
```diff
@@ -0,0 +1,6 @@
+import Controller from '@ember/controller';
+import {alias} from '@ember/object/computed';
+
+export default Controller.extend({
+    guid: alias('model')
```
allouis: I had an eslint error saying I must "alias" my model - so I copied this from controllers/site.js

app/routes/billing.js
```diff
@@ -0,0 +1,14 @@
+import AuthenticatedRoute from 'ghost-admin/routes/authenticated';
+
+export default AuthenticatedRoute.extend({
+    model() {
+        return (new Date()).valueOf();
```
allouis: :shrimp: Copied from site route

app/templates/components/gh-nav-menu.hbs
```diff
@@ -83,6 +83,11 @@
                     </button>
                     <LinkTo @route="settings.labs" data-test-nav="labs">{{svg-jar "labs"}}Labs</LinkTo>
                 </li>
+                <li class="relative">
+                    <LinkTo @route="billing" data-test-nav="billing">
+                        {{svg-jar "house"}} View billing
```
allouis: Needs a different icon

app/components/gh-billing-iframe.js
```diff
@@ -0,0 +1,24 @@
...
+    didRender() {
+        let iframe = this.$('iframe')[0];
+        window.addEventListener('message', (event) => {
+            if (event && event.data && event.data.request === 'token') {
+                const ghostIdentityUrl = this.get('ghostPaths.url').api('ghost-identity');
+
+                this.ajax.post(ghostIdentityUrl).then(({token}) => {
+                    iframe.contentWindow.postMessage({
+                        request: 'token',
+                        response: token
```
allouis: Probably we want to actually design the protocol for communicating between the window->iframe

app/controllers/billing.js
```diff
@@ -0,0 +1,6 @@
+import Controller from '@ember/controller';
```
allouis: Is it possible to remove this entire file?

app/templates/components/gh-billing-iframe.hbs
```diff
@@ -0,0 +1,15 @@
+<iframe id="billing-frame" class="billing-frame" src="http://localhost:5432" frameborder="0" allowtransparency="true"></iframe>
```
allouis: `src` needs to be read from config

app/controllers/billing.js
rishabhgrg: @allouis Don't think so, as we'll need atleast the basic skeleton controller here with `guid` I think😬

rishabhgrg:
@allouis Think this is good to merge on master in this state as this puts both billing UI and route behind config from server along with basic cleanup, let me know if you think anything else is needed :)

PR #1497, opened 2020-02-21
✨  Added member impersonation

gargol:
refs https://github.com/TryGhost/Ghost/pull/11607

@kevinansfield single thing that I have in question for this PR is the use of adapter on member model 
2b98a3d, instead of a separate model for `signin_url` resource. Do you think this is clean approach or should be done differently?


gargol:
@kevinansfield refactored the adapter call to a straight ajax in the member model. Will be merging this later today unless there are any other objections :wink:

app/components/modal-impersonate-member.js
```diff
@@ -0,0 +1,39 @@
...
+    store: service(),
+
+    classNames: 'modal-impersonate-member',
+
+    signinUrl: null,
+    member: alias('model'),
+
+    init() {
+        this._super(...arguments);
+    },
```
kevinansfield: This isn't needed 🙂

app/components/modal-impersonate-member.js
```diff
@@ -0,0 +1,39 @@
...
+
+    didInsertElement() {
+        this._super(...arguments);
+
+        this._signinUrlUpdateTask.perform();
+    },
+
+    actions: {},
```
kevinansfield: Nor this

app/components/modal-impersonate-member.js
gargol: Corrected both. Thanks for catching!

PR #1505, opened 2020-02-24
Fixed email messaging for publish and send as Editor

rishabhgrg:
no issue

### DIFF
--- a/app/components/modal-confirm-email-send.js
+++ b/app/components/modal-confirm-email-send.js
@@ -1,4 +1,6 @@
 import ModalComponent from 'ghost-admin/components/modal-base';
+import {computed} from '@ember/object';
+import {pluralize} from 'ember-inflector';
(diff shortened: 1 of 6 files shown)


app/components/modal-confirm-email-send.js
```diff
@@ -10,6 +12,19 @@ export default ModalComponent.extend({
     // Allowed actions
     confirm: () => {},
 
+    deliveredToMessage: computed('model.{paidOnly,memberCount}', function () {
+        const isEditor = this.get('session.user.isEditor');
+        if (this.model.paidOnly) {
```
rishabhgrg: @gargol Thinking its probably a good idea to add an extra layer of check here for `this.model` being defined ? 🤔

rishabhgrg: @gargol Updated

PR #1551, opened 2020-04-21
Added handling for /billing route with query param passing to BMA

gargol:
The purpose if this change is to allow opening "billing" popup through a URL. It is not meant to render any information. It's purpose is:
1. Trigger open modal action
2.  Parse and pass relevant query parameter - "action"

@kevinansfield wanted to know your opinion on this approach and if the use of billing route is appropriate for this usecase. The naming of the parameters is not final (the "action" part). Naming will be reviewed once the whole flow works :) Let me know if this billing modal could be approached in a more ember-friendly way.

Example use: 
1. Open`http://localhost:2368/ghost/#/billing?action=success
2. This opens a modal with a BMA iframe which has `?action=success` passed in
3. After closing the BMA modal we return to default Ghost-Admin view

The rest of actions opening BMA - "upgrade" button and "view billing" should work as before without passing any additional parameters (action parameter should not be persisted between open/close of the BMA popup)

### DIFF
--- a/app/components/modal-billing.js
+++ b/app/components/modal-billing.js
@@ -6,7 +6,6 @@ export default ModalComponent.extend({
 
     actions: {
         closeModal() {
-            this._super(arguments);
             this.billing.closeBillingWindow();
         }
     }
--- a/app/components/modal-billing.js
+++ b/app/components/modal-billing.js
@@ -1,4 +1,13 @@
 import ModalComponent from 'ghost-admin/components/modal-base';
+import {inject as service} from '@ember/service';
 
 export default ModalComponent.extend({
+    billing: service(),
+
+    actions: {
+        closeModal() {
+            this._super(arguments);
+            this.billing.closeBillingWindow();
+        }
+    }
 });
--- a/app/controllers/billing.js
+++ b/app/controllers/billing.js
@@ -1,6 +1,0 @@
-import Controller from '@ember/controller';
-import {alias} from '@ember/object/computed';
-
-export default Controller.extend({
-    guid: alias('model')
-});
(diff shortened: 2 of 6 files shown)


app/components/modal-billing.js
```diff
@@ -1,4 +1,13 @@
 import ModalComponent from 'ghost-admin/components/modal-base';
+import {inject as service} from '@ember/service';
 
 export default ModalComponent.extend({
+    billing: service(),
+
+    actions: {
+        closeModal() {
+            this._super(arguments);
```
kevinansfield, line 9: Not needed for actions. `_super(...arguments)` is only needed when you're extending from a base class and want to run the logic in the base class before your own logic

app/routes/billing.js
```diff
@@ -0,0 +1,26 @@
...
+export default Route.extend({
+    billing: service(),
+
+    queryParams: {
+        action: {refreshModel: true}
+    },
+
+    init() {
+        this._super(...arguments);
+    },
```
kevinansfield, line 13: Not needed as you have no custom logic, without the override the base class' `init()` will be called anyway

app/controllers/billing.js
```diff
@@ -1,6 +0,0 @@
-import Controller from '@ember/controller';
-import {alias} from '@ember/object/computed';
-
-export default Controller.extend({
-    guid: alias('model')
-});
```
kevinansfield, line 6: I'd recommend keeping this file because a route always has a controller (it's created automatically for you in the bg if no explicit file exists) and it's useful to set the initial state for any any query params. Query params are always stored on the controller so an `action: null` property is useful purely for explicitness and to know what it's empty/base state is.

gargol: Oh cool, this makes a lot of sense. Thanks!

kevinansfield: Oh, the query params will also need to be defined here with `queryParams: ['action']`. The object in the route file is only used to adjust the refresh behaviour.

PR #1560, opened 2020-05-05
Added special message for "locked" users when they fail signin

naz:
refs https://github.com/TryGhost/Ghost/pull/11790

- After user fails to signin due to "locked" being locked a new view with automatic email message and more instructions will be shown to make the password reset process easier.

### DIFF
--- a/app/controllers/signin.js
+++ b/app/controllers/signin.js
@@ -21,6 +21,8 @@ export default Controller.extend(ValidationEngine, {
     authProperties: null,
 
     flowErrors: '',
+    passwordResetEmailSent: false,
+
     // ValidationEngine settings
     validationType: 'signin',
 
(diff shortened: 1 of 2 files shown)


app/controllers/signin.js
```diff
@@ -55,6 +57,10 @@ export default Controller.extend(ValidationEngine, {
 
                 this.set('flowErrors', (mainError.context.string || mainError.message.string));
 
+                if (mainError.message.string.match(/your password must be reset/i)) {
```
kevinansfield, line 60: We should be switching on the `PasswordResetRequiredError` that should be available on `mainError.type` if possible - messaging can change so it's brittle to inspect the message string.

naz:
Was superseeded by https://github.com/TryGhost/Ghost-Admin/pull/1561

PR #1562, opened 2020-05-05
✨Added regenerate button to integration api keys

rishabhgrg:
no issue

- Adds new regenerate button next to admin and content api keys for an integration
- Regenerates content or admin API key after confirmation and shows user new keys


app/controllers/settings/integration.js
```diff
@@ -131,9 +155,13 @@ export default Controller.extend({
         deleteWebhook() {
             return this.webhookToDelete.destroyRecord();
         }
-
     },
 
+    updateIntegration: task(function* () {
+        yield this.store.findRecord('integration', this.integration.id, {backgroundReload: true});
```
rishabhgrg, line 161: @kevinansfield Decided to reload the integration from server as was not sure whats the best pattern in this case, or if we should somehow hook the return value from api key update directly to refresh integration 😬

app/components/modal-regenerate-key.js
```diff
@@ -0,0 +1,38 @@
...
+            let url = this.get('ghostPaths.url').api('/integrations/', this.integration.id, 'api_key', this.apiKey.id, 'refresh');
+            try {
+                yield this.ajax.post(url, {
+                    data: {
+                        integrations: [{id: this.integration.id}]
+                    }
+                });
+                yield this.confirm();
+            } catch (e) {
+                let errMessage = 'Unable to regenerate new keys, please try again.';
```
kevinansfield, line 30: Should be `regenerate key` or `generate new key`

app/controllers/settings/integration.js
kevinansfield: Best practice would be to use [`store.pushPayload(response)`](https://api.emberjs.com/ember-data/release/classes/Store/methods/push?anchor=pushPayload) after [performing the reset](https://github.com/TryGhost/Ghost-Admin/pull/1562/files#diff-933b75966ff6010eae08c1d11b78fc15R23). There are a few examples of this in the codebase 🙂

app/components/modal-regenerate-key.js
rishabhgrg: Ah yeah, this will now get updated with new copy from Zimo - "There was an error regenerating the Admin API Key. Please try again"

rishabhgrg:
@kevinansfield Updated with the `store.pushPayload` change, thats pretty neat, thanks

PR #1568, opened 2020-05-12
Fixes change password button auto reset state

m1kl:
Made changes to allow the change password button to turn green for the success state and to automatically reset.

Fixes TryGhost/Ghost#11789 (Use case described in the referenced issue)


rishabhgrg:
Hey @m1kl , thanks for submitting PR for this fix. Have included this change in another PR which also switches on auto-reset for all task buttons by default so it doesn't need to be added explicitly. Closing this in favor of changes there to resolve conflicts and merge, have added credits ref to this PR in commit description. 😄

PR #1586, opened 2020-06-01
Added Stripe Connect UI

allouis:
no-issue

This adds the Stripe Connect UI which consists of two things!

1. A "Connect with Stripe" button which is an anchor tag to /ghost/api/v3/admin/members/stripe-connect (Redirect is handled on server)
2. An input to set the `stripe_connect_integration_token` setting via the admin API

@kevinansfield Would you mind have a check of this, just wanna make sure I'm working correctly _with_ ember, and following the style guides correctly!

:warning: 
If you want to run this locally you have to set `stripeDirect` to `false` in your config


app/components/gh-members-lab-setting.hbs
```diff
@@ -51,6 +51,46 @@
...
+                <div>
+                    <button type="button" class="gh-btn" {{action (toggle "membersStripeOpen" this)}} data-test-toggle-membersstripe><span>{{if this.membersStripeOpen "Close" "Expand"}}</span></button>
+                </div>
+            </div>
+
+            {{#liquid-if this.membersStripeOpen}}
+            <div class="flex flex-column flex-row-l items-start justify-between mb4 mt6">
+                <div class="w-100 w-50-l">
+                    <div class="mb4">
+                        <a href="api/v3/members-stripe-connect" class="stripe-connect" target="_blank"><span>Connect with Stripe</span></a>
```
allouis, line 69: Should probably use the `ghostUrls` service

app/components/gh-members-lab-setting.js
```diff
@@ -29,16 +29,22 @@ export default Component.extend({
...
 
     // passed in actions
     setMembersSubscriptionSettings() {},
+    setStripeConnectIntegrationTokenSetting() {},
 
     defaultContentVisibility: reads('settings.defaultContentVisibility'),
 
     stripeDirect: reads('config.stripeDirect'),
 
+    stripeConnectAuthUrl: computed('ghostPaths', function () {
```
allouis, line 44: Is this correct? It _seems_ like it is!

app/components/gh-members-lab-setting.js
```diff
@@ -29,16 +29,22 @@ export default Component.extend({
     feature: service(),
     config: service(),
     mediaQueries: service(),
+    ghostPaths: service(),
 
     currencies: null,
 
     // passed in actions
     setMembersSubscriptionSettings() {},
+    setStripeConnectIntegrationTokenSetting() {},
```
allouis, line 38: This name is used throughout the components and services - is it okay?

app/components/gh-members-lab-setting.js
allouis: To be more specific, is the use of `computed` correct? And is it correct that this should be at the component level, and not higher?

app/components/gh-members-lab-setting.js
kevinansfield: Could possibly remove the `Integration` part to make it a bit shorter? Otherwise seems fine 👍

app/components/gh-members-lab-setting.js
kevinansfield: It doesn't need `computed`, can be a regular getter. The only thing you'd lose is built-in memoization of the result but that memoization is a bit confusing anyway because `ghostPaths` (dependent key) does not change over time and isn't tracked/used with `.set()`

app/components/gh-members-lab-setting.js
allouis: I think I'm gonna leave it a little longer for now - just to be explicit!

app/components/gh-members-lab-setting.js
allouis: Updated and force pushed!

PR #1590, opened 2020-06-04
Enabled auto reset state for action buttons

rishabhgrg:
Closes https://github.com/TryGhost/Ghost/issues/11789

- By default, action buttons had auto-reset off and reset had to be explicitly set
- Auto reset for action buttons is on by default now, and any button that should not reset should explicitly switch it off

### DIFF
--- a/tests/acceptance/staff-test.js
+++ b/tests/acceptance/staff-test.js
@@ -522,7 +522,12 @@ describe('Acceptance: Staff', function () {
                 // Save changes
                 await click('[data-test-save-button]');
 
-                expect(find('[data-test-save-button]').textContent.trim(), 'save button text').to.equal('Saved');
+                // Since we reset save status so there's no on-screen indication
+                // that we've had a save, check the request was fired instead
+                let [lastRequest] = this.server.pretender.handledRequests.slice(-1);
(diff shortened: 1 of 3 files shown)


tests/acceptance/staff-test.js
```diff
@@ -522,7 +522,12 @@ describe('Acceptance: Staff', function () {
                 // Save changes
                 await click('[data-test-save-button]');
 
-                expect(find('[data-test-save-button]').textContent.trim(), 'save button text').to.equal('Saved');
+                // Since we reset save status so there's no on-screen indication
+                // that we've had a save, check the request was fired instead
+                let [lastRequest] = this.server.pretender.handledRequests.slice(-1);
+                let params = JSON.parse(lastRequest.requestBody);
+
+                expect(params.users[0].name).to.equal('Test User');
```
rishabhgrg, line 530: @kevinansfield Not 💯 here if this is the best way to test the saved status since we turned on the auto-reset for all buttons, so checking button text as "Saved" is not gonna work here anymore. Copied this pattern from just below where we already checked the save status with request, lmk tho if this is not ideal and should be changed.

kevinansfield: 👍 LGTM

PR #1598, opened 2020-06-08
🐛 Fixed staff profile editing incorrectly showing saved state

rishabhgrg:
closes https://github.com/TryGhost/Ghost/issues/11866

Updating a slug on user profile triggers [`updateSlug`](https://github.com/TryGhost/Ghost-Admin/blob/28aa16f9a96b13bdcac116777e2be9b41b8ca437/app/controllers/staff/user.js#L369) task, which is in the [same task group](https://github.com/TryGhost/Ghost-Admin/blob/28aa16f9a96b13bdcac116777e2be9b41b8ca437/app/controllers/staff/user.js#L452) as `save` task which is tied to the Save button on page. Every-time `updateSlug` is called after editing a slug, it triggers the task button states causing the button to incorrectly show "Saving" -> "Saved" status change though no save action is called.

- Adds extra check to task button states if the last task ran was same as the current associated task
- In case of task groups, triggering one task in group also triggered states for other tasks in group
- Editing slug called `updateSlug` task which is part of the same task group as `save` task, which triggered the state for save button

### DIFF
--- a/app/components/gh-task-button.js
+++ b/app/components/gh-task-button.js
@@ -59,9 +59,11 @@ const GhTaskButton = Component.extend({
     }),
 
     isRunning: computed('task.last.isRunning', 'hasRun', 'showSuccess', function () {
-        let isRunning = this.get('task.last.isRunning');
+        let taskName = this.get('task.name');
+        let lastTaskName = this.get('task.last.task.name');
 
-        if (this.hasRun && this.get('task.last.value') && !this.showSuccess) {
+        let isRunning = (taskName === lastTaskName) && this.get('task.last.isRunning');
+        if (this.hasRun && (taskName === lastTaskName) && this.get('task.last.value') && !this.showSuccess) {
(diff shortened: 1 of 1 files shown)


rishabhgrg:
@kevinansfield Weirdly, it seems `ember-concurrency` seems to be triggering all the tasks in a task group as in this case we are explicitly passing down the `save` task to task button but its states are still triggered when the `updateSlug` task in the same task group is called, although from the docs i couldn't find anything that suggests this or how to bypass this. 

I added a check to our task buttons to always check if the last task's name is same as the task name passed down before updating state, which takes care of group tasks and also shouldn't cause any side-effects imo, but want to run this by in case I missed anything that this will break or if there is a better way to handle this.

rishabhgrg:
@kevinansfield Good find, this issue is exactly whats causing the problem, will merge this today, thanks 👍

PR #1619, opened 2020-06-24
Fixed  #11814

devaman:
Code tag within a heading in the editor ignores font-size fixed #11814.

https://github.com/TryGhost/Ghost/issues/11814

- [x] 🐛 Bug Fixed : replace font-size from rem to em (same size that is used in ghost).


kevinansfield:
Thanks @devaman!

PR #1625, opened 2020-07-01
Added basic handling for invalid plan amount

allouis:
refs https://github.com/TryGhost/Ghost/issues/11765

This adds error handling to the stripe plan prices to ensure they are set for at least 1 USD (or other currency) or more.

### DIFF
--- a/app/components/gh-members-lab-setting.hbs
+++ b/app/components/gh-members-lab-setting.hbs
@@ -171,32 +171,35 @@
             </div>
             <div class="w-100 w-50-l flex flex-column flex-row-ns">
                 <div class="w-100 w-50-ns mr3-ns">
-                    <GhFormGroup>
+                    <GhFormGroup @errors={{this.settings.errors}} @hasValidated={{settings.hasValidated}} @property="stripePlans">
                     <label class="fw6 f8">Monthly price</label>
 
                     <div class="flex items-center justify-center mt1 gh-input-group gh-labs-price-label">
                         <GhTextInput
                             @value={{readonly this.stripePlans.monthly.amount}}
                             @type="number"
-                            @input={{action "setStripePlan" "month"}}
+                            @input={{action (mut this._scratchStripeMonthlyAmount) value="target.value"}}
+                            @focus-out={{action "validateStripePlans"}}
                         />
                         <span class="gh-input-append"><span class="ttu">{{this.stripePlans.monthly.currency}}</span>/month</span>
                     </div>
                     </GhFormGroup>
                 </div>
                 <div class="w-100 w-50-ns ml2-ns">
-                    <GhFormGroup @class="description-container">
+                    <GhFormGroup @errors={{this.settings.errors}} @hasValidated={{settings.hasValidated}} @property="stripePlans">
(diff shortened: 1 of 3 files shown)


allouis:
Hey @peterzimon i've updated the subscription inputs to display the error, but had some issues getting the margins to work - would you be able to take a look?


image

I had to get rid of the margins to make the inputs sit in the same flex box row :grimacing: Not sure if that's correct or not

allouis:
Whoooops - forgot Zimo is off the rest of the week, @kevinansfield Is this something you might be able to help with?

allouis:
So I think there's a few questions here that need answering:

1. Is the error text correct/okay?
2. How do we get the inputs to have the red border around the whole thing?
3. How can we get the error message to sit below both inputs?

@kevinansfield You might be able to help with 2. and 3. - if you have some time?

kevinansfield:
> How do we get the inputs to have the red border around the whole thing?

A new style needs to be added:

```
.error .gh-input-append {
    border-color: var(--red);
}
```

app/components/gh-members-lab-setting.js
```diff
@@ -93,11 +96,11 @@ export default Component.extend({
 
         return {
             monthly: {
-                amount: parseInt(monthly.amount) / 100 || 0,
+                amount: parseInt(monthly.amount) / 100 || 1,
                 currency: monthly.currency
             },
             yearly: {
-                amount: parseInt(yearly.amount) / 100 || 0,
+                amount: parseInt(yearly.amount) / 100 || 1,
```
matthanley, line 103: Should default values for monthly/yearly match [default settings](https://github.com/TryGhost/Ghost/blob/master/core/server/data/schema/default-settings.json#L249) of 5/50?

allouis:
I've updated this to make less changes to the HTML, but i've reintroduced the spacing bug when theres an error:

With no error:
image
 
With error:
image

app/components/gh-members-lab-setting.hbs
```diff
@@ -171,32 +171,35 @@
...
-                            @input={{action "setStripePlan" "year"}}
+                            @input={{action (mut this._scratchStripeYearlyAmount) value="target.value"}}
+                            @focus-out={{action "validateStripePlans"}}
                         />
                         <span class="gh-input-append"><span class="ttu">{{this.stripePlans.yearly.currency}}</span>/year</span>
                     </div>
                     </GhFormGroup>
                 </div>
+                <GhErrorMessage @errors={{settings.errors}} @property="stripePlans" class="w-100 red"/>
             </div>
```
matthanley, line 203: This looks like it fixes it to me, but I'm not a frontend dev 😬 

```suggestion
            </div>
            <div class="w-100 w-50-l flex flex-column flex-row-ns">
                <GhErrorMessage @errors={{settings.errors}} @property="stripePlans" class="w-100 red"/>
            </div>
```

image

also
* nitpick: 1 is valid, but the error says "greater than 1"
* might be cool to add the currency in the error: "Please enter a value greater than $1.00"

pinging @JohnONolan for feedback

allouis:
@matthanley Added your changes and improved the error message!

PR #1627, opened 2020-07-02
Added field mapping column to members importer gird

naz:
ronseal

@kevinansfield have one Q here regarding enforcing `currentlyDisplayedData` getter recalculation. Will put the deets in the code itself. If you have time for a quick glance on the approach would be very thankful as well :pray:

### DIFF
--- a/app/components/gh-members-import-table.js
+++ b/app/components/gh-members-import-table.js
@@ -8,14 +8,14 @@ export default class GhMembersImportTable extends Component {
     get currentlyDisplayedData() {
         let rows = [];
 
-        if (this.args && this.args.importData && this.args.mapping) {
+        if (this.args && this.args.importData && this.args.mapping && this.args.mapping.mapping) {
             let currentRecord = this.args.importData[this.dataPreviewIndex];
 
             for (const [key, value] of Object.entries(currentRecord)) {
                 rows.push({
                     key: key,
                     value: value,
-                    mapTo: this.args.mapping[key]
+                    mapTo: this.args.mapping.get(key)
                 });
             }
         }
--- a/app/components/gh-members-import-table.js
+++ b/app/components/gh-members-import-table.js
@@ -6,11 +6,26 @@ export default class GhMembersImportTable extends Component {
     @tracked dataPreviewIndex = 0;
 
     get currentlyDisplayedData() {
-        if (this.args && this.args.importData) {
-            return this.args.importData[this.dataPreviewIndex];
+        let rows = [];
+
+        if (this.args && this.args.importData && this.args.mapping) {
+            let currentRecord = this.args.importData[this.dataPreviewIndex];
+
+            for (const [key, value] of Object.entries(currentRecord)) {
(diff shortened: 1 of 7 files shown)


app/components/gh-members-import-table.js
```diff
@@ -6,11 +6,26 @@ export default class GhMembersImportTable extends Component {
...
+        let rows = [];
+
+        if (this.args && this.args.importData && this.args.mapping) {
+            let currentRecord = this.args.importData[this.dataPreviewIndex];
+
+            for (const [key, value] of Object.entries(currentRecord)) {
+                rows.push({
+                    key: key,
+                    value: value,
+                    mapTo: this.args.mapping[key]
```
naz, line 18: There's one bug which I have been struggling with. It has to do with the `mapping` component argument. The way to reproduce the bug is:
1. Star an import process and import CSV with multiple valid members
2. In the field mapper choose an "Import as..." value that has already been used (for example map 2 different fields to "notes")
3. The value of the map is recalculated correctly (visible through Ember inspector) and does not allow to map to the same value twice, but that is never reflected in this component display value. 
4. If you try navigating to another record through "Data" columns previous/next buttons (effects `dataPreviewIndex` value), then mapping kicks is and displays correctly :thinking: 

What I'm thinking is Ember is not picking up the value changes on this object for whatever reason. Maybe you know some trick to force it recalculate this getter without changing the `dataPreviewIndex`

naz: One approach that I'm yet to try is changing the structure of this `mapping` hash - reverse places for property and value, that way objects keys will be changing instead of values, so maybe Ember will we easier to detect that change.

kevinansfield: > What I'm thinking is Ember is not picking up the value changes on this object for whatever reason

Correct, it will only pick up changes to the `mapper` reference, not property values inside that object (this isn't much different to pre-octane ember). There are two options, either make sure that you're replacing `mapper` rather than modifying it's properties, or make `mapper` a class with the respective keys marked as tracked.

https://guides.emberjs.com/release/in-depth-topics/autotracking-in-depth/#toc_plain-old-javascript-objects-pojos

kevinansfield: You might be able to get await with doing `this.mapping = this.mapping` (octane) or `this.set('mapping', mapping)` (classic) inside `modal-import-members.js` any time you make changes to the properties as that would trigger auto tracking.

kevinansfield: If the latter classic example doesn't work then you could try `this.notifyPropertyChange('mapper')` instead.

naz: Thanks @kevinansfield :raised_hands: Went with POJO class and `@tracked` decorator on mapping attribute, this has solved the problem completely.

naz:
merged via https://github.com/TryGhost/Ghost-Admin/commit/ba346215645fe16f3bac8678129c2d6d672ba31c

PR #1632, opened 2020-07-08
Tag metadata UI

allouis:
### Changes

This adds new UI to the tag setting form, allowing the user to set the metadata properties on the tag object.

- [x] New color input above the feature image
- [x] Canonical URL added to metadata section
- [x] Twitter section added with preview
- [x] Facebook/OG section added with preview
- [x] Codeinjection head/foot section added with preview

### Prerequisites

You need to be on Ghost `master` and have run the 3.23 migrations.
- Migrate existing database `knex-migrator migrate --v 3.23 --force`
- Backup existing sqlite3 db & start fresh one `mv content/data/ghost-dev.db{,.old} && yarn start`

### Screenshots

image

image

image

image

image


allouis:
Hey @peterzimon!

Would you be able to check this PR out and give it some design review? There should be a solid base for everything here and only some tweaks needed, but if you run into any larger issues - let me know!

peterzimon:
yo @allouis 
I'm more ore less done with the review and pushed some detailed refinements. the only bug I found is:
- enter a non-valid color code
- click on save button (without removing the focus from the color field)

-> there's a flashing error message below the fields which disappears. what would be expected is to not be able to save and show the error message (similarly to e.g. the name field)

allouis:
I've fixed this now :)

PR #1633, opened 2020-07-08
🐛 Fixed social settings not saved on keyboard shortcut

rishabhgrg:
no issue

- Social account(twitter/facebook) values in general settings were only updated to settings upstream on focus out
- Keyboard shortcut save (Cmd/Ctr + S) doesn't trigger focus out event so values were not getting updated
- Updates save task to use new twitter/facebook values if changed on all save

### DIFF
--- a/app/controllers/settings/general.js
+++ b/app/controllers/settings/general.js
@@ -147,7 +147,7 @@ export default Controller.extend({
             return transition.retry();
         },
 
-        validateFacebookUrl() {
+        validateFacebookUrl(skipRenderLoop) {
             let newUrl = this._scratchFacebook;
             let oldUrl = this.get('settings.facebook');
             let errMessage = '';
@@ -186,6 +186,13 @@ export default Controller.extend({
                 }
 
                 this.set('settings.facebook', '');
+                if (skipRenderLoop) {
+                    this.set('settings.facebook', newUrl);
+                } else {
+                    run.schedule('afterRender', this, function () {
+                        this.set('settings.facebook', newUrl);
+                    });
+                }
                 run.schedule('afterRender', this, function () {
                     this.set('settings.facebook', newUrl);
                 });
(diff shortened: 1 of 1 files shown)


rishabhgrg:
@kevinansfield The issue is caused by using scratch values for updating twitter/facebook social URLs as we do a validation loop on them before updating. The fix here adds the missing update on keyboard save, but interested in your opinion here if this is a good fix or we need to refactor out the scratch value here.

app/controllers/settings/general.js
```diff
@@ -200,6 +200,13 @@ export default Controller.extend({
                 }
 
                 this.set('settings.facebook', '');
+                if (skipRenderLoop) {
```
rishabhgrg, line 203: I was not sure of side-effects of skipping render loop all-together here, but for when we update it directly from CMD+S save action, allowing render loop causes updated values to not save in settings so I added a skip option when its called from `save` 😬

kevinansfield: I think the `afterRender` was to avoid "x has been modified twice in a single render loop" error caused by our round-a-bout way of setting `""` then setting the url to force the value change to be picked up when the underlying value hasn't changed but we want to show the full url (eg: original input value of "https://facebook.com/ghost", change the input value to "ghost" and click out, the value stays as "ghost" instead of being changed back to the full url because the underlying settings value never changed to override the scratch value.

We can avoid all of that by using `this.settings.notifyPropertyChange('facebook')` which simplifies everything 🙂 I'll push a commit here in a sec because I already have it from testing.

Longer-term, all of this usage of scratch values needs to disappear, I'm sure there is a better way to do this now.

kevinansfield (approved):
@rishabhgrg I added a commit with a simplified approach. Hopefully it makes sense 😄 Feel free to merge if you agree with the change and the fix still works for you

app/controllers/settings/general.js
rishabhgrg: >  all of this usage of scratch values needs to disappear, I'm sure there is a better way to do this now.

💯 agreed 😄

rishabhgrg:
@kevinansfield Thanks 🙌 Yeah this is much cleaner in both approach and readability 😄 Merging

PR #1638, opened 2020-07-14
Added AMP Gtag UI

allouis:
refs https://github.com/TryGhost/Ghost/issues/11980

This adds a new UI to set the `amp_gtag_id` setting.

When AMP is disabled:

image

When AMP is enabled:

image


tolgaaaltas:
What about custom trackers? For example Yandex.Metrica?

PR #1644, opened 2020-07-18
startDateLabel populated

umuur:
closes #[12014](https://github.com/TryGhost/Ghost/issues/12014)

-startDateLabel variable populated in order to show start date for the members chart.

### DIFF
--- a/app/components/gh-members-chart.js
+++ b/app/components/gh-members-chart.js
@@ -18,7 +18,9 @@ export default Component.extend({
     stats: null,
     chartData: null,
     chartOptions: null,
-    startDateLabel: '',
+    startDateLabel: computed('membersStats.days', function () {
+        return moment(new Date()).add(- this.membersStats.days + 1, 'days').format(DATE_FORMAT);
+    }),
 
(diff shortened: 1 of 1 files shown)


app/components/gh-members-chart.js
```diff
@@ -18,7 +18,9 @@ export default Component.extend({
     stats: null,
     chartData: null,
     chartOptions: null,
-    startDateLabel: '',
+    startDateLabel: computed('membersStats.days', function () {
+        return moment(new Date()).add(-this.membersStats.days + 1, 'days').format(DATE_FORMAT);
```
kevinansfield, line 22: Unfortunately this won't work when the "All time" range is chosen because `memberStats.days` will be `"all-time"` and not a number.

What you could do instead is to use `memberStats.stats` as the dependent key and use `Object.keys(memberStats.stats.total_on_date)[0]` to grab the first date in the range and build the label from there.

umuur: ahh, thank you for that!

PR #1645, opened 2020-07-20
Removed signin form hint to which field is incorrect

zann1x:
As an extra layer of security, the sign in form does not show whether the mail or password field is wrong.
Without this, one could guess whether or not a certain mail address is registered as a user and can therefore go straight to guessing the user's password.
Refers to https://github.com/TryGhost/Ghost/pull/12068


ErisDS:
This has been discussed multiple times, this change is not desirable. The brute force protection we have in place ensures security. This feature is for UX, which is far more important than "security through obscurity"

PR #1647, opened 2020-07-24
Updated delete member UI to add toggle to cancel subscriptions

allouis:
no-issue

This adds a new adapter for the member model so we can (optionally) add a `cancel` query parameter to the DELETE URL for members

Also updates the delete member modal to include a conditionally (based on active subscriptions) displayed toggle to enable/disable the `cancel` query parameter.

### DIFF
--- a/app/components/modal-delete-member.hbs
+++ b/app/components/modal-delete-member.hbs
@@ -7,6 +7,21 @@
     <p>
         You're about to delete "<strong>{{or this.member.name this.member.email}}</strong>". This is permanent! We warned you, k?
     </p>
+
+    {{#if this.hasActiveStripeSubscriptions}}
+        <div class="flex justify-between">
+            <div>
+                <h4>Cancel Stripe subscription</h4>
+                <p class="pa0 ma0">If disabled, member's Stripe subscription will continue</p>
+            </div>
+            <div class="for-switch">
+                <label class="switch" >
+                    <input type="checkbox" checked={{this.cancelSubscriptions}} class="gh-input" onclick={{action "toggleCancelSubscriptions"}}>
+                    <span class="input-toggle-component mt1"></span>
+                </label>
+            </div>
+        </div>
+    {{/if}}
 </div>
 
 <div class="modal-footer">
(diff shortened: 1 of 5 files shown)


app/components/modal-delete-member.js
```diff
@@ -1,25 +1,46 @@
 import ModalComponent from 'ghost-admin/components/modal-base';
 import {alias} from '@ember/object/computed';
+import {computed} from '@ember/object';
 import {inject as service} from '@ember/service';
 import {task} from 'ember-concurrency';
 
 export default ModalComponent.extend({
     membersStats: service(),
 
+    cancelSubscriptions: false,
```
kevinansfield, line 10: Can we change this to `shouldCancelSubscriptions`? I was tripped up thinking we were passing a function/action around when reading later code 😄

app/components/modal-delete-member.hbs
```diff
@@ -7,6 +7,21 @@
...
+
+    {{#if this.hasActiveStripeSubscriptions}}
+        <div class="flex justify-between">
+            <div>
+                <h4>Cancel Stripe subscription</h4>
+                <p class="pa0 ma0">If disabled, member's Stripe subscription will continue</p>
+            </div>
+            <div class="for-switch">
+                <label class="switch" >
+                    <input type="checkbox" checked={{this.cancelSubscriptions}} class="gh-input" onclick={{action "toggleCancelSubscriptions"}}>
```
kevinansfield, line 19: `onclick=` is kinda deprecated now. We should use element modifiers wherever possible, in this case the [`{{on}}` modifier](https://guides.emberjs.com/release/components/template-lifecycle-dom-and-modifiers/#toc_event-handlers)

app/components/modal-delete-member.js
allouis: Yes! Good point :relaxed:

app/components/modal-delete-member.hbs
allouis: :+1: Thanks

PR #1649, opened 2020-07-25
✨Added bulk delete of filtered members list in admin

kevinansfield:
requires TBD

- display a "Delete" button in the members list header whenever a filter or search is applied, clicking the button opens the delete members modal
- added a pre-check when opening the bulk delete modal that queries the API to get a count of all paid members matching the selected filter and displays a warning about bulk delete not cancelling Stripe subscriptions if any exist
- added `{{gh-pluralize}}` helper that is the same as `{{pluralize}}` but formats larger numbers nicely using our `{{format-number}}` helper
- bulked out member/label mocking in mirage
  - added details of label model/relationships and added label factory
  - added basic 'paid' trait to member factory
  - moved `extractFilterParam` out of posts mock into top-level mirage utils file
  - updated 'GET /members/` mock to filter mirage db based on the filter/search/paid query params we currently have in use

### DIFF
--- a/app/components/gh-members-filter.hbs
+++ b/app/components/gh-members-filter.hbs
@@ -1,22 +1,28 @@
 <span class="dropdown dropdown-topmenu">
     <GhDropdownButton
         @dropdownName="members-label-menu"
-        @classNames="gh-btn gh-btn-white gh-btn-filter first" @title="Member Labels"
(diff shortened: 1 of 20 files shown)


kevinansfield:
Closing as it's out of date. Branch is kept for future reference when the feature is picked up again.

PR #1668, opened 2020-08-09
:art Improved UI for 'locked' staff users

bonaert:
closes TryGhost/Ghost#12086

The current UI shows a 'Locked' label that looks like a user role label. This is unclear because it does not indicate why the user is locked; it is also presented as a user label even though it is not in the same category of information as the role. As suggested in the issue, this was improved in 2 ways:
  - Replaced 'Locked' word by a lock icon
  - Hovering over staff user shows a tooltip with the text 'Requires password reset to log in' above the user information

**Before**
previous_ui

**After**:
ghost_admin_locked_ui

**Info**:

### DIFF
--- a/app/components/gh-user-list-item.hbs
+++ b/app/components/gh-user-list-item.hbs
@@ -1,4 +1,4 @@
-<div class="apps-grid-cell">
+<div class="apps-grid-cell" data-tooltip="{{ if user.isLocked 'Requires password reset to log in'}}">
     <LinkTo @route="staff.user" @model={{user.slug}} data-test-user-id={{user.id}}>
     <article class="apps-card-app">
         <div class="apps-card-left">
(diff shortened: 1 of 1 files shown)


JohnONolan:
Thanks for the PR! The implementation looks correct :) in terms of style I think we'd like to try and keep the elements all a little closer together and more related — with a box for the lock icon, and the tooltip appearing beside it, like this:

image

Or another way to do it, if easier, could be to make the text appear as the hover state within the lock box, like this:

image

kevinansfield:
Closing as this was implemented as part of https://github.com/TryGhost/Ghost-Admin/commit/e1c05db64a4527edbcaecec4daeb630cc6878169 which built on top of this PR. Thanks for your contribution!

PR #1684, opened 2020-08-19
Increase the user bio field length to 800 characters.

johnbellone:
There have been a [few posts about this in the forum][1]. The primary
use case here is to have a longer bio for an author. This is an
arbitrary number that I chose - if it should be longer am absolutely
open to suggests!

[1]: https://forum.ghost.org/t/bio-length-for-users/1084

### DIFF
--- a/app/validators/user.js
+++ b/app/validators/user.js
@@ -31,7 +31,7 @@ const userValidator = BaseValidator.extend(PasswordValidatorMixin, {
         let bio = model.get('bio');
 
         if (this.isActive(model)) {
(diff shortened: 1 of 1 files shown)


ErisDS:
Hey @johnbellone - the forum post you linked to is one where I asked for a use case and never got a reply. We don't tend to make changes like this without clear use cases and you've said yourself you picked a random number.

IMO this should stay as-is until there's a clear case for changing it.

naz:
See https://github.com/TryGhost/Ghost/pull/12147#issuecomment-703328016

PR #1693, opened 2020-08-31
Fixed account menu spacing issue

sanne-san:
Not related to any issue.

- Configured the dropdown to auto position itself vertically and enabled the renderInPlace flag. 

- Removed unused function ```userDropdownPosition```

- Changed overflow value to visible to ensure the entire content space was visible


sanne-san:
Hey @daniellockyer I've reverted back to absolute positioning as it introduced the scrolling bug. I think I found a different way to solve the spacing issue now - could you see if this fixed this space gap for you as well?

PR #1702, opened 2020-09-10
🐛 Fixed data binding for tag meta-data in Admin Console.

iamscottcab:
closes TryGhost/Ghost#12190

Binds the values in `gh-tag-settings-form.hbs` to the `tag` object rather than the `scratchTag` object which only exposes fields that are present in the "basic info" section and not the "meta data" section.

Double checked that all fields are correctly bound in admin console and also confirmed values were being passed through to the `<head>` tag, `{{ghost_head}}` and `{{ghost_foot}}` properties of a given post.

### DIFF
--- a/app/components/gh-tag-settings-form.hbs
+++ b/app/components/gh-tag-settings-form.hbs
@@ -129,7 +129,7 @@
                         @id="canonical-url"
                         @name="canonicalUrl"
                         @tabindex="4"
-                        @value={{this.scratchTag.canonicalUrl}}
+                        @value={{this.tag.canonicalUrl}}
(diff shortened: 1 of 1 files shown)


allouis:
Hey!

Thanks for your PR, it's appreciated! :relaxed: 

Although this fixes the issue, it doesn't follow our pattern of "scratch" values (which was implemented incorrectly in this file, causing the issue) - I've opened a PR [here](https://github.com/TryGhost/Ghost-Admin/pull/1714) with what I think is the correct approach - I'm going the close this PR in favour of that one.

PR #1703, opened 2020-09-10
Allow YAML file for redirects.

sainthkh:
ref TryGhost/Ghost#11085
ref TryGhost/Ghost#12187

This PR allows users to upload yaml format redirects files to the server.

### DIFF
--- a/app/templates/settings/labs.hbs
+++ b/app/templates/settings/labs.hbs
@@ -84,8 +84,8 @@
             <div class="gh-expandable">
                 <div class="gh-expandable-block">
                     <GhUploader
-                        @extensions={{this.jsonExtension}}
-                        @uploadUrl="/redirects/json/"
+                        @extensions={{this.redirectsFileExtensions}}
+                        @uploadUrl="/redirects/upload/"
                         @paramName="redirects"
                         @onUploadSuccess={{perform this.redirectUploadResult true}}
                         @onUploadFailure={{perform this.redirectUploadResult false}}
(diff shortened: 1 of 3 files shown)


app/templates/settings/labs.hbs
```diff
@@ -157,15 +157,15 @@
                                 {{else if this.redirectFailure}}
                                     {{svg-jar "retry"}} Upload Failed
                                 {{else}}
-                                    Upload redirects JSON
+                                    Upload redirects JSON/YAML
```
naz, line 160: @peterzimon we are adding second file format (`.yaml`) support here, needs a copy check.

naz:
Todo:
- [ ] update redirects file upload endpoint based on [this convo](https://github.com/TryGhost/Ghost/pull/12187/files#r511699113)

app/templates/settings/labs.hbs
naz: My thinking is, if our preferred format is YAML, maybe we should hide the `JSON` option completely and still support it if somebody tries uploading a `.json` file.

peterzimon: > My thinking is, if our preferred format is YAML, maybe we should hide the JSON option completely and still support it if somebody tries uploading a .json file.

I like this idea. since YAML is our preferred format we could use `Upload redirects YAML/JSON` or even `Upload redirects YAML` as a button label?

sainthkh: I think completely removing `JSON` from the button might confuse users because they would think JSON is not supported anymore. I think it might be better to remove `JSON` in v4.

sainthkh: @peterzimon When you have time, can you review the copy [here](https://github.com/TryGhost/Ghost/pull/12187#discussion_r508144220)?

ErisDS:
@naz would you please take a quick look at this PR again? The Ghost part  was merged, but this prevents it being used. Thank you!

naz:
@ErisDS have rebased and tested this change minimally. While on the topic, confirming we are going to drop support for `.json` format in API v5 instead of the current v4, correct?

app/templates/settings/labs.hbs
naz: @peterzimon ping about the above :wink:

naz:
If the above is correct I think we are ready to merge this after the copy is reviewed [here](https://github.com/TryGhost/Admin/pull/1703#discussion_r515886890).

peterzimon (approved):

PR #1705, opened 2020-09-13
Postpublish

smithahebbar:

- [x ] There's a clear use-case for this code change
- [x ] Commit message has a short title & references relevant issues
- [x ] The build will pass (run `ember test` from the repo root - will be `core/client` if working from the submodule in Ghost).


kevinansfield:
@smithahebbar I'm going to close this PR as there's no description for why this test is being added and it's not using any of our test patterns so will result in quite a brittle test that will break any time markup or styles are changed.

If you want to create a PR for something, please open an issue first so that the problem can be validated and the appropriate path to fixing it can be discussed 🙂

PR #1714, opened 2020-09-21
Fixed tag metadata fields not being populated

allouis:
closes https://github.com/TryGhost/Ghost/issues/12190

The "scratch" concept was being used incorrectly for codeinjection
fields and was not initialised with all the required fields.


allouis:
Hey @kevinansfield can you check that I've correctly used our "scratch" concept here?

kevinansfield:
> Hey @kevinansfield can you check that I've correctly used our "scratch" concept here?

In this instance I'm not sure there's a need for scratch values at all, the only reason I can think of for why they would be used here is so that the title shown at the top of the screen doesn't update whilst you're typing. Fixing the scratch value usage is a bigger refactor though, this change is working so we can merge and come back to it later 👍

allouis:
:dancers: Thanks!!!

PR #1716, opened 2020-09-23
Added support for filtering the exported Members CSV

allouis:
:warning: Relies on this https://github.com/TryGhost/Ghost/pull/12228

#### Export all
image
#### Export filtered
image
#### Export filtered (no matches)
The button has the `disabled` attribute, so will not allow to export - can be styled based on this attribute too
image


allouis:
No idea why there's scrollbars on the screenshots, hopefully this gives you an idea of what it looks like though @peterzimon

PR #1739, opened 2020-10-26
Dispatch event to root only if the key combo is registered as shortcut

sainthkh:
Closes TryGhost/Ghost#12294.

Cache registered shortcuts and check KeyboardEvent before dispatching event to the root.

Fixed the bug introduced with TryGhost/Ghost-Admin#1707.

I wish there were a way to test ctrl+v event. But unfortunately, native events are not executed by triggering event in the acceptance tests.

----

### DIFF
--- a/app/components/gh-token-input.js
+++ b/app/components/gh-token-input.js
@@ -1,6 +1,7 @@
 /* global key */
 import Component from '@glimmer/component';
 import Ember from 'ember';
+import shortcuts from '../utils/shortcuts';
 import {A, isArray} from '@ember/array';
 import {action, get} from '@ember/object';
 import {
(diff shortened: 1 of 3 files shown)


ErisDS:
@sainthkh Thank you for this PR! I apologise for the delay but this is on my list to review ASAP.

kevinansfield:
Hey @sainthkh 👋 So sorry this was left for such a long time. I've tested and it works for the main use-case of copy/paste in the post tags input, thanks! 🎉 

I'm going to merge but just wanted to note two concerns for future reference:
1. we're trying to move away from keymaster and the shortcuts mixins, replacing all shortcut handling with [ember-keyboard](https://github.com/adopted-ember-addons/ember-keyboard) so there may be some places where we use the token input where there aren't any keymaster shortcuts registered
2. the shortcuts cache util doesn't take the keymaster scope into account so it's possible there are some edge cases where it doesn't quite work as expected

Thanks again for the contribution 😊

PR #1740, opened 2020-10-26
Added `past_due` and `unpaid` subscriptions for members

rishabhgrg:
refs https://github.com/TryGhost/Ghost/issues/12256 , https://github.com/TryGhost/Ghost/issues/12255

Currently when listing subscriptions for Members, we only show the subscriptions which have a status of trialing or active.

Based on discussion, the `unpaid` and `past_due` states on Stripe also represent owner's intention of considering a subscription as active instead of `cancelled`, so we allow any subscriptions under these 2 states to be also listed for a member and consider them as `paid`.

- Subscriptions will go into a past_due state if the payment is missed, this should be considered a grace period where the member still has access.

- After this the subscriptions will either go to the unpaid or the cancelled state - this can be configured on an account by account basis in the Stripe dashboard. `unpaid` is considered as an intention to keep the subscription to allow for re-activation later.

### DIFF
--- a/app/components/gh-member-settings-form.js
+++ b/app/components/gh-member-settings-form.js
@@ -33,13 +33,14 @@ export default Component.extend({
         let subscriptions = this.member.get('stripe');
         if (subscriptions && subscriptions.length > 0) {
             return subscriptions.map((subscription) => {
-                const status = subscription.status === 'past_due' ? 'Past due' : subscription.status;
+                const statusLabel = subscription.status === 'past_due' ? 'Past due' : subscription.status;
                 return {
                     id: subscription.id,
                     customer: subscription.customer,
                     name: subscription.name || '',
                     email: subscription.email || '',
-                    status: status,
+                    status: subscription.status,
+                    statusLabel: statusLabel,
                     startDate: subscription.start_date ? moment(subscription.start_date).format('D MMM YYYY') : '-',
                     plan: subscription.plan,
                     amount: parseInt(subscription.plan.amount) ? (subscription.plan.amount / 100) : 0,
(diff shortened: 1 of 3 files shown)


app/components/gh-member-settings-form.js
```diff
@@ -33,12 +33,13 @@ export default Component.extend({
...
         if (subscriptions && subscriptions.length > 0) {
             return subscriptions.map((subscription) => {
+                const status = subscription.status === 'past_due' ? 'Past due' : subscription.status;
                 return {
                     id: subscription.id,
                     customer: subscription.customer,
                     name: subscription.name || '',
                     email: subscription.email || '',
-                    status: subscription.status,
+                    status: status,
```
allouis, line 42: IMO we should have an extra property that's like "human readable status" or "display status" - just so that we can use the original status values if we ever need to switch on them in the future - what do you think?

rishabhgrg: Yeah agree, good point. I'll add a  `statusLabel` field with the human readable version so it can be easily extended 👍

app/components/gh-member-settings-form.js
```diff
@@ -33,12 +33,14 @@ export default Component.extend({
         let subscriptions = this.member.get('stripe');
         if (subscriptions && subscriptions.length > 0) {
             return subscriptions.map((subscription) => {
+                const statusLabel = subscription.status === 'past_due' ? 'Past due' : subscription.status;
```
allouis, line 36: Does this need to change `'active'` to `'Active'` and that? Or is that done in CSS or something?

rishabhgrg: That's done in the CSS already, basically capital-cases the status value

allouis (approved):

PR #1761, opened 2020-11-12
Allow non-FQDNs in webhook URLs

BlackCetha:
no issue

I need this change in order to comfortably use webhooks in my Docker environment.
My plan was to have my ghost container along with a custom webhook receiver application in the same network.
This way, automatic DNS discovery would be enabled and I could just use the name of my service as a webhook endpoint.
However, docker doesn't allow dots in service names, so the DNS entry for my receiver is always a single level domain.
The fix for this is simple: Just disable the check for a top-level domain in the webhook validator.
That happens here: https://github.com/validatorjs/validator.js/blob/master/src/lib/isFQDN.js#L24
I don't see any issues that could arise from disabling this check.

### DIFF
--- a/app/validators/webhook.js
+++ b/app/validators/webhook.js
@@ -28,7 +28,7 @@ export default BaseValidator.create({
     targetUrl(model) {
         if (isBlank(model.targetUrl)) {
             model.errors.add('targetUrl', 'Please enter a target URL');
-        } else if (!validator.isURL(model.targetUrl || '', {require_protocol: false})) {
+        } else if (!validator.isURL(model.targetUrl || '', {require_protocol: false, require_tld: false})) {
             model.errors.add('targetUrl', 'Please enter a valid URL');
         } else if (!validator.isLength(model.targetUrl, 0, 2000)) {
             model.errors.add('targetUrl', 'Target URL is too long, max 2000 chars');


ErisDS:
@BlackCetha  Thank you for this PR! I apologise for the delay but this is on my list to review ASAP.

BlackCetha:
Hi, thanks for coming back to this.

In the meantime, I have found a workaround using alternative names for in the docker configuration.
However, I still think should change should be implemented.

There is also corresponding validation code in the core repo which would have to be changed aswell.

BlackCetha:
Or at least I thought so, now I can't seem to find it.

ErisDS:
Hi @BlackCetha, I'm really sorry that we didn't get around to reviewing this PR in a reasonable timeframe. There are two things we need to iron out with this PR - the first is that it doesn't have a partner PR to do the server piece, and the second is we have some concerns about this opening up security holes, so it would need to be audited. 

We're in the process of cleaning up all the PRs on this repo, before moving the admin to be part of the [Ghost monorepo](https://github.com/TryGhost/Ghost). This will make it far easier to do PRs like this where there is both a client and server component, as you'll be able to do both in a single PR and more easily test that everything is working together.

If you're still interested in making these changes, please re-create the PR on the Ghost monorepo once admin is merged in (should be in the next few days). We'll then take a look at the PR right away.

PR #1771, opened 2020-11-18
Added the Personal Tokens feature

tpatel:
refs TryGhost/Ghost#12373

:warning: depends on TryGhost/Ghost#12374

### DIFF
--- a/app/adapters/api-key.js
+++ b/app/adapters/api-key.js
@@ -1,0 +1,16 @@
+import ApplicationAdapter from './application';
+
+export default ApplicationAdapter.extend({
+
+    queryRecord(store, type, query) {
+        if (!query || query.id !== 'me') {
+            return this._super(...arguments);
+        }
+
+        let url = `${this.buildURL('users', 'me')}token/`;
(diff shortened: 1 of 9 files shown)


app/adapters/api-key.js
```diff
@@ -0,0 +1,12 @@
+import ApplicationAdapter from './application';
+
+export default ApplicationAdapter.extend({
+
+    queryRecord() {
+        let url = `${this.buildURL('users', 'me')}token/`;
```
kevinansfield, line 6: It would be good to add a guard here so we can still query api-tokens that aren't the personal token similar to what we do in [the `users` adapter](https://github.com/TryGhost/Ghost-Admin/blob/master/app/adapters/user.js#L6-L8)

kevinansfield (changes requested):
Minor change request but otherwise it all LGTM 👍

app/adapters/api-key.js
tpatel: Thanks for the feedback :pray:  I've updated my code :heavy_check_mark:

kevinansfield (approved):

PR #1818, opened 2021-01-09
🐛 Fixed flash of unstyled content appearing when toggling night mode

harshmaur:
closes TryGhost/Ghost#12396

We use two stylesheets ghost-light.css and ghost-dark.css for light mode and night mode respectively.
The problem occurs when we disable the light stylesheet when we enable dark stylesheet, as there is a moment when there is no style applied
This causes flash of unstyled content (or a big logo when users flip the night mode switch)

Since the ghost-dark style is loaded after the ghost-light stylesheet, we only need to enable or disable the dark stylesheet.
When the dark stylesheet is enabled, the light stylesheet will be overridden automatically by the browser.
I could not find any performance implications on overriding styles. The only performance implication is around loading a new stylesheet
which we already do.

### DIFF
--- a/app/services/feature.js
+++ b/app/services/feature.js
@@ -126,11 +126,9 @@ export default Service.extend({
 
         return this.lazyLoader.loadStyle('dark', 'assets/ghost-dark.css', true).then(() => {
             $('link[title=dark]').prop('disabled', !nightShift);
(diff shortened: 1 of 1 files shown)


kevinansfield:
Thanks @harshmaur

mcnaveen:
Wow. Thanks @harshmaur

PR #1838, opened 2021-02-12
Move billing iFrame to render in main content area

aileen:
no issue

The billing iFrame needs to be rendered within the `main` content area rather than as fullscreen.

This needs most probably more tidying up, but is working for now without breaking the functionality.


aileen:
@kevinansfield Not sure if this is the correct approach for that. The requirement is that the iframe still works as before, but without being rendered fullscreen. Please feel free to merge/update however needed 😄

kevinansfield (approved):

PR #1844, opened 2021-02-18
Added top members by email open rate to dashboard

allouis:
refs https://github.com/TryGhost/Team/issues/469

### DIFF
--- a/app/controllers/dashboard.js
+++ b/app/controllers/dashboard.js
@@ -35,6 +35,10 @@ export default class DashboardController extends Controller {
         loading: false
     };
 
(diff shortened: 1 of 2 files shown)


app/controllers/dashboard.js
```diff
@@ -22,6 +22,10 @@ export default class DashboardController extends Controller {
         loading: false
     };
 
+    get showTopMembers() {
+        return this.feature.get('emailAnalytics');
```
allouis, line 26: This always returns true, even if I switch the setting off in email newsletter settings :thinking:

PR #1851, opened 2021-02-23
Made PSM social subviews wider

kevinansfield:
no issue

- adds an extra class to the PSM container to increase it's width when viewing a wide subview
- made meta and social subviews "wide" to create space for the more realistic previews that are used in the preview modal


kevinansfield:
Closing because we went with an [outright wider PSM](https://github.com/TryGhost/Admin/commit/c6a12755f66e15f03bf46cb3515166f57ba84502) instead of changing size based on which subview is shown.

PR #1859, opened 2021-03-01
Ensured redirect for /billing route when not signed in

aileen:
no issue

Fixes a bug where the billing iframe would show a blank page when called directly and user is not logged in. This commit ensures to that we have a valid authenticated session and redirects to the signin page if not.

### DIFF
--- a/app/routes/billing.js
+++ b/app/routes/billing.js
@@ -3,12 +3,21 @@ import {inject as service} from '@ember/service';
 
 export default Route.extend({
     billing: service(),
+    session: service(),
 
     queryParams: {
         action: {refreshModel: true}
     },
 
     beforeModel(transition) {
+        this._super(...arguments);
+
+        // Ensure that we redirect to signin if billing route is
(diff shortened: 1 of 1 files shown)


aileen:
@kevinansfield not 100% sure if that's the right approach, hence this PR. Feel free to merge/change.

kevinansfield:
@aileen no need to handle this manually, we have an `AuthenticatedRoute` class to extend from that handles this automatically 🙂 Pushed the change here https://github.com/TryGhost/Admin/commit/a31ce312212255c826f9d6793eae5f3584a12078

aileen:
@kevinansfield I'm still not getting redirected with the `AuthenticatedRoute`. With my first approach it worked fine.

Steps to reproduce:
1. Sign out of Admin
2. Open http://localhost:2368/ghost/#/billing/plans
3. See blank page with `AuthenticatedRoute`, but get redirected to `/signin` with my first implementation

PR #1881, opened 2021-04-02
 🐛 Fixed host limit error when resending a pending invite

naz:
refs https://github.com/TryGhost/Team/issues/587

 - It was not possible to invite a new Contributor to a site that has reached its limit
 - This fix divides the resend invite action into two separate steps: 1. Deleting existing invite 2. Creating a new one for the same data
 - It was hitting the limit because the permissions were correctly checking for "add" action being executed when the limit of invites/staff users was already used up. By removing existing invite we mimic the behavior the server would do for us anyway (https://github.com/TryGhost/Ghost/blob/2111992dd42cd9d9400eebea52006d023b4b19b3/core/server/services/invites/invites.js#L18-L22)

@rishabhgrg I struggled to figure out why the test for the invites was failing and not picking up the second request. I'm assuming the "run" block in which the `invite.resend` is placed doesn't wait till the second request to complete :thinking: Can you help with this? You have a lot more experience with Ember and it's test, so thought I'd reach out!

### DIFF
--- a/app/models/invite.js
+++ b/app/models/invite.js
@@ -24,9 +24,12 @@ export default Model.extend({
 
         let inviteUrl = this.get('ghostPaths.url').api('invites');
 
-        return this.ajax.post(inviteUrl, {
-            data: JSON.stringify({invites: [inviteData]}),
-            contentType: 'application/json'
-        });
+        return this.ajax.del(`${inviteUrl}${this.get('id')}`)
+            .then(() => {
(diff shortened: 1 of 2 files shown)


rishabhgrg:
@naz Sorry missed this on Friday, will take a look here today on why tests are misbehaving :)

naz:
Thaaanks!

rishabhgrg (approved):

PR #1925, opened 2021-04-20
✨ Added option to disable member subscriptions

kevinansfield:
refs https://github.com/TryGhost/Team/issues/579

- new Members Access settings screen that moves subscription access and default post visibility from the Members Payments settings screen
- expanded "Free signup" toggle into three signup access options
  - "anyone" - same as previous "allow free member signup" option set to `true`
  - "invite only" -  same as previous "allow free member signup" option set to `false`
  - "nobody" - completely disables member signup and login. Removes injected portal and stripe scripts on the front-end and hides member related sections on the admin dashboard

(this commit moves the above changes out from behind the developer experiments flag and cleans up now-unused code)


ramrami:
@kevinansfield how do we check for this in theme handlebars ?

rajivshah3:
@ramrami I haven't seen a way to do this yet but I'm looking for the same thing. I created a [forum post](https://forum.ghost.org/t/determining-whether-members-feature-is-enabled/22948)

ramrami:
@rajivshah3 Kevin answered [here](https://forum.ghost.org/t/ghost-4-4-0-you-can-now-disable-membership-feature-and-portal/22487/11) 

"For the time being the `@labs.members` property still works although it’s deprecated and may be removed in the next major release. A non-deprecated replacement is being worked on."

rajivshah3:
Oh, great! Thanks @ramrami

PR #1929, opened 2021-04-23
[WIP] Wire up member details

allouis:
:icecream:


app/components/modal-member-product.js
```diff
@@ -1,6 +1,89 @@
...
+    *addPriceTask() {
+        let url = this.ghostPaths.url.api('members', this.member.get('id'), 'subscriptions');
+
+        let response = yield this.ajax.post(url, {
+            data: {
+                stripe_price_id: this.price.stripe_price_id
+            }
+        });
+
+        this.store.pushPayload('member', response);
```
allouis, line 85: The members products are not updated after this, you need to refresh the page to see them

app/routes/member.js
```diff
@@ -26,10 +26,10 @@ export default class MembersRoute extends AuthenticatedRoute {
     }
 
     model(params) {
-        this._requiresBackgroundRefresh = false;
+        this._requiresBackgroundRefresh = true;
```
allouis, line 29: Not 100% sure if I need this

app/routes/member.js
```diff
@@ -26,10 +26,10 @@ export default class MembersRoute extends AuthenticatedRoute {
...
 
         if (params.member_id) {
-            return this.store.queryRecord('member', {id: params.member_id, include: 'email_recipients'});
+            return this.store.queryRecord('member', {id: params.member_id, include: 'email_recipients,products'});
```
allouis, line 32: I've added this here AND to the controller, I'm not sure which is correct, or if we need both

PR #1938, opened 2021-04-30
Display error message from limit service on theme upload

ceecko:
no issue

- Addresses https://github.com/TryGhost/Admin/commit/16debb75a994f16ff5079f2f2896ed9958e4cbbe#r49241896
- Uses limit service to display the configured error message

### DIFF
--- a/app/components/modal-upgrade-host-limit-custom-theme.hbs
+++ b/app/components/modal-upgrade-host-limit-custom-theme.hbs
@@ -6,7 +6,11 @@
 
 <div class="modal-body">
     <p>
-        Your current plan only supports official themes. You can install them from the <a href="https://ghost.org/marketplace/">Ghost theme marketplace</a>.
+        {{#if this.model.limitErrorMessage}}
+            {{html-safe this.model.limitErrorMessage}}
+        {{else}}
+            Your current plan only supports official themes. You can install them from the <a href="https://ghost.org/marketplace/">Ghost theme marketplace</a>.
+        {{/if}}
     </p>
 </div>
 
--- a/app/controllers/settings/theme.js
+++ b/app/controllers/settings/theme.js
@@ -55,6 +55,7 @@ export default Controller.extend({
     themes: null,
     themeToDelete: null,
     displayUpgradeModal: false,
+    limitErrorMessage: null,
 
     init() {
         this._super(...arguments);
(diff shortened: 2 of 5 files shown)


tpatel:
Hi @ceecko, thank you for the high quality PR, I liked that you built it from a comment you found in another commit 🙏 

I noticed that the default messages still appears when clicking on `Upload a theme`, would that be ok for you to also make it use the configured error message?

ceecko:
I have just double-checked the `Upload a theme` button and it appears to be showing the message properly.
There's a race condition though. The `model()` in `UploadthemeRouter` does not "await" the result from the limit service. I intentionally did not put an await there because there's no loading indicator when you press the button and in case the await takes a long time the UI may seem unresponsive.
See this line for reference https://github.com/TryGhost/Admin/pull/1938/files#diff-47b05d92b88b2c2ae4780bebbaf24342b0ec1fa1933c0bfc7c09430ec2dfe144R10

I believe this is the root cause why you don't see the message updated.
Would it make sense to put an "await" there even though there's no loading indicator? With the current setup the await shall be resolved instantly since there's no call to the database in the limit service at the moment.

tpatel:
Interesting, I seem to fall in the race condition most of the time!  I think we can have an `await` for now, maybe you could add a TODO comment to remind us to introduce a spinner when the limit service has an actual async operation?

Another thing, you'd need to use this trick to make the `errorIfWouldGoOverLimit` fail with a `HostLimitError` instead of an `IncorrectUsageError`:
```
// Sending a bad string to make sure it fails (empty string isn't valid)
await limitService.errorIfWouldGoOverLimit('customThemes', {value: '.'});
```
([source](https://github.com/TryGhost/Ghost/blob/273e22032761d39aef10e0b9a90584f31b0271c9/core/server/api/canary/themes.js#L150))

I'm not happy about this trick but I wanted to be sure that developers would always provide a value to `errorIfWouldGoOverLimit`.

ceecko:
I have added `await`.
It's strange though as even if you fall into the race condition the message shall update automatically within a couple milliseconds. I suppose it was the `IncorrectUsageError` instead of `HostLimitError` being triggered.

tpatel (approved):
Thanks @ceecko ! I agree the issue must have been the `IncorrectUsageError` instead of a race condition, thanks for updating your PR and thanks for contributing to Ghost 🙌

tpatel:
Your contribution landed in [Ghost v4.6.1](https://github.com/TryGhost/Ghost/releases/tag/v4.6.1) 🚢

PR #1996, opened 2021-06-07
Adding Match Helper UI

ErisDS:
- Added toggle UI for the match helper

@kevinansfield I'm trying to setup an example for how to add feature flags 😬 is there anything about this that could or should be improved?

Note this PR introduced the Alpha feature section on the labs page

### DIFF
--- a/app/templates/settings/labs.hbs
+++ b/app/templates/settings/labs.hbs
@@ -241,6 +241,19 @@
                         {{/if}}
                     </div>
                 </div>
+                <div class="gh-expandable-block">
+                    <div class="gh-expandable-header">
+                        <div>
+                            <h4 class="gh-expandable-title">Match helper</h4>
+                            <p class="gh-expandable-description">
+                               Experimental <code>\{{match}}</code> helper for themes.
+                            </p>
+                        </div>
+                        <div class="for-switch">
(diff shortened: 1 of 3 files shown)


app/templates/settings/labs.hbs
```diff
@@ -179,65 +179,93 @@
...
+                            <label class="switch" for="settings-labs-match">
+                                <input
+                                    type="checkbox"
+                                    checked={{this.isMatchEnabled}}
+                                    id="settings-labs-match"
+                                    onclick={{action "toggleLabsMatchHelper" value="target.checked"}}
+                                    data-test-private-checkbox
+                                >
+                                <span class="input-toggle-component"></span>
+                            </label>
```
kevinansfield, line 262: I think this can be replaced with `<GhFeatureFlag @flag="matchHelper" />` then the `isMatchEnabled` and `toggleLabsMatchHelper` changes on the controller can be removed.

The only change that is needed I think is to change `checkbox` to `switch` here https://github.com/TryGhost/Admin/blob/main/app/components/gh-feature-flag.js#L10

ErisDS: Oooooooh

ErisDS: Yeah that totally works 👍

PR #2018, opened 2021-06-24
Re-enable multiple products segmentation on post and email

rishabhgrg:
- Added back product options to members segment select
- Added back ability to set post access to segments
- Enabled product options for email recipient segment

### DIFF
--- a/app/components/gh-members-segment-select.js
+++ b/app/components/gh-members-segment-select.js
@@ -6,6 +6,7 @@ import {tracked} from '@glimmer/tracking';
 
 export default class GhMembersSegmentSelect extends Component {
     @service store;
+    @service feature;
 
     @tracked _options = [];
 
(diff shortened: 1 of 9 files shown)


app/components/gh-members-segment-select.js
```diff
@@ -68,26 +69,28 @@ export default class GhMembersSegmentSelect extends Component {
             class: 'segment-status-paid'
         }];
 
-        // fetch all labels w̶i̶t̶h̶ c̶o̶u̶n̶t̶s̶
-        // TODO: add `include: 'count.members` to query once API is fixed
-        const labels = yield this.store.query('label', {limit: 'all'});
```
kevinansfield, line 73: Removing labels from this component will effect non-labs code. It's used for the default email recipients select.

I'd suggest adding an argument to the component that allows for labels/products to be optional, that way when used for default email recipients select it could show labels+products and in the visibility select it could show only products.

rishabhgrg: Ah yes, good catch, updating

kevinansfield (approved):

PR #2069, opened 2021-08-24
Moved email-only selector to post publish menu

naz:
refs https://github.com/TryGhost/Team/issues/947

The PR aims to implement following design to allow the end user selecting the "distribution" option:
image

@kevinansfield there are couple places I've got stuck with the changes after multiple different approach. Would love your expert help :)

Below are related to changes introduced in `Added publishing action dropdown to publish menu ` commit:
1. The basic test I've tried covering the new component with is failing due to "post" not being an ember model object :thinking: I've taken the pattern from [gh-psm-tags-input-test.js](https://github.com/TryGhost/Admin/blob/faabc6018b01d7839f4b7d98fa3e66442978ae4f/tests/integration/components/gh-psm-tags-input-test.js#L83) test and it doesn't seem to work? Is there a correct pattern to follow when mocking a post model in component unit tests?
2. The second problem is related to the one from pt. 1 I suspect. The component doesn't seem to assign the `distributionValue` correctly, which results on lost state when the publish menu is closed and reopened again.

Soz, these seem like noob level problems but have been stuck at them long enough, so decided to just ask :sweat_smile: :pray:

### DIFF
--- a/app/components/gh-distribution-action-select.hbs
+++ b/app/components/gh-distribution-action-select.hbs
@@ -1,0 +1,12 @@
+<span class="gh-select">
+    <OneWaySelect
+        @value={{this.distributionValue}}
+        @options={{this.availablePublishActions}}
+        @optionValuePath="value"
+        @optionLabelPath="name"
+        @includeBlank={{false}}
+        @update={{this.setPublishAction}}
+        @id="publish-action"
+        @name="publish-action" />
+    {{svg-jar "arrow-down-small"}}
+</span>
--- a/app/components/gh-distribution-action-select.js
+++ b/app/components/gh-distribution-action-select.js
@@ -1,0 +1,50 @@
+import Component from '@glimmer/component';
+import {action} from '@ember/object';
+import {tracked} from '@glimmer/tracking';
+
+export default class GhDistributionActionSelect extends Component {
+    @tracked
+    distributionValue;
+
+    constructor(...args) {
+        super(...args);
+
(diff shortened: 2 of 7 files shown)


app/components/gh-distribution-action-select.js
```diff
@@ -0,0 +1,47 @@
...
+import {tracked} from '@glimmer/tracking';
+
+export default class GhDistributionActionSelect extends Component {
+    @tracked
+    distributionValue;
+
+    constructor() {
+        super(...arguments);
+
+        this.distributionValue = this.args.post.get('emailOnly')
```
kevinansfield, line 12: FYI, `.get()` is not needed anywhere except for accessing proxy objects which at the moment are just the `settings` and `config` services. We need to remove the proxy mixin from those at somepoint too because it's deprecated.

app/components/gh-distribution-action-select.hbs
```diff
@@ -0,0 +1,12 @@
+<span class="gh-select">
+    <OneWaySelect
+        @value={{this.distributionValue}}
+        @options={{this.availablePublishActions}}
+        @optionValuePath="value"
+        @optionLabelPath="name"
+        @includeBlank={{false}}
+        @update={{action "setPublishAction"}}
```
kevinansfield, line 8: `{{action}}` shouldn't be used any more unless you're dealing with an old non-Glimmer component and an actions hash. You've used the `@action` decorator so this can be `@update={{this.setPublishAction}}`.

PR #2072, opened 2021-08-26
Handled dummy subscription for comped members

rishabhgrg:
refs https://github.com/TryGhost/Team/issues/986#issuecomment-900281186

The tiers features updated the "comped" functionality to no longer create a Stripe Subscription for members. Since theme developers expect comped members will have a subscription, we added a dummy subscription object for comped members in API.

Since Admin previously expected comped members to be ones without any subscription but access to product, this change updates Admin behavior to identify comped subscriptions on a product and show them the same way as before.


rishabhgrg:
Closed as changes no longer needed on Admin.

PR #2082, opened 2021-09-06
Added missing schedule events to webhook dropdown

tatsuya-ogawa:
refs https://github.com/TryGhost/Ghost/blob/main/core/server/services/webhooks/listen.js#L18

- update AVAILABLE_EVENTS to latest ghost webhooks.

- [ ] There's a clear use-case for this code change
- [ ] Commit message has a short title & references relevant issues

### DIFF
--- a/app/helpers/event-name.js
+++ b/app/helpers/event-name.js
@@ -11,6 +11,8 @@ export const AVAILABLE_EVENTS = [
     {event: 'post.published', name: 'Post published', group: 'Posts'},
     {event: 'post.published.edited', name: 'Published post updated', group: 'Posts'},
     {event: 'post.unpublished', name: 'Post unpublished', group: 'Posts'},
+    {event: 'post.scheduled', name: 'Post scheduled', group: 'Posts'},
+    {event: 'post.unscheduled', name: 'Post unscheduled', group: 'Posts'},
     {event: 'post.tag.attached', name: 'Tag added to post', group: 'Posts'},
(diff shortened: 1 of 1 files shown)


ErisDS:
Really sorry that this was just ignored. I am unsure about what the use case for these events is, and not sure they should be in the available events list in the first place 🤔 

Going to close for now as we're about to move the code in this repo into a monorepo. 

If you're still keen, please reopen the PR and explain the use case (it's one of the requests in our PR todo list shown above)

cc @naz I'm not sure why we have webhooks for the scheduled state - do you know? 🤔

tatsuya-ogawa:
@ErisDS 
Thank you for your comment.
My use case is to handle scheduled event on other platform(like zapier, aws lambda, etc...) for later task.

In our environment , ghost content is deployed by JAMStack, and ghost server is not running when it is not writing time.
So we cant use ghost's internal cron task, and external elastic job scheduler is needed.

naz:
The event was added way back in 2019 as a part of https://github.com/TryGhost/Ghost/issues/10461. I do not recall and could not find any concrete references behind the "why?"

naz:
Just remembered we do have support for these events in [Ghost-Zapier integration triggers](https://github.com/TryGhost/Zapier/commit/3f8d56a6a1ebf68baa39a232c6665bf492e44b83).

ErisDS:
Ok thanks for the update @naz, have reopened and merged. Thanks for the PR @tatsuya-ogawa 😄

PR #2114, opened 2021-10-18
Add intl.translate

JoseBieco:
# Admin translation
Using ember-intl, it was possible to translate the entire Admin project in a simple and practical way. Being possible to ralize the translations in hbs and js files. Considering that the project will be translated to the language defined in the configurations.

A folder, containing all the json translation files, is located in ``` translations/```.

## Setup application
In the ```app/routes/application.js``` file we have the project language selection.
~~~javascript
await this.get('intl').setLocale([this.settings.get('lang') ? this.settings.get('lang') : 'en']);
~~~
In this code we have a little validation, if ```this.settings.get('lang')``` doesn't return any information, the 'en' language will be set in ```setLocale()```.

This validation was implemented because it is not possible to perform ```this.settings.get('lang')``` while the user is not logged in.

## HBS files
The translation in a .hbs file is very simple:
1. Add the word/phrase you will translate into the .json files;
2. In the .hbs file, where the word would be inserted, it will be done in the following way ```{{t "Manual.YOUR_WORD"}}``` 

## JS files
For components that import ```@ember/component```, intl is instantiated as follows:
~~~javascript 
   intl: service(),
~~~

If the component comes from ```@glimmer/component```, it uses the annotation ```@service``` to instantiate it.
~~~javascript 
export default class MyClass extends Component {
    @service intl;
~~~

The use of intl in both components is given in the following manner:
~~~javascript
    this.intl.t('Manual.YOUR_WORD')
~~~

## Issues
* Project is not 100% translated;
* There is no validation in case the selected language doesn't have a corresponding translation file;
* When switching languages, translations are not automatically applied, so you have to press F5.


ErisDS:
Hi @JoseBieco, I'm really sorry that we didn't get around to reviewing this PR in a reasonable timeframe. I know that i18n is important to the community, and we have not done a good job in supporting contributors to get these things done in recent times.

We're in the process of cleaning up all the PRs on this repo, before moving the admin to be part of the [Ghost monorepo](https://github.com/TryGhost/Ghost). Once this is done, we will be spread less thin over so many repos and more able to help and support contributors working on projects like these - as well as making it much easier to contribute with everything in one place.

If you're still interested in working on i18n for Ghost, I recommend starting a thread in the [contributor section](https://forum.ghost.org/c/contributing/27) of the forum, so that we can discuss all the details of how to approach the project. With something as big as this we need to consider both how the translations work technically and how we go about getting translations done, including docs on language & tone of voice that are super important to making sure that the experience across languages maintains a high level.

There are many, many aspects to i18n that we are falling short on at the moment, and I'm going to be actively looking for contributors to help us improve the experience across the board.

Thanks to both you and @ErickEscagion for the time you put into this, and maybe chat to you in the forums soon.

PR #2144, opened 2021-11-08
Add subject field to email preview

sanne-san:
Refs https://github.com/TryGhost/Team/issues/1082

Currently the email subject field can only be changed in the post settings menu. The reason for moving this to the email preview is so we can get rid of the "Email newsletter" section in the PSM altogether. This is related to a bug where email CTA cards don't show up in test emails if they're triggered from the PSM, because it's not specifically sent to free or paid members. Rather than adding a free/paid selector in the PSM, we'd like to get rid of the duplicate functionality and move everything over to the preview modal.


kevinansfield:
Closing to come back to at a later date as it needs looking at again and we want to clear out PRs before moving Admin to the Ghost monorepo.

PR #2201, opened 2022-01-16
Add optional secret to webhook form

Regrau:
As per Pullrequest https://github.com/TryGhost/Ghost/pull/13980
there needs to be a possibility to add the secret to a webhook.

This PR adds another text input field to the webhook form.

### DIFF
--- a/app/components/modal-webhook-form.hbs
+++ b/app/components/modal-webhook-form.hbs
@@ -63,25 +63,23 @@
             <GhErrorMessage @errors={{this.webhook.errors}} @property="targetUrl" data-test-error="webhook-targetUrl" />
         </GhFormGroup>
     </fieldset>
-    {{#if (enable-developer-experiments)}}
-        <fieldset>
-            <GhFormGroup @errors={{this.webhook.errors}} @hasValidated={{this.webhook.hasValidated}} @property="secret">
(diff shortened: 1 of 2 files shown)


ErisDS:
I added a note to the Ghost PR to explain why this hadn't been merged (which I really thought I had done a long time ago) - the Ghost PR needs updating to only output the header if a secret is set, so that secret stays optional.

Maybe we can fixup both PRs and get them in, as this would be nice to have in.

ErisDS:
Hi @sainthkh , I'm really sorry that we didn't get around to reviewing this PR in a reasonable timeframe. At the moment there are some changes needed to the [server PR](https://github.com/TryGhost/Ghost/pull/13980) and this one needs updating as it doesn't pass tests.

We're in the process of cleaning up all the PRs on this repo, before moving the admin to be part of the [Ghost monorepo](https://github.com/TryGhost/Ghost). Once this is done, getting this over the line if you're still interested will be much easier, as you'll be able to update your [Ghost PR](https://github.com/TryGhost/Ghost/pull/13980) with both sets of changes, and more easily test that it all works together.

If you're still interested in making these changes, please update the PR on the Ghost monorepo once admin is merged in (should be in the next few days). We'll then take a look at the PR right away.

PR #2220, opened 2022-01-21
Updated Ember dependencies to 3.22

kevinansfield:
no issue

- bumped `ember-source`, `ember-data`, and `ember-cli` to latest 3.22.x versions
- fixed errors caused by updating properties inside of a render
  - `<GhPublishMenu>` removed insta-call of `@setSaveType` when rendering sub-components in favour of pre-setting the default `@saveType` value when the underlying post status changes
  - updated `<GhNavMenu>` to use the run-loop to update the `firstRender` property once rendering has finished rather than mid-render


kevinansfield:
The 3.22 upgrade was previously failing due to [this issue](https://github.com/pzuraq/liquid-wormhole/issues/70) in the liquid-wormhole dependency. However, the tests are no longer failing and I've not been able to replicate the error with a basic click-through smoke test.

PR #2226, opened 2022-01-27
[chore] bump to ember 3.28

gabrielcsapo:

Tracking the 3.28 progress, it seems like there are some outstanding octane tasks left, this should pass when those are done. I ran a few codemods locally to see if we can get some of them done.


kevinansfield:
Hey @gabrielcsapo 👋 We've cleared some of the other blockers for this upgrade but I've run into odd errors from Ember Data when it tries to read errors on the response, failing with `TypeError: Cannot read properties of null (reading 'getErrors')`.

We're prepping for moving the Admin repo into the [recently created Ghost monorepo](https://github.com/tryghost/ghost) so old PRs are being cleared up. Seeing as there's no immediately obvious fix and the changes here are fairly small/easy to recreate after the move I'm going to close for now.

Thanks for all your help with the Octane migrations 😊

PR #2227, opened 2022-01-27
[chore] runs native classes codemod

gabrielcsapo:

Runs the native classes codemod

```
All done. 
Results: 
46 errors
70 unmodified
0 skipped
80 ok
```

Failed conversions stem from the same error message about not being able to get the field name.

e.g.

```
 ERR app/components/gh-publishmenu-scheduled.js Transformation error (did not recognize object of type "PropertyDefinition")
```

I will follow up with the rest of the component files as a part of manual conversion.


kevinansfield:
Thanks for looking at this @gabrielcsapo. We're currently doing this migration area by area with a full switch to native classes without using the `@classic` decorator, we tried using that in some places in the past but found it introduced too much confusion with behaviour differences in true native vs @classic classes.

gabrielcsapo:
Hey @kevinansfield ! 

No worries! Long time fan, first time contributor.

We are almost done migrating LinkedIn.com to be fully octane compliant (almost a year later). We had luck running the codemods and then widdleling down the last issues with eslint and template lint enforcement to be the best way to migrate.

The @classic decorator is sometimes confusing, but we found getting the component almost there helped devs focus on the last piece which was cleaning up the component to remove the need for the @classic decorator.

More than happy to migrate and help out however y'all are doing it.

kevinansfield:
@gabrielcsapo that's awesome, very excited to have you as a contributor!

I've discussed the use of `@classic` internally and we think it's time to make the change, especially if we can get the codemod to cover the majority of the codebase in one go. The confusion that came up previously was mostly down to having a sporadic mix of EmberObject, `@classic`, and glimmer. Narrowing down to just `@classic` and glimmer will certainly help with that.

One issue we did run into was with our older modal implementation. When old modal components (the `components/modal-*.js` ones) were switched to `@classic` we started getting `Uncaught TypeError: Cannot set property confirm of [object Object] which has only a getter` errors - this was many months ago and the change was reverted so I don't know if it was an implementation bug or a general incompatibility. It's not a blocker, just something to be aware of. We're slowly migrating these components to use an `ember-promise-modals` implementation with native glimmer components.

I'm happy to help co-ordinate these changes, provide assistance, and unblock any issues that are run into!

gabrielcsapo:
Woot! Looks like a lot of things got worked out as those errors aren't popping up anymore. I will make sure this passes locally and smoke test by actually building it in the context of the ghost-cli.

gabrielcsapo:
@kevinansfield would it be possible to have the workflow approved for this PR to make sure I am passing all checks? I can run it locally, but it will not be as comprehensive as having CI run.

kevinansfield:
Workflow approved again, not sure why it didn't remember the previous approval!

kevinansfield:
Noticed the tests were failing on the custom view modal. In `main` I've [migrated it to use the newer modals setup](https://github.com/TryGhost/Admin/commit/1ce4c10bae534d4d23126e353c8bd5ab75151330)

gabrielcsapo:
I will rebase @kevinansfield

gabrielcsapo:
It is so weird the workflow keeps needing to be approved. I am running tests locally.

gabrielcsapo:
Working through the errors locally, should have a working PR the time the workflow is approved.

gabrielcsapo:
One of the failing tests is
```
not ok 26 Chrome 97.0 - [501 ms] - Exam Partition 2 - Acceptance: Editor when logged in renders the editor correctly, PSM Publish Date and Save Button
111
    ---
112
        message: >
113
            Cannot read properties of null (reading 'textContent')
```

Which is also failing for  https://github.com/TryGhost/Admin/pull/2232. I am fixing that in this PR and will pull in that dep update.

gabrielcsapo:
As a side note, these docs https://ghost.org/docs/install/source/ are amazing 🙌

gabrielcsapo:
`renders the editor correctly, PSM Publish Date and Save Button` definitely fails due to the timezones being wrong in tests. I can validate the behavior when serving the admin within a running ghost instance, but can't figure out why the tests timezone is one day ahead.

gabrielcsapo:
Okay I figured it out, this is actually not broken the post needs to be saved before validation happens, this is the case in production as well as these tests fail without any changes locally right now.

gabrielcsapo:
Screen Shot 2022-01-31 at 8 08 07 PM

Fixed locally, looking through more errors.

gabrielcsapo:
Currently only 4 errors remaining

gabrielcsapo:
okay with this we should be good, to go. I am rolling back the modal changes because that requires we go through each modal and verify they work

gabrielcsapo:
rebasing now

gabrielcsapo:
Ran tests locally and no failures after my rebase and merge conflict fix.

gabrielcsapo:
Screen Shot 2022-02-01 at 12 26 48 AM

kevinansfield:
This is great, thank you @gabrielcsapo! ❤️ I agree with rolling back the modal changes, it will be better to handle those separately.

Let's get this batch of changes merged and we can look at the more troublesome parts individually so they're easier to review 🙂

gabrielcsapo:
https://github.com/TryGhost/Ghost/issues/14101

PR #2254, opened 2022-02-09
[chore] migrate app/controllers/settings/navigation.js to use native classes

gabrielcsapo:
Required as a part of https://github.com/TryGhost/Admin/pull/2245


app/controllers/settings/navigation.js
```diff
@@ -1,133 +1,162 @@
...
+    @computed('config.blogUrl')
+    get blogUrl() {
         let url = this.get('config.blogUrl');
 
         return url.slice(-1) !== '/' ? `${url}/` : url;
-    }),
+    }
 
-    actions: {
-        save() {
```
gabrielcsapo, line 34: This action was actually never used, so it was removed.

PR #2263, opened 2022-02-15
Switched to fork of `ember-mocha` with up-to-date dependencies

kevinansfield:
refs https://github.com/TryGhost/Team/issues/1371

- we were running in to issues with expected Ember test helpers being missing due to an old version of `@ember/test-helpers` being bundled with `ember-mocha`
- `ember-mocha` appears to be unmaintained but a fork does exist that sorts out the dependency issues
  - https://github.com/yads/ember-mocha
- updated `<GhTaskButton>` tests because `defineProperty` no longer appears to work with the updated dependencies
  - assigned a class instance containing a task to a local `context` property


gabrielcsapo:
Just saw https://github.com/TryGhost/Admin/pull/2261 which is a migration from mocha to qunit! Super cool

PR #2265, opened 2022-02-15
🐛 Fixed bug where the slug of a tag wasn't shown on the tag list view

cino:
This will solve an issue where the tag slugs aren't shown on the tag list view.

For example, in the screenshot below you'll see that none of the slug fields are filled; 
image

I did modify the test regarding the tags, however, as the complete test is being skipped I experienced that the test in a whole is outdated. I am willing to rewrite the whole test later this week to make it actually represent the current state of the tag section in the admin panel. 

Tests:
- [x] Rewrite skipped acceptance test
- [ ] Rewrite skipped integration test

### DIFF
--- a/app/components/gh-tag-settings-form.hbs
+++ b/app/components/gh-tag-settings-form.hbs
@@ -103,7 +103,7 @@
                 <h4 class="gh-expandable-title">Meta data</h4>
                 <p class="gh-expandable-description">Extra content for search engines.</p>
             </div>
-            <button type="button" class="gh-btn gh-btn-expand" {{action (toggle "metadataOpen" this)}}><span>{{if this.metadataOpen "Close" "Expand"}}</span></button>
+            <button type="button" name="meta" class="gh-btn gh-btn-expand" {{action (toggle "metadataOpen" this)}}><span>{{if this.metadataOpen "Close" "Expand"}}</span></button>
         </div>
 
         <div class="gh-expandable-content">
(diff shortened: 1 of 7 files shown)


ErisDS:
Hey @cino it would be awesome if you could rewrite the tests. We're having a big push to get our tests sorted atm, and that would be an incredible help!

cino:
@ErisDS Cool, I'll take some time this week to look into that specific test. Would you prefer it in a separate pull request?

cino:
@ErisDS It's probably good to first decide if https://github.com/TryGhost/Admin/pull/2261 is getting merged before dedicating time to it.

kevinansfield:
@cino please continue with the existing test suite.

cino:
I've updated the acceptance test to include the following working situations:

- Correct redirects for roles that aren't supposed to be here
- Creating/editing a tag and confirming the changes
- Deleting a tag
- Displaying the correct slug on the tag list
- Sorting the tag list correctly
- Displaying the correct tags on the internal/external tag list
- Updated the sorting of the tags to sort correctly when using numbers at the start of the tag (11 after 2 instead of first 1, second 11, third 2)

Will be updating with additional tests for;
- Showing the correct post count on the tag list
- Look into testing the image uploads for the tag/socials

However, found additional issues;
- When creating a new tag and directly pushing the save button without entering any info the form will invalidate and you will not be able to save the form again (even when filling it correctly)
- The Slug of the tag doesn't update after modifying the name when refocussing on the name field.

ErisDS:
Hey @cino did you mean to close this?

PR #2268, opened 2022-02-16
Onboarding refinements

peterzimon:
- Added dynamic properties per theme: not all the properties (logo, cover image etc.) make sense for all the themes so we need to add a basic way to filter out the unwanted properteis.
- Removed showing the last modal (didn't remove the modal itself)
- Refined UI for finishing screen

### DIFF
--- a/app/controllers/setup/finishing-touches.js
+++ b/app/controllers/setup/finishing-touches.js
@@ -1,12 +1,67 @@
 import Controller from '@ember/controller';
+import {isEmpty} from '@ember/utils';
 import {inject as service} from '@ember/service';
 import {task} from 'ember-concurrency';
+import {tracked} from '@glimmer/tracking';
+
+const THEME_PROPERTIES = {
+    casper: ['description', 'color', 'coverImage'],
+    edition: ['description', 'color', 'coverImage'],
+    dawn: ['description', 'color', 'icon'],
+    dope: ['description', 'color', 'logo'],
+    bulletin: ['description', 'color', 'logo'],
+    alto: ['description', 'color', 'logo'],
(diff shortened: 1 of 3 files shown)


app/controllers/setup/finishing-touches.js
```diff
@@ -1,17 +1,71 @@
...
+    @tracked coverImageVisible;
+
+    themes = null;
+
+    // Default properties to display
+    themeProperties = ['description', 'color', 'coverImage'];
+
+    constructor() {
+        super(...arguments);
+        this.initThemeProperties.perform();
```
kevinansfield, line 43: This is a controller not a component so we don't really want to use a constructor here because controllers are singletons and we can't guarantee when the instance will be created.

I think the better approach would be to move the theme loading to the route's `model()` hook so we know themes will be available before the template is rendered:

```
model() {
   return this.store.findAll('theme');
}
```

That will allow us to remove everything in the controller except for a couple of getters, eg:

```
get activeTheme() {
    return this.store.peekAll('theme').findBy('active');
}

get themeProperties() {
    const themeProperties = THEME_PROPERTIES[this.activeTheme.name] || ['description', 'color', 'coverImage'];

    // converts ['color', 'description'] to {color: true, description: true} for easier use in template
    return Object.fromEntries(themeProperties.map(prop => [prop, true]));
}
```

in the template you can then change the conditionals to be `{{#if this.themeProperties.description}}`, {{#if this.themeProperties.color}}` etc

kevinansfield:
As far as I can tell the finishing-touches screen no longer exists so this PR isn't necessary. Closing as cleanup prior to monorepo move.

PR #2272, opened 2022-02-21
Allows for usernames to begin with http or www

wwwroth:
This extra checking on the Twitter username after removing the Twitter URL is redundant and blocks anyone with a username that actually starts with "http" or "www". In my case, mine is @wwwroth and I cannot save my Twitter URL.

### DIFF
--- a/app/controllers/settings/general.js
+++ b/app/controllers/settings/general.js
@@ -233,8 +233,7 @@ export default class GeneralController extends Controller {
                 [username] = newUrl.match(/([^/]+)\/?$/mi);
             }
 
-            // check if username starts with http or www and show error if so
-            if (username.match(/^(http|www)|(\/)/) || !username.match(/^[a-z\d._]{1,15}$/mi)) {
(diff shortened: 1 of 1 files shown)


cino (approved):

ErisDS:
Hi @sainthkh , I'm really sorry that we didn't get around to reviewing this PR in a reasonable timeframe. I understand why this bug is annoying and it would be great to get in a fix - ideally one that preserved the original intention of preventing people pasting twitter URLs into the field.

We're in the process of cleaning up all the PRs on this repo, before moving the admin to be part of the [Ghost monorepo](https://github.com/TryGhost/Ghost). This is to make developing on and contributing to Ghost easier. With PRs all consolidated in one repo, we intend to do a better job of keeping on top of PRs.

If you're still interested in making these changes, please re-create the PR on the Ghost monorepo once admin is merged in (should be in the next few days). We'll then take a look at the PR right away.

PR #2284, opened 2022-03-01
🐛 Remove empty benefits before saving

SimonBackx:
refs https://github.com/TryGhost/Team/issues/1188

- The api throws a validation error when we try to add benefits with an empty name
- Before saving, we remove benefits with an empty name


SimonBackx:
Todo: add a test for this behaviour

SimonBackx:
@matthanley Test is added 😊

matthanley (approved):

PR #2286, opened 2022-03-02
Removed third step to invite staff users from self-hoster setup

sanne-san:
No issue

- The step to invite staff users during site setup for self-hosters is unnecessary and out of style with the rest of admin, and is therefore removed.


matthanley:
❤️ really looking forward to not having to skip this every time!

PR #2287, opened 2022-03-03
✨ Added canceled subscriptions in member detail screen

tpatel:
refs https://github.com/TryGhost/Team/issues/1141

- The canceled subscriptions provide a more complete picture of the activity of a member.

### DIFF
--- a/app/components/gh-member-settings-form.js
+++ b/app/components/gh-member-settings-form.js
@@ -28,6 +28,10 @@ export default class extends Component {
     }
 
     get isAddComplimentaryAllowed() {
+        if (this.member.get('isNew')) {
+            return false;
+        }
+
         if (!this.membersUtils.isStripeEnabled) {
             return false;
         }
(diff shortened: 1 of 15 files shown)


tpatel:
Depends on https://github.com/TryGhost/Ghost/pull/14298

app/components/gh-member-settings-form.js
```diff
@@ -45,8 +45,14 @@ export default class extends Component {
     }
 
     get products() {
-        let products = this.member.get('products') || [];
         let subscriptions = this.member.get('subscriptions') || [];
+        // Create the products from `subscriptions.price.product`
+        let products = subscriptions
```
matthanley, line 50: This change causes a regression [here](https://github.com/TryGhost/Admin/blob/0ab6acd8d9dd43686e843ccb7c166f7c0d25592c/app/components/gh-member-settings-form.hbs#L69), because products is no longer empty - which means there's no option to give the member a comped subscription.

matthanley (changes requested):

app/components/gh-member-settings-form.js
matthanley: @tpatel we can reinstate the "Add complimentary plan" (not exact wording) functionality using a similar pattern to the one used in the UI for Tiers:

image

PR #2296, opened 2022-03-08
Switched "Last seen" filter back to standard datepicker without shortcuts

kevinansfield:
refs https://github.com/TryGhost/Team/issues/1418

- drops the block usage of `<GhDatePicker>` for the "Last seen" input that added shortcut buttons
  - not a full revert of https://github.com/TryGhost/Admin/pull/2291 as allowing block usage of the component is still useful for other component customisation
- removes CSS related to the calendar-with-shortcut buttons as we don't have any other use-cases for it at the moment


moreofmorris:
Looks good to me 👌

moreofmorris:
Looks good to me 👌

PR #2332, opened 2022-04-11
[WIP] Updated subscription event handling

allouis:
@moreofmorris Here's some scaffolded changes for the updated subscription event types

### DIFF
--- a/app/components/dashboard/latest-member-activity.hbs
+++ b/app/components/dashboard/latest-member-activity.hbs
@@ -37,7 +37,7 @@
                                                         <div class="gh-dashboard-activity-event">
                                                             {{parsedEvent.action}}
                                                             {{parsedEvent.object}}
-                                                            <span class="highlight">{{parsedEvent.info}}</span>
+                                                            {{parsedEvent.info}}
                                                         </div>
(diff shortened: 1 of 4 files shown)


moreofmorris:
It's a little difficult to test all the event types locally with any "real" data to test all icons — we usually something like Rediverse for that to do a proper test. But one thing I'm noticing is that since you're changing this: `event.data.from_plan` and `event.data.to_plan` in `getAction()`, do we need to do the same in `getIcon()` in the same file as that still seems to reference those?

moreofmorris:
It's a little difficult to test all the event types locally with any "real" data to test all icons — we usually something like Rediverse for that to do a proper test. But one thing I'm noticing is that since you're changing this: `event.data.from_plan` and `event.data.to_plan` in `getAction()`, do we need to do the same in `getIcon()` in the same file as that still seems to reference those?

allouis:
Yeah good shout - I think we do wanna update `getIcon` to show the cancelled icon when the type is `'canceled'`!

moreofmorris:
All good now @allouis, feel free to merge 👍

PR #2371, opened 2022-05-09
New empty state for new Dashboard

moreofmorris:
Just wanted to get an eye on the changes here for this as a lot has been introduced that may mess with states. I may or may not get a review on this.


app/components/dashboard/v5/charts/anchor.js
```diff
@@ -65,6 +65,7 @@ export default class Anchor extends Component {
     @service dashboardStats;
     @service feature;
     @tracked chartDisplay = 'total';
+    @tracked totalMembersOverall = null;
```
SimonBackx, line 68: This is not used, so can be removed :) 
```suggestion
```

app/components/dashboard/v5/charts/anchor.js
```diff
@@ -112,6 +113,14 @@ export default class Anchor extends Component {
...
     }
 
+    get isTotalMembersZero() {
+        return this.dashboardStats.memberCounts && this.totalMembers === 0;
+    }
+
+    get isTotalMembersMoreThanZero() {
+        return this.dashboardStats.memberCounts && this.totalMembers > 0;
+    }
+
```
SimonBackx, line 123: I don't think this is used
```suggestion
```

app/components/dashboard/v5/charts/overview.js
```diff
@@ -20,6 +20,14 @@ export default class Overview extends Component {
         return this.dashboardStats.memberCounts?.total ?? 0;
     }
 
+    get isTotalMembersZero() {
+        return this.dashboardStats.memberCounts && this.totalMembers === 0;
+    }
```
SimonBackx, line 25:
```suggestion
```

moreofmorris:
Just attended to those things you mentioned @SimonBackx, good catches!

I'm not really sure how to fix the Codecov Report and ensure more coverage, that's not something I've done at Ghost yet — is that something you can give some guidance or help on? Also, I'm still wondering what to do with all the hard coded fake data I have placed inside the components — is that really the best way?

SimonBackx:
@moreofmorris I've spent some time working out a better solution and was trying to get it to work with as few adjustments as possible. But a better pattern here would essentially mean that we have to rebuild the whole fetching system of the dashboard, and with the deadline getting closer, I don't think it is a very good idea to start going down this rabbit hole. If we have time left, we might reconsider this later. For now, I would just merge as is.

A better pattern in this case could mean that we would use the `dashboardMocks` service to generate the 'empty state' graphs. But we cannot use the `dashboardMocks` service twice at the same time when we are already using it for the real data mocking (the one that will contain zero members). One idea would be to make the dashboardMocks more class based, so we can have multiple instances, one for the real data, and one for the empty state. Also, all the state variables in the dashboardStats service would need to be wrapped, so that depending on the amount of members, we would either return the actual data or the fake data. 

As for the code coverage: there are currently no tests for dashboard 5.0. But I'm not sure if I can spend time on writing tests now. We need to revisit this later.

SimonBackx (approved):

moreofmorris:
This has already been merged

PR #2372, opened 2022-05-09
Renamed products to tiers

ErisDS:
This PR does a find and replace of the word product to tier :D 

It avoids:
- kg-product-card, that really is meant to say product
- `stripe_product` settings as those have not been renamed yet (if they are going to be)
- `product-cadence` on offers 
- the word production 😂 
- references to the product team

refs: TryGhost/Team#1145

- this should allow us to remove the /products endpoint in v5


app/services/dashboard-stats.js
```diff
@@ -350,15 +350,15 @@ export default class DashboardStatsService extends Service {
...
         }
 
-        yield this.loadPaidProducts();
+        yield this.loadPaidTiers();
 
         const paidMembersByTier = [];
 
         for (const tier of result.meta.tiers) {
-            const product = this.paidProducts.find(x => x.id === tier);
+            const _tier = this.paidTiers.find(x => x.id === tier);
```
ErisDS, line 358: Not sure why product and tier were being used interchangeably here, but I had to rename this to `_tier` as there's already a `tier` variable 🤔

app/services/members-utils.js
```diff
@@ -121,8 +121,8 @@ export default class MembersUtilsService extends Service {
             settingsParam.append('portalPrices', encodeURIComponent(portalPlans));
         }
 
-        if (portalProducts && this.feature.get('multipleProducts')) {
-            settingsParam.append('portalProducts', encodeURIComponent(portalProducts));
+        if (portalTiers && this.feature.get('multipleProducts')) {
```
ErisDS, line 124: Not sure about this one

app/models/stripe-price.js
```diff
@@ -3,7 +3,7 @@ import EmberObject from '@ember/object';
 export default EmberObject.extend({
     id: 'ID in Ghost',
     stripe_price_id: 'ID of the Stripe Price',
-    stripe_product_id: 'ID of the Stripe Product the Stripe Price is associated with',
+    stripe_product_id: 'ID of the Stripe Tier the Stripe Price is associated with',
```
ErisDS, line 6: ⚠️ Doh this one needs reverting

app/templates/settings/tier.hbs
```diff
@@ -3,16 +3,16 @@
         <h2 class="gh-canvas-title" data-test-screen-title>
             <LinkTo @route="settings">Settings</LinkTo>
             <span>{{svg-jar "arrow-right"}}</span>
-            <LinkTo @route="settings.products">Products</LinkTo>
+            <LinkTo @route="settings.tiers">Tiers</LinkTo>
```
ErisDS, line 6: Is this a defunct template? I didn't think we had the word Product anywhere user-facing any more 🤔

ErisDS (commented):
@rishabhgrg I have left some comments on parts I'm not sure about. Would really appreciate your eyes on this. I'm aware there's one issue I need to fixup.

app/models/stripe-price.js
ErisDS: Fixed 👍

ErisDS:
https://github.com/TryGhost/Ghost/blob/c606048ab729235c5e66f8728f1609fff172c4e0/core/server/api/canary/utils/serializers/output/tiers.js#L104 

hmm, now I'm not sure about stripe_product_id and where it should be what 🙈

app/services/members-utils.js
rishabhgrg: Hm yeah this one might not work right now, because we use the setting called `portalProducts` in Portal preview to show all the selected products in Portal, so this won't be just a variable change. I think it's ok to keep this one as `portalProducts` for now ?

allouis:
> hmm, now I'm not sure about stripe_product_id and where it should be what 

That's my mistake - I'm currently in the process of removing that key from the API entirely though!

app/services/members-utils.js
rishabhgrg: Updated!

PR #2380, opened 2022-05-13
Updated publishing with renamed newsletter and email_segment options

SimonBackx:
refs https://github.com/TryGhost/Team/issues/1596

This PR updates admin to align with the changes in the backend in https://github.com/TryGhost/Ghost/pull/14798 (not yet merged)

- `post.email_recipient_filter` option and property is renamed to `email_segment`
- `newsletter_id` option is renamed to `newsletter` (and now uses slug instead of id)
- Sending a post via email, means you need to set the `newsletter` option. The `email_segment` option is optional now and defaults to `all`. Not setting `newsletter` means not sending an email, setting `email_segment` is ignored if `newsletter` is missing.
- We can no longer use `email_recipient_filter` (now renamed to `email_segment`) to know whether a post is published/scheduled to be sent as an email. We need to rely on the newsletter relation now. The `email_segment` `none` value has been dropped and will be `all` even if no email will be sent (newsletter will be null). 
- `sendEmailWhenPublished` option is no longer supported in the backend.

### DIFF
--- a/app/components/gh-posts-list-item.hbs
+++ b/app/components/gh-posts-list-item.hbs
@@ -51,13 +51,11 @@
         {{#if (and this.feature.emailAnalytics (eq @post.displayName "post"))}}
             <LinkTo @route="editor.edit" @models={{array @post.displayName @post.id}} class="permalink gh-list-data gh-post-list-recipients">
                 <div class="flex fw4">
-                    {{#if (or @post.email @post.willEmail)}}
-                        {{#if (eq @post.email.status "submitted")}}
-                            <span class="flex" data-tooltip="{{capitalize @post.email.recipientFilter}} members">
-                                <span class="darkgrey fw5 gh-content-email-stats">{{format-number @post.email.emailCount}}</span>
-                                <span class="midgrey-l2 fw4 gh-content-email-stats-mobile">{{gh-pluralize @post.email.emailCount "send"}}</span>
-                            </span>
-                        {{/if}}
+                    {{#if (eq @post.email.status "submitted")}}
+                        <span class="flex" data-tooltip="{{capitalize @post.email.recipientFilter}} members">
+                            <span class="darkgrey fw5 gh-content-email-stats">{{format-number @post.email.emailCount}}</span>
+                            <span class="midgrey-l2 fw4 gh-content-email-stats-mobile">{{gh-pluralize @post.email.emailCount "send"}}</span>
+                        </span>
                     {{/if}}
                 </div>
             </LinkTo>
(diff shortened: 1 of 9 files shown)


app/components/gh-posts-list-item.hbs
```diff
@@ -51,7 +51,7 @@
         {{#if (and this.feature.emailAnalytics (eq @post.displayName "post"))}}
             <LinkTo @route="editor.edit" @models={{array @post.displayName @post.id}} class="permalink gh-list-data gh-post-list-recipients">
                 <div class="flex fw4">
-                    {{#if (or @post.email @post.willEmail)}}
```
SimonBackx, line 54: Removed this because I don't think this made sense because the next line specifically checks for email, and a scheduled, not sent post, will never have email counts.

app/components/gh-posts-list-item.js
```diff
@@ -11,11 +11,6 @@ export default class GhPostsListItemComponent extends Component {
 
     @tracked isHovered = false;
 
-    get sendEmailWhenPublished() {
```
SimonBackx, line 14: This was not used

app/components/gh-posts-list-item.hbs
kevinansfield: 👍 We can remove the conditional completely, the following `{{#if (eq @post.email.status "submitted")}}` is all that's needed

kevinansfield (approved):

app/models/post.js
```diff
@@ -160,8 +160,8 @@ export default Model.extend(Comparable, ValidationEngine, {
     hasEmail: computed('email', 'emailOnly', function () {
         return this.email !== null || this.emailOnly;
     }),
-    willEmail: computed('emailRecipientFilter', function () {
-        return this.emailRecipientFilter !== null;
+    willEmail: computed('isScheduled', 'newsletter', function () {
+        return this.isScheduled && !!this.newsletter;
```
kevinansfield, line 164: The logic here wasn't quite correct, updated to reflect current main

```suggestion
    willEmail: computed('isScheduled', 'newsletter', 'email', function () {
        return this.isScheduled && !!this.newsletter && !!this.email;
```

app/components/gh-posts-list-item.hbs
SimonBackx: Removed 🔥

app/models/post.js
kevinansfield: > Do you think we also should update the same check in HBS at app/components/editor-labs/modals/update-flow.hbs

Yeah, replacing `(and post.isScheduled post.newsletter (not post.email))` with `post.willEmail` is nicer 👌

PR #2418, opened 2022-06-18
Updated Tenor API to v2

arsereg:
Fixes the Tenor integration

Address #14980

### DIFF
--- a/app/services/tenor.js
+++ b/app/services/tenor.js
@@ -201,7 +201,6 @@ export default class TenorService extends Service {
         const [width, height] = gif.media_formats.tinygif.dims;
         gif.ratio = height / width;
 
-
         // add to general gifs list
         this.gifs.push(gif);
 
--- a/app/services/tenor.js
+++ b/app/services/tenor.js
@@ -160,9 +160,8 @@ export default class TenorService extends Service {
 
                 if (this.error && this.error.startsWith('API key not valid')) {
                     // Added an html error field, so that we don't pass raw API errors from tenor to triple-braces in the frontend
-                    this.htmlError = `This version of the Tenor API is no longer supported.<br />
-Please generate a new API key and update your configuration by following our
-<a href="https://ghost.org/docs/config/#tenor" target="_blank" rel="noopener noreferrer"> documentation here</a>.`;
+                    this.htmlError = `This version of the Tenor API is no longer supported. Please update your API key by following our
+<a href="https://ghost.org/docs/config/#tenor" target="_blank" rel="noopener noreferrer"> documentation here</a>.<br />`;
                 }
                 console.error(e); // eslint-disable-line
             });
--- a/app/services/tenor.js
+++ b/app/services/tenor.js
@@ -6,8 +6,8 @@ import {isEmpty} from '@ember/utils';
 import {task, taskGroup, timeout} from 'ember-concurrency';
 import {tracked} from '@glimmer/tracking';
 
-const API_URL = 'https://g1.tenor.com';
-const API_VERSION = 'v1';
+const API_URL = 'https://tenor.googleapis.com';
+const API_VERSION = 'v2';
 const DEBOUNCE_MS = 600;
 
 export default class TenorService extends Service {
@@ -17,14 +17,18 @@ export default class TenorService extends Service {
     @tracked columns = null;
     @tracked error = null;
     @tracked gifs = new TrackedArray([]);
-    @tracked searchTerm = '';
-    @tracked loadedType = 'trending';
+    @tracked searchTerm = 'excited';
+    @tracked loadedType = 'search';
 
     _columnHeights = [];
     _nextPos = null;
 
     get apiKey() {
-        return this.config.get('tenor.publicReadOnlyApiKey');
+        return this.config.get('tenor.apiKey');
+    }
+
+    get clientId() {
+        return this.config.get('tenor.clientId');
     }
 
     get contentfilter() {
@@ -94,9 +98,10 @@ export default class TenorService extends Service {
 
     @task({group: 'loadingTasks'})
     *loadTrendingTask() {
-        this.loadedType = 'trending';
+        this.loadedType = 'search';
 
         yield this._makeRequest(this.loadedType, {params: {
+            q: 'excited',
             media_filter: 'minimal'
         }});
     }
@@ -134,6 +139,7 @@ export default class TenorService extends Service {
 
         const params = new URLSearchParams(options.params);
         params.set('key', this.apiKey);
+        params.set('client_key', this.clientId);
         params.set('contentfilter', this.contentfilter);
 
         url.search = params.toString();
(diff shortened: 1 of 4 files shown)


app/services/tenor.js
```diff
@@ -6,8 +6,8 @@ import {isEmpty} from '@ember/utils';
 import {task, taskGroup, timeout} from 'ember-concurrency';
 import {tracked} from '@glimmer/tracking';
 
-const API_URL = 'https://g1.tenor.com';
-const API_VERSION = 'v1';
+const API_URL = 'https://tenor.googleapis.com';
+const API_VERSION = 'v2';
```
markstos, line 10: 👍🏼 Matches the endpoint changes docs.

app/services/tenor.js
```diff
@@ -17,14 +17,18 @@ export default class TenorService extends Service {
     @tracked columns = null;
     @tracked error = null;
     @tracked gifs = new TrackedArray([]);
-    @tracked searchTerm = '';
-    @tracked loadedType = 'trending';
+    @tracked searchTerm = 'excited';
+    @tracked loadedType = 'search';
```
markstos, line 21: ❔ Why the change from an empty search term to 'excited'? This is not required for the migration. 

❔Also, why change from "trending" to search? The new API also supports returning trending GIFs?

app/services/tenor.js
```diff
@@ -17,14 +17,18 @@ export default class TenorService extends Service {
...
 
     _columnHeights = [];
     _nextPos = null;
 
     get apiKey() {
-        return this.config.get('tenor.publicReadOnlyApiKey');
+        return this.config.get('tenor.apiKey');
```
markstos, line 27: ❔ Why rename the API key in the config system?

app/services/tenor.js
```diff
@@ -94,9 +98,10 @@ export default class TenorService extends Service {
 
     @task({group: 'loadingTasks'})
     *loadTrendingTask() {
-        this.loadedType = 'trending';
+        this.loadedType = 'search';
 
         yield this._makeRequest(this.loadedType, {params: {
+            q: 'excited',
```
markstos, line 104: It's also not clear where the change was made here from a "trending" query to an "excited" search.

app/services/tenor.js
```diff
@@ -192,9 +198,10 @@ export default class TenorService extends Service {
 
     _addGif(gif) {
         // re-calculate ratio for later use
-        const [width, height] = gif.media[0].tinygif.dims;
+        const [width, height] = gif.media_formats.tinygif.dims;
```
markstos, line 201: 👍🏼 This matches the API migration doc notes about what to change.

app/services/tenor.js
```diff
@@ -192,9 +198,10 @@ export default class TenorService extends Service {
...
         gif.ratio = height / width;
 
+
```
markstos, line 204: This extra blank line can be removed.

lib/koenig-editor/addon/components/koenig-card-image/selector-tenor.js
```diff
@@ -58,7 +58,8 @@ export default class KoenigCardImageTenorSelector extends Component {
         event?.preventDefault();
         event?.stopPropagation();
 
-        const media = gif.media[0].gif;
+
```
markstos, line 61: Remove the extra blank line here.

markstos (changes requested):
I'm felllow Ghost user comparing these changes to the API documentation for how to migrate from the Tenor v1 API to V2:

https://github.com/TryGhost/Admin/pull/2418/files

The changes generally look good but I have questions about some of the changes. Thanks.

app/services/tenor.js
arsereg: The API has changed not only in the URL but also in functionality. We can no longer search for "trending" gifs, as the result for that will be gifs representing "Trending stuff", such as statistics graphs or arrows pointing up.

I'm using excited as it is part of the new documentation they provide to show "happy" or "exciting" gifs on the search when there hasn't been any input yet.

app/services/tenor.js
arsereg: The property in the new tenor API has been modified. It doesn't represent a Public Read Only Key anymore, but a Google API api-key.

app/services/tenor.js
arsereg: In the previous API definition, it allowed to pass a empty query, and the endpoint would receive the "trending" parameter. This would return the trending gifs in the app. This functionality has been removed, instead, the new documentation indicates that there must be a search done on each request, we can no longer send an empty query, neither can we request for the "trending" gifs.

In the new documentation (https://tenor.com/gifapi/documentation) they make examples querying for "excited" to show "Happy" and "Excited" gifs. I've took that example and implemented it here, so when the user loads the Tenor card, it'll load with "excited" gifs.

If we don't update the loadedType to `search` it will provoke a 404 error since that endpoint no longer exists.
Also, if we use 'search" as loadedType but we don't send parameter, it will return a 400 error.

arsereg:
Thanks for the comments @markstos!

In regards to the "excited" search by default, I think it is the best approach as even Tenor is using that as the "default" value for searches on their documentation, however, I'm not 100% convinced that this is the best solution. Please let me know if you think there is the best approach for this.

Thanks!

arsereg:
Correct, we need to update the documentation as well, since current documentation is pointing to get a V1 API Key wich is deprecated now

arsereg:
@markstos I created a Pull Request for the "Ghost" project: https://github.com/TryGhost/Ghost/pull/15087

markstos:
@arsereg I left a comment there-- LGTM.

ErisDS:
As this involves some changes to functionality, it needs a little bit of a deeper review than just code. We're actively looking at it now, but it may take a few days 🙂

ErisDS:
I have reviewed these changes in great detail, and made a couple of adjustments:

1. Instead of using the "excited" search, I'm using the new "featured" API which kind of replaces the old trending API.
2. I've fixed up the error handling to support Google's error format, and also caught and replaced the error that everyone with old keys will see to make it clearer.
3. I've replaced client ID with a fixed string as it was undefined

I have one last thing to do which is confirm the changes to the config structure - this may also get a tweak, which will require a similar update in the server side PR.

app/services/tenor.js
```diff
@@ -153,6 +156,13 @@ export default class TenorService extends Service {
                 if (!options.ignoreErrors && !this.error) {
                     this.error = 'Uh-oh! Trouble reaching the Tenor API, please check your connection';
                 }
+
+                if (this.error && this.error.startsWith('API key not valid')) {
+                    // Added an html error field, so that we don't pass raw API errors from tenor to triple-braces in the frontend
+                    this.htmlError = `This version of the Tenor API is no longer supported.<br />
```
kevinansfield, line 162: `htmlError` needs to be marked as a tracked property with `@tracked htmlError = null;` otherwise templates won't react to it's value changing.

PR #2433, opened 2022-07-27
Ghost Explore integration

aileen:
needs https://github.com/TryGhost/Ghost/pull/15103
no issue

- Added `/explore` route which requires min. Admin and is behind a feature flag
- Fetches Admin API key and ID to create a token and return back to Ghost Explore with the correct query params
- Fullscreen UI

### DIFF
--- a/app/routes/explore.js
+++ b/app/routes/explore.js
@@ -1,0 +1,19 @@
+import AdminRoute from 'ghost-admin/routes/admin';
+import {inject as service} from '@ember/service';
+
+export default class ExploreRoute extends AdminRoute {
+    @service session;
+    @service store;
+    @service feature;
+
+    beforeModel() {
+        super.beforeModel(...arguments);
+        if (!this.feature.explore) {
+            return this.transitionTo('home');
+        }
+    }
+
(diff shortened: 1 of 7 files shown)


app/routes/explore.js
```diff
@@ -0,0 +1,19 @@
...
+import {inject as service} from '@ember/service';
+
+export default class ExploreRoute extends AdminRoute {
+    @service session;
+    @service store;
+    @service feature;
+
+    beforeModel() {
+        super.beforeModel(...arguments);
+        if ((!this.session.user.isAdminOnly && !this.session.user.isOwnerOnly) || !this.feature.explore) {
```
kevinansfield, line 11: `(!this.session.user.isAdminOnly && !this.session.user.isOwnerOnly)` isn't needed here because that's already handled by extending from `AdminRoute` 🙂

aileen: Removed!

aileen:
@kevinansfield this is good to go from my side if you don't have objections!

kevinansfield (approved):

